PR #3, opened 2020-04-06
DOCSP-9548: Quick Start guide

ccho-mongodb:
JIRA: 
https://jira.mongodb.org/browse/DOCSP-9548

Staging:
https://docs-mongodbcom-staging.corp.mongodb.com/a8289ad/java/docsworker-xlarge/DOCSP-9548-quickstart/quick-start


source/includes/quick-start/next-steps.rst
```diff
@@ -0,0 +1,3 @@
+Learn how to read and modify data using the Java driver in our Fundamentals
+CRUD guide or how to perform common operations from our
```
ccho-mongodb, line 2: When the CRUD guide is created, this will be updated to link to it.

source/includes/quick-start/atlas-setup.rst
```diff
@@ -0,0 +1,41 @@
...
+**Clusters** section and click the **Connect** button for the cluster that you
+want to connect to as shown below.
+
+.. figure:: /includes/figures/atlas_connection_select_cluster.png
+
+Proceed to the **Connect Your Application** step and select the Java driver.
+Then, select the "Connection String Only" tab and click the **Copy**
+button to copy the *connection string* to your clipboard as shown below.
+
+.. figure:: /includes/figures/atlas_connection_copy_string.png
```
terakilobyte, line 38: Image isn't showing

source/quick-start-legacy.txt
```diff
@@ -0,0 +1,112 @@
...
+   import com.mongodb.MongoClientURI;
+   import com.mongodb.client.MongoCollection;
+   import com.mongodb.client.MongoDatabase;
+
+Next, add the following method, replacing the value of the ``uri`` variable
+with your MongoDB Atlas connection string. Make sure to replace the
+"<password>" section of the connection string with the password you created for
+your user that has **atlasAdmin** permissions:
+
+.. code-block:: java
```
terakilobyte, line 84:
```suggestion
.. code-block:: java

```

terakilobyte: Without the space it's eating the method declaration in render

source/quick-start.txt
```diff
@@ -7,7 +7,109 @@ Quick Start
...
+   import com.mongodb.client.MongoClients;
+   import com.mongodb.client.MongoCollection;
+   import com.mongodb.client.MongoDatabase;
+
+Next, add the following method, replacing the value of the ``uri`` variable
+with your MongoDB Atlas connection string. Make sure to replace the
+"<password>" section of the connection string with the password you created for
+your user that has **atlasAdmin** permissions:
+
+.. code-block:: java
```
terakilobyte, line 90:
```suggestion
.. code-block:: java

```

source/quick-start-legacy.txt
```diff
@@ -0,0 +1,112 @@
...
+Add the following imports for this example.
+
+.. code-block:: java
+
+   import org.bson.Document;
+
+   import com.mongodb.MongoClient;
+   import com.mongodb.MongoClientURI;
+   import com.mongodb.client.MongoCollection;
+   import com.mongodb.client.MongoDatabase;
```
terakilobyte, line 77:
```suggestion
   import com.mongodb.client.MongoDatabase;
   
   import static com.mongodb.client.model.Filters.eq;
```

terakilobyte: I had to add the bottom import

source/includes/quick-start/atlas-setup.rst
ccho-mongodb: File was in the wrong place, fixed in 2ee804f400ce3d2cc43058a0054b78f456f8445a

ccho-mongodb:
> What about just adding the entire contents of Quickstart.java in one shot rather than breaking up deps from the main method? This way it's easier to copy/paste.

Good idea, will update.

source/quick-start-legacy.txt
```diff
@@ -0,0 +1,115 @@
...
+
+.. code-block:: xml
+
+   <dependencies>
+       <dependency>
+           <groupId>org.mongodb</groupId>
+           <artifactId>mongodb-driver-legacy</artifactId>
+           <version>4.0.1</version>
+       </dependency>
+   </dependencies>
```
terakilobyte, line 49:
```suggestion
   </dependencies>
       <properties>
        <maven.compiler.source>1.9</maven.compiler.source>
        <maven.compiler.target>1.9</maven.compiler.target>
    </properties>
```

terakilobyte: I had to specify this as maven defaults to 1.5, or at least it did on my machine.

source/quick-start.txt
```diff
@@ -7,7 +7,110 @@ Quick Start
...
+
+.. code-block:: xml
+
+   <dependencies>
+       <dependency>
+           <groupId>org.mongodb</groupId>
+           <artifactId>mongodb-driver-sync</artifactId>
+           <version>4.0.1</version>
+       </dependency>
+   </dependencies>
```
terakilobyte, line 48:
```suggestion
   </dependencies>
       <properties>
        <maven.compiler.source>1.9</maven.compiler.source>
        <maven.compiler.target>1.9</maven.compiler.target>
    </properties>
```

source/includes/quick-start/create-project.rst
```diff
@@ -0,0 +1,11 @@
+This guide shows you how add the MongoDB Java driver dependencies using
+Maven or Gradle. We recommend that you use an integrated development
+environment (IDE) such as Intellij IDEA, Eclipse IDE, or NetBeans Java IDE to
```
jyemin, line 3: According to https://www.jetbrains.com/lp/devecosystem-2019/java/, NetBeans has lost.  Let's not even mention it.

source/includes/quick-start/overview.rst
```diff
@@ -0,0 +1,14 @@
+This guide shows you how to create an application that uses the **Java driver**
+to connect to a **MongoDB Atlas cluster**. If you prefer to connect to MongoDB
+using a different driver or programming language, see our
+:driver:`list of official MongoDB drivers <>`.
+
+The Java driver is an interface through which you can connect to and
```
jyemin, line 6: It's an interface, yes, but it's the implementation of the interface that allows you to connect.  Maybe just say:

        The Java driver lets you connect to and communicate with ... from a Java application.

or something to that effect

source/quick-start-legacy.txt
```diff
@@ -0,0 +1,127 @@
+===========================
```
jyemin, line 1: For a quick start, do we even need to talk about the legacy API at all?  If someone is using legacy, it's only because they already have an existing application that uses it, and therefore doesn't need a quick start guide in the first place.

source/quick-start.txt
```diff
@@ -7,7 +7,122 @@ Quick Start
...
+
+   For example, if you build with Maven in the IntelliJ IDE, you need to
+   specify your JDK version explicitly in the ``properties`` section in your
+   ``pom.xml`` file. The following example instructs Maven to use JDK
+   version 1.9 (Java 9):
+
+   .. code-block:: xml
+
+      <properties>
+          <maven.compiler.source>1.9</maven.compiler.source>
```
jyemin, line 72: I'm not sure you should have to do this.  Will have to check it.

source/quick-start-legacy.txt
ccho-mongodb: I considered that, and was thinking of the situation in which a developer is introduced to an existing application/service that uses the legacy API and wants to learn from the ground up. I can remove it if you think it's not necessary.

I am curious what your opinion is on how much information about the legacy API that you wanted to keep.
 
For example, in the Usage Examples section, should we offer examples using both the current and legacy API, or add a note that mentions the difference in the MongoClient classes (along with any other differences)?

source/quick-start.txt
ccho-mongodb: It's not required in Eclipse, but IntelliJ IDEA has some strange behavior of defaulting to 1.5 if not specified as Nathan discovered.
https://youtrack.jetbrains.com/issue/IDEA-215058?_ga=2.114859969.797257447.1586200689-1373391973.1586200689

source/quick-start-legacy.txt
jyemin: > was thinking of the situation in which a developer is introduced to an existing application/service that uses the legacy API and wants to learn from the ground up

I think it's fine in this case to learn the new API.  If the application is existing already, generally they are not going to be having to construct a MongoClient themselves anyway.  It will already be created somewhere and stashed in a global variable.   So I think you can remove.

> I am curious what your opinion is on how much information about the legacy API that you wanted to keep.

For the reference documentation, I think we should continue to give examples using both the new and old MongoClient, since the old is still widely used.

source/quick-start.txt
jyemin: I don't think it's enough just to set the property values though.  You would need to use them in a block like:

    <build>
        <plugins>
            <plugin>
                <groupId>org.apache.maven.plugins</groupId>
                <artifactId>maven-compiler-plugin</artifactId>
                <version>3.8.1</version>
                <configuration>
                    <release>8</release>
                    <source>8</source>
                    <target>8</target>
                </configuration>
            </plugin>
        </plugins>
    </build>

jyemin: OK, I see it too.  Let's specify 1.8 though since it's the minimum version required by the driver and still most popular.

source/includes/quick-start/overview.rst
```diff
@@ -0,0 +1,14 @@
+This guide shows you how to create an application that uses the **Java driver**
+to connect to a **MongoDB Atlas cluster**. If you prefer to connect to MongoDB
+using a different driver or programming language, see our
+:driver:`list of official MongoDB drivers <>`.
+
+The Java driver lets you connect to and communicate with MongoDB instances
```
jyemin, line 6: Nit: is `instances` the word we use in general?  I would think `cluster` or `server`.

ccho-mongodb: I don't think we've decided on a term across docs teams and we refer to them in several different ways. I can use cluster across docs-java if you'd prefer.

In docs-node, we used "instances" much more often than cluster, server, or replica set.

ccho-mongodb: Let me know what you think of this change from "instance(s)" to "cluster" https://github.com/mongodb/docs-java/pull/3/commits/3693d5e28b8322bbab51e11da2db4d20cd4f64c6

PR #8, opened 2020-04-24
DOCSP-9593: Usage Example - insertMany

ccho-mongodb:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-9593

https://docs-mongodbcom-staging.corp.mongodb.com/58ca7b6/java/docsworker-xlarge/DOCSP-9593-insert-many/usage-examples/insertMany


ccho-mongodb:
There appears to be some additional div with spacing added in the latest autobuilder, so disregard the additional whitespace around headers for now.

source/usage-examples/insertMany.txt
```diff
@@ -0,0 +1,54 @@
...
+Insert Multiple Documents
+=========================
+
+.. default-domain:: mongodb
+
+You can insert a multiple documents into a collection in a single
+operation by calling the ``insertMany()`` method on a ``MongoCollection``
+object. To insert them, add your ``Document`` objects that contain the
+fields and values that you want to store to a ``List`` and pass it to
+your call to ``insertMany()``. If you call the ``insertMany()`` method on
```
nathan-contino-mongo, line 11: The second sentence of this paragraph is a little hard to parse -- can we simplify it at all? I think the "objects that contain the fields and values that you want to store" phrase is a little too wordy. Can we just say "objects" instead? I think we should also change "to a ``List``" into "into a ``List``".

ccho-mongodb: I'll simplify by omitting "fields and values that you want to store" -- I added this to make it more specific, but I guess it could be too verbose.

I'm not sure whether there's anything incorrect about "add to", but "add into" is an idiom which we should avoid.

source/includes/usage-examples/code-snippets/InsertMany.java
```diff
@@ -0,0 +1,39 @@
+// ignored first line
+package usage.examples;
```
jyemin, line 2: At some point we should fix up the directory structure here such that we can easily compile this code.

source/usage-examples/insertMany.txt
```diff
@@ -0,0 +1,53 @@
+=========================
+Insert Multiple Documents
+=========================
+
+.. default-domain:: mongodb
+
+You can insert a multiple documents into a collection in a single
```
jyemin, line 7: insert a multiple documents => insert multiple documents

source/includes/usage-examples/code-snippets/InsertMany.java
ccho-mongodb: Sure, I can change them now if you let me know what the package name should be for automated testing.

jyemin: The package name can be anything.  It's more that the source code has to be in a structure that is consumable by Maven or Gradle (i.e. for this file in `src/main/java/usage/examples`)

ccho-mongodb: Oh I see. We'll probably need a script to copy these over to where they can be built, but I'll think about a different parent directory name to implement in a future PR.

PR #13, opened 2020-06-17
DOCSP-9595: Update Multiple usage example

ccho-mongodb:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-9595

https://docs-mongodbcom-staging.corp.mongodb.com/6a98491/java/docsworker-xlarge/DOCSP-9595-update-multiple/usage-examples/updateMany

### DIFF
--- a/source/includes/usage-examples/code-snippets/UpdateMany.java
+++ b/source/includes/usage-examples/code-snippets/UpdateMany.java
@@ -1,0 +1,44 @@
+// ignored first line
+package usage.examples;
+
+import static com.mongodb.client.model.Filters.gt;
+
+import org.bson.Document;
+import org.bson.conversions.Bson;
+
+import com.mongodb.MongoException;
+import com.mongodb.client.MongoClient;
+import com.mongodb.client.MongoClients;
+import com.mongodb.client.MongoCollection;
+import com.mongodb.client.MongoDatabase;
+import com.mongodb.client.model.Updates;
+import com.mongodb.client.result.UpdateResult;
+
+public class UpdateMany {
+
+    public static void main(String[] args) {
+        // Replace the uri string with your MongoDB deployment's connection string
+        String uri = "mongodb+srv://<user>:<password>@<cluster-url>?retryWrites=true&w=majority";
+
+        try (MongoClient mongoClient = MongoClients.create(uri)) {
+
+            MongoDatabase database = mongoClient.getDatabase("sample_mflix");
+            MongoCollection<Document> collection = database.getCollection("movies");
+
+            Bson query = gt("num_mflix_comments", 50);
+
+            Bson updates = Updates.combine(
+                    Updates.addToSet("genres", "Frequently Discussed"),
+                    Updates.currentTimestamp("lastUpdated"));
+
+            try {
+                UpdateResult result = collection.updateMany(query, updates);
+
+                System.out.println("Modified document count: " + result.getModifiedCount());
+
+            } catch (MongoException me) {
+                System.err.println("Unable to update due to an error: " + me);
+            }
+        }
+    }
+}
--- a/source/usage-examples/updateMany.txt
+++ b/source/usage-examples/updateMany.txt
@@ -1,0 +1,114 @@
+===========================
+Update a Multiple Documents
+===========================
+
+.. default-domain:: mongodb
+
+You can update multiple documents using the ``updateMany()`` method on
+a ``MongoCollection`` object. The method accepts a **filter** that matches the
+document you want to update and an **update** statement that instructs the
+driver to change the matching document. The ``updateMany()`` method only
+updates the first document that matches the filter.
+
+To perform an update with the ``updateMany()`` method, you must pass
+a query filter and an update document. The query filter specifies the criteria
+for which document to perform the update on and the update document provides
+instructions on what changes to make to them.
+
+You can optionally pass an instance of ``UpdateOptions`` to the method in
+order to specify the behavior. For example, if you set the ``upsert`` field of
+the ``UpdateOptions`` object to ``true`` and no documents match the
+specified query filter, the operation inserts a new document composed of
+the fields from both the query and update document. See the link to the
+``UpdateOptions`` API documentation at the bottom of this page for more
+information.
+
+Upon successful execution, the ``updateMany()`` method returns an instance
+of ``UpdateResult``.  You can retrieve information such as the number of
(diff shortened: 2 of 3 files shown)


source/includes/usage-examples/code-snippets/UpdateMany.java
```diff
@@ -0,0 +1,44 @@
+// ignored first line
```
nathan-contino-mongo, line 1: apparently this ignored first line behaviour is going to disappear soon, I can't find the platform ticket for it though.

source/usage-examples/updateMany.txt
```diff
@@ -0,0 +1,113 @@
...
+Update a Multiple Documents
+===========================
+
+.. default-domain:: mongodb
+
+You can update multiple documents using the ``updateMany()`` method on
+a ``MongoCollection`` object. The method accepts a **filter** that matches the
+document you want to update and an **update** statement that instructs the
+driver to change the matching document. The ``updateMany()`` method only
+updates all the documents in the collection that match the filter.
```
nathan-contino-mongo, line 11: personally I feel that the "only...all" construction here is a little hard to parse. Do you think we could remove "only", or is there a good reason for it to remain here?

source/usage-examples/updateMany.txt
```diff
@@ -0,0 +1,113 @@
...
+
+To perform an update with the ``updateMany()`` method, you must pass
```
nathan-contino-mongo, line 13:
```suggestion
To perform an update with the ``updateMany()`` method, you must provide
```
To me, "provide" implies that you should construct these yourself, while "pass" implies that these parameters might already exist and you just have to pass them along. But I admit that it's entirely subjective.

source/usage-examples/updateMany.txt
```diff
@@ -0,0 +1,113 @@
...
+a query filter and an update document. The query filter specifies the criteria
+for which document to perform the update on and the update document provides
+instructions on what changes to make to them.
```
nathan-contino-mongo, line 16: This sentence is a little bit of a mouthful for me, would it make sense to break this up into two sentences, each explaining a single parameter? "which document to perform the update on" also implies that this method only updates a single document -- can we pluralize that?

source/usage-examples/updateMany.txt
```diff
@@ -0,0 +1,113 @@
...
+
+You can optionally pass an instance of ``UpdateOptions`` to the method in
+order to specify the behavior. For example, if you set the ``upsert`` field of
```
nathan-contino-mongo, line 19:
```suggestion
order to adjust behavior. For example, if you set the ``upsert`` field of
```
I'm not sure if this sounds any better, but I definitely think we should at least ditch the "the" before behavior.

source/usage-examples/updateMany.txt
```diff
@@ -0,0 +1,113 @@
...
+Upon successful execution, the ``updateMany()`` method returns an instance
+of ``UpdateResult``. You can retrieve information such as the number of
+documents modified calling the ``getModifiedCount()`` method. If you
+specified ``upsert(true)`` in the ``UpdateOptions`` object and the
+operation results in an insert, you can retrieve the ``_id`` field of the
+new document by calling the ``getUpsertedId()`` method on the
+``UpdateResult``.
+
+If your update operation fails, the driver raises an exception and none of
+the documents matching the filter are updated. For example, if you try to set
```
nathan-contino-mongo, line 33: "none of the documents matching the filter are update"
Is this strictly true? For instance, if you update half the documents, then violate a unique field constraint for a single document halfway through the update, wouldn't some of the documents (the first half) stay updated?

It might be a good idea to avoid mentioning the side-effects of failed operations entirely since that's likely highly dependent on configuration and when the error occurs. Though I'll grant that in the case you provided, none of the documents would update.

source/usage-examples/updateMany.txt
```diff
@@ -0,0 +1,113 @@
...
+- Add ``Frequently Discussed`` to the array of ``genres`` only if it does not
+  already exist
+- Set the value of ``lastUpdated`` to the current time.
+
+We use the ``Updates`` builder, a factory class that contains static
+helper methods to construct the update document. While you can pass an update
+document instead of using the builder, the builder provides type checking and
+simplified syntax. Read our
+:doc:`guide on Updates </fundamentals/builders/updates>` in the Builders
+section for more information.
```
nathan-contino-mongo, line 70: Love this little bit of background on the ``Updates`` builder, this is a great way to pull slightly more advanced concepts into the usage examples!

source/usage-examples/updateMany.txt
ccho-mongodb: As far as I've been able to test it, yes. This might be different from "bulkWrite" operations. Will double check with Jeff.

source/includes/usage-examples/code-snippets/UpdateMany.java
ccho-mongodb: Release for this feature is tonight, so I'll create a separate PR to address all the issues here and other driver repos.

source/usage-examples/updateMany.txt
ccho-mongodb: Whoops, good catch. Will fix.

source/usage-examples/updateMany.txt
ccho-mongodb: They imply the same thing to me and both require qualifiers to discern whether it's a pre-existing object or if you're declaring them inline. In my own experience, I have encountered "pass parameters" more often than "provide parameters" in documentation and other forms of communication.

source/usage-examples/updateMany.txt
ccho-mongodb: Will address.

source/usage-examples/updateMany.txt
ccho-mongodb: I think I accidentally left "document" in there, so it doesn't belong there. I'll update to condense the sentence.

source/usage-examples/updateMany.txt
nathan-contino-mongo: Seems reasonable, more common = clearer for our readers.

source/includes/usage-examples/code-snippets/UpdateMany.java
nathan-contino-mongo: nice, thanks for staying on top of that!

PR #19, opened 2020-07-08
(DOCSP-9601): Usage Examples > Run a Command

nathan-contino-mongo:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-9601

https://docs-mongodbcom-staging.corp.mongodb.com/18cb1fc/java/docsworker-xlarge/DOCSP-9601/usage-examples/command

### DIFF
--- a/source/includes/usage-examples/code-snippets/Command.java
+++ b/source/includes/usage-examples/code-snippets/Command.java
@@ -22,7 +22,6 @@ public class RunCommand {
             MongoDatabase database = mongoClient.getDatabase("sample_mflix");
 
             try {
-                // find the storage statistics for the "sample_mflix" database using the 'dbStats' command
                 Bson command = new BsonDocument("dbStats", new BsonInt64(1));
                 Document commandResult = database.runCommand(command);
                 System.out.println("dbStats: " + commandResult.toJson());
--- a/source/includes/usage-examples/code-snippets/Command.java
+++ b/source/includes/usage-examples/code-snippets/Command.java
@@ -1,0 +1,34 @@
+package usage.examples;
+
+import org.bson.BsonDocument;
+import org.bson.BsonInt64;
+import org.bson.Document;
+
+import com.mongodb.ConnectionString;
+import com.mongodb.MongoClientSettings;
+import com.mongodb.MongoException;
+import com.mongodb.client.MongoClient;
+import com.mongodb.client.MongoClients;
+import com.mongodb.client.MongoDatabase;
+
+
+public class RunCommand {
+    public static void main(String[] args) {
+        // Replace the uri string with your MongoDB deployment's connection string
+        String uri = "mongodb+srv://<user>:<password>@<cluster-url>?retryWrites=true&w=majority";
+
+        try (MongoClient mongoClient = MongoClients.create(uri)) {
+
+            MongoDatabase database = mongoClient.getDatabase("sample_mflix");
+
+            try {
+                // find the storage statistics for the "sample_mflix" database using the 'dbStats' command
+                Bson command = new BsonDocument("dbStats", new BsonInt64(1));
+                Document commandResult = database.runCommand(command);
+                System.out.println("dbStats: " + commandResult.toJson());
+            } catch (MongoException me) {
+                System.err.println("An error occurred: " + me);
+            }
+        }
+    }
+}
--- a/source/usage-examples/command.txt
+++ b/source/usage-examples/command.txt
@@ -6,11 +6,11 @@ Run a Command
 
 You can run all raw database operations using the
 ``MongoDatabase.runCommand()`` method. A raw database operation is a
-command you can execute directly on the MongoDB server CLI. Commands
-include administrative and diagnostic tasks, such as fetching server
-stats or initializing a replica set. Call the ``runCommand()`` method
-with a Bson command object on an instance of a database to run your raw
-database operation.
+command you can execute directly on the MongoDB server CLI. These
+commands include administrative and diagnostic tasks, such as fetching
+server stats or initializing a replica set. Call the ``runCommand()``
+method with a Bson command object on an instance of a ``MongoDatabase``
+to run your raw database operation.
 
 .. note::
     Use the :manual:`mongo shell </reference/mongo-shell>` for
@@ -24,6 +24,9 @@ parameter.
 
 Example
 -------
+
+In the following sample code, we send the ``dbStats`` command to request
+statistics of a specific MongoDB database.
 
 .. include:: /includes/connect-guide-note.rst
 
@@ -43,4 +46,4 @@ page, see the following API documentation:
 
 - :java-sync-api:`runCommand <com/mongodb/client/MongoDatabase.html#runCommand(org.bson.conversions.Bson)>`
 - :manual:`command <reference/command/>`
-- :manual:`read preference <core/read-preference/>`
+- :manual:`dbStats <reference/command/dbStats/>`
--- a/source/usage-examples/command.txt
+++ b/source/usage-examples/command.txt
@@ -26,7 +26,7 @@ Example
 -------
 
 In the following sample code, we send the ``dbStats`` command to request
-statistics of a specific MongoDB database.
+statistics from a specific MongoDB database.
 
 .. include:: /includes/connect-guide-note.rst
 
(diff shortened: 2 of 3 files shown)


nathan-contino-mongo:
For some reason the staging link is a 404 right now.

source/includes/usage-examples/code-snippets/Command.java
```diff
@@ -0,0 +1,34 @@
...
+    public static void main(String[] args) {
+        // Replace the uri string with your MongoDB deployment's connection string
+        String uri = "mongodb+srv://<user>:<password>@<cluster-url>?retryWrites=true&w=majority";
+
+        try (MongoClient mongoClient = MongoClients.create(uri)) {
+
+            MongoDatabase database = mongoClient.getDatabase("sample_mflix");
+
+            try {
+                // find the storage statistics for the "sample_mflix" database using the 'dbStats' command
```
ccho-mongodb, line 25: Nitpick: Other than the URI replacement, let's avoid using comments in the code since our driver docs user studies showed that readers are more likely to read text than comments in code. I think I've followed this in the other usage examples, but if you find an instance, I'll fix it.

source/usage-examples/command.txt
```diff
@@ -0,0 +1,48 @@
...
+
+You can run all raw database operations using the
+``MongoDatabase.runCommand()`` method. Call the ``runCommand()`` method
+with your command object on an instance of a database for diagnostic and
+administrative tasks such as fetching server stats or initializing a
+replica set.
+
+.. note::
+    Use the :manual:`mongo shell </reference/mongo-shell>` for
+    administrative tasks instead of the Java Driver whenever possible.
```
ccho-mongodb, line 15: Question: should this be "Java driver" or is "Java Driver" a specific proper noun? I think it's the former since there isn't just a single one in existence.

source/usage-examples/command.txt
```diff
@@ -0,0 +1,48 @@
+=============
+Run a Command
+=============
+
+.. default-domain:: mongodb
+
+You can run all raw database operations using the
```
ccho-mongodb, line 7: Although I know this resembles the Node usage example, I'm currently wondering whether it appropriately addresses the Jasmine persona.  If we re-order some of the information, I think we could improve the learning experience. E.g.

"You can run all raw database operations ... method. A raw database operation is a command you can execute directly on the MongoDB server CLI and includes administrative and diagnostic commands such as fetching server stats or initializing a replica set. Call the ``runCommand()`` method with a Bson command object on an instance of a database to run your raw database operation."

source/usage-examples/command.txt
```diff
@@ -0,0 +1,48 @@
...
+with your command object on an instance of a database for diagnostic and
+administrative tasks such as fetching server stats or initializing a
+replica set.
+
+.. note::
+    Use the :manual:`mongo shell </reference/mongo-shell>` for
+    administrative tasks instead of the Java Driver whenever possible.
+
+The ``runCommand()`` method accepts a :manual:`command
+<reference/command/>` in Bson format. By default, ``runCommand`` returns
```
ccho-mongodb, line 18: Let's not link to the manual until the end as this is "external" to the Java driver docs. Instead mention "for more information on commands, see the link at the bottom of this page."

source/usage-examples/command.txt
```diff
@@ -0,0 +1,48 @@
...
+an object of type ``org.bson.Document`` containing the output of the
+database command.
+
+You can specify a :manual:`read preference <core/read-preference/>` or
```
ccho-mongodb, line 22: Same comment as above. Also, unless you think many of the commands they will run require read preference, let's omit it from the usage example. I think the current paragraph is also missing the ordering if they do not want a read preference but want a return type.

source/usage-examples/command.txt
```diff
@@ -0,0 +1,48 @@
...
+parameter.
+
+Example
+-------
+
+.. include:: /includes/connect-guide-note.rst
+
+.. literalinclude:: /includes/usage-examples/code-snippets/Command.java
+  :language: java
+  :linenos:
```
ccho-mongodb, line 35: I know some team is working on the "linenos" directive, but for now let's omit this.

source/usage-examples/command.txt
nathan-contino-mongo: Will try out a couple of different orders, see what sticks.

source/usage-examples/command.txt
nathan-contino-mongo: "you can specify a read preference or a return type ... as an optional second parameter" covers both read preference and return type only, but I can trim this to just cover return type. Only issue I have there is that I hate including just *one* of the arguments arbitrarily, but I think you have a good point that read preference is not something that most users need to know about in this context.

source/usage-examples/command.txt
```diff
@@ -0,0 +1,46 @@
+=============
+Run a Command
+=============
+
+.. default-domain:: mongodb
+
+You can run all raw database operations using the
+``MongoDatabase.runCommand()`` method. A raw database operation is a
+command you can execute directly on the MongoDB server CLI. Commands
```
ccho-mongodb, line 9: Nitpick: would prefer specificity on what commands you're referring to in "Commands include", e.g. "These commands include..."

source/usage-examples/command.txt
```diff
@@ -0,0 +1,46 @@
...
+include administrative and diagnostic tasks, such as fetching server
+stats or initializing a replica set. Call the ``runCommand()`` method
+with a Bson command object on an instance of a database to run your raw
```
ccho-mongodb, line 12: Nitpick: should've picked this up before, but since we're specifying types, how about ``MongoDatabase`` instead of database to make it clear for a first-time MongoDB user?

source/usage-examples/command.txt
```diff
@@ -0,0 +1,46 @@
...
+.. code-block:: none
+
+   dbStats: {"db": "sample_mflix", "collections": 5, "views": 0, "objects": 75595, "avgObjSize": 692.1003770090614, "dataSize": 52319328, "storageSize": 29831168, "numExtents": 0, "indexes": 9, "indexSize": 14430208, "fileSize": 0, "nsSizeMB": 0, "ok": 1}
+
+For additional information on the classes and methods mentioned on this
+page, see the following API documentation:
+
+- :java-sync-api:`runCommand <com/mongodb/client/MongoDatabase.html#runCommand(org.bson.conversions.Bson)>`
+- :manual:`command <reference/command/>`
+- :manual:`read preference <core/read-preference/>`
```
ccho-mongodb, line 46: Nitpick: since you removed mention of the read preference, we should probably omit it here.

source/usage-examples/command.txt
```diff
@@ -0,0 +1,46 @@
...
+
+The ``runCommand()`` method accepts a command in Bson format.
+By default, ``runCommand`` returns an object of type
+``org.bson.Document`` containing the output of the database command. You
+can specify a return type for ``runCommand()`` as an optional second
+parameter.
+
+Example
+-------
+
```
ccho-mongodb, line 27: I think it would be useful to add a description of the code snippet here:

In the following sample code, we send the ``dbStats`` command to request statistics of a specific MongoDB database.

(and include dbStats in the list of links at the end)

nathan-contino-mongo:
Staging link fixed by switching to xlarge worker :disappointed_relieved:

source/usage-examples/command.txt
```diff
@@ -0,0 +1,49 @@
...
+``MongoDatabase.runCommand()`` method. A raw database operation is a
+command you can execute directly on the MongoDB server CLI. These
+commands include administrative and diagnostic tasks, such as fetching
+server stats or initializing a replica set. Call the ``runCommand()``
+method with a Bson command object on an instance of a ``MongoDatabase``
+to run your raw database operation.
+
+.. note::
+    Use the :manual:`mongo shell </reference/mongo-shell>` for
+    administrative tasks instead of the Java driver whenever possible.
```
rozza, line 17: Suggestion: It might be good to include why?  They amount to the same thing, its just the mongo shell is often quicker and easier.

nathan-contino-mongo: Done, thanks for the suggestion! I imagine a lot of new users would question why, so this is definitely a good addition.

PR #21, opened 2020-07-15
(DOCSP-9602): Usage Example -> Watch

nathan-contino-mongo:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-9602

https://docs-mongodbcom-staging.corp.mongodb.com/c89c9ff/java/docsworker-xlarge/DOCSP-9602/usage-examples/watch

### DIFF
--- a/source/includes/usage-examples/code-snippets/Watch.java
+++ b/source/includes/usage-examples/code-snippets/Watch.java
@@ -1,0 +1,28 @@
+package usage.examples;
+
+import org.bson.Document;
+
+import com.mongodb.client.ChangeStreamIterable;
+import com.mongodb.client.MongoClient;
+import com.mongodb.client.MongoClients;
+import com.mongodb.client.MongoCollection;
+import com.mongodb.client.MongoDatabase;
+import com.mongodb.client.model.changestream.FullDocument;
+
+public class Watch {
+    public static void main( String[] args ) {
+
+        // Replace the uri string with your MongoDB deployment's connection string
+        String uri = "mongodb+srv://<user>:<password>@<cluster-url>?retryWrites=true&w=majority";
+
+        try (MongoClient mongoClient = MongoClients.create(uri)) {
+
+            MongoDatabase database = mongoClient.getDatabase("sample_mflix");
+            MongoCollection<Document> collection = database.getCollection("movies");
+
+            ChangeStreamIterable<Document> changeStream = database.watch()
+                .fullDocument(FullDocument.UPDATE_LOOKUP);
+            changeStream.forEach(event -> System.out.println("received a change to the collection: " + event));
+        }
+    }
+}
--- a/source/includes/usage-examples/code-snippets/WatchCompanion.java
+++ b/source/includes/usage-examples/code-snippets/WatchCompanion.java
@@ -1,0 +1,34 @@
+package usage.examples;
+
+import java.util.Arrays;
+
+import org.bson.Document;
+import org.bson.types.ObjectId;
+
+import com.mongodb.MongoException;
+import com.mongodb.client.MongoClient;
+import com.mongodb.client.MongoClients;
+import com.mongodb.client.MongoCollection;
+import com.mongodb.client.MongoDatabase;
+import com.mongodb.client.result.InsertOneResult;
+
+public class WatchCompanion {
+    public static void main(String[] args) {
+        // Replace the uri string with your MongoDB deployment's connection string
+        String uri = "mongodb+srv://<user>:<password>@<cluster-url>?retryWrites=true&w=majority";
+
+        try (MongoClient mongoClient = MongoClients.create(uri)) {
+
+            MongoDatabase database = mongoClient.getDatabase("sample_mflix");
+            MongoCollection<Document> collection = database.getCollection("movies");
+
+            try {
+                InsertOneResult result = collection.insertOne(new Document("test", "sample movie document"));
+
+                System.out.println("Success! Inserted document id: " + result.getInsertedId());
+            } catch (MongoException me) {
+                System.err.println("Unable to insert due to an error: " + me);
+            }
+        }
+    }
+}
--- a/source/usage-examples/watch.txt
+++ b/source/usage-examples/watch.txt
@@ -1,0 +1,139 @@
+=================
+Watch for Changes
+=================
+
+.. default-domain:: mongodb
+
+Open a Change Stream
+--------------------
+
+You can keep track of changes to data in MongoDB, such as changes to a
+collection, database, or deployment, by opening a **change stream**. A change
+stream allows applications to watch for changes to data and react to them.
+You can open a change stream by calling the ``watch()`` method on
+a ``MongoCollection``, ``MongoDatabase``, or ``MongoClient`` object. The
+change stream returns **change event** documents when they occur.
+
+The ``watch()`` method optionally takes an **aggregation pipeline**  which
+consists of an array of **stages** as the first parameter to filter and
+transform the change event output as follows:. 
+
+.. code-block:: java
+
+   List<Bson> pipeline = Arrays.asList(
+                           Aggregates.match(
+                              Filters.lt("fullDocument.runtime", 15)));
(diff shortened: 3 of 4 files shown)


source/usage-examples/watch.txt
```diff
@@ -0,0 +1,139 @@
...
+                           Aggregates.match(
+                              Filters.lt("fullDocument.runtime", 15)));
+   ChangeStreamIterable<Document> changeStream = database.watch(pipeline);
+
+To configure options for the change stream, use member methods of the
+``ChangeStreamIterable`` object returned by ``watch()``. Refer to the
+links at the end of this section for more information on the settings
+you can configure.
+
+The ``watch()`` method returns an instance of a ``ChangeStreamIterable``.
```
ccho-mongodb, line 33: Not mandatory, but I wonder if it might be helpful to describe the class in more detail. For example, here's what I did for MongoIterable, guided by advice from Jeff:
https://github.com/mongodb/docs-java/blob/master/source/usage-examples/find.txt#L21

source/usage-examples/watch.txt
```diff
@@ -0,0 +1,139 @@
+=================
+Watch for Changes
+=================
+
+.. default-domain:: mongodb
+
+Open a Change Stream
```
ccho-mongodb, line 7: Nitpick: feel free to disagree, but when using the imperative form in these headings, it reads to me more as a procedural tutorial (which you might find in the Fundamentals section) than a usage example. Since the other Usage Examples use an approach just to list information, I wonder if it might be better to adhere to that format.

source/usage-examples/watch.txt
```diff
@@ -0,0 +1,139 @@
...
+- :manual:`change streams </changeStreams/>`
+- :manual:`change events </reference/change-events/>`
+- :manual:`aggregation pipeline </reference/operator/aggregation-pipeline/>`
+- :manual:`aggregation stages </changeStreams/#modify-change-stream-output>`
+- :java-sync-api:`ChangeStreamIterable <com/mongodb/client/ChangeStreamIterable.html>`
+- :java-sync-api:`MongoCollection.watch() <com/mongodb/client/MongoCollection.html#watch()>`
+- :java-sync-api:`MongoDatabase.watch() <com/mongodb/client/MongoDatabase.html#watch()>`
+- :java-sync-api:`MongoClient.watch() <com/mongodb/client/MongoClient.html#watch()>`
+
+Process the Change Stream Events
```
ccho-mongodb, line 54: Nitpick: Similar comment to the one under the "Open a Change Stream" heading. I like the material itself, but perhaps it could be labeled as "How to Process Change Stream Events with a Callback", or just be rolled up into the Example. Alternatively, we could move it to Fundamentals content on change streams. It might not be the right material for Jasmin, but definitely appropriate for Lucas.

source/usage-examples/watch.txt
```diff
@@ -0,0 +1,139 @@
...
+   Calls to ``forEach()`` block the current thread as long as the
+   corresponding change stream listens for events. If your program
+   needs to continue executing other logic, such as processing requests or
+   responding to user input, consider creating and listening to your
+   change stream in a separate thread.
+
+.. note::
+
+   For update operation change events, change streams only return the modified
+   fields by default rather than the full updated document. You can configure
```
ccho-mongodb, line 80: Nitpick: how about "complete" instead of "full"? It's a little confusing whether you mean "complete" or "fully-updated" with that word combo.

source/usage-examples/watch.txt
```diff
@@ -0,0 +1,139 @@
...
+      ChangeStreamIterable<Document> changeStream = database.watch()
+                .fullDocument(FullDocument.UPDATE_LOOKUP);
+
+Example
+-------
+
+The following example uses two separate applications to demonstrate how
+to listen for changes using a change stream:
+
+- The first application, named ``Watch``, opens a change stream
```
ccho-mongodb, line 96: Nitpick: since you labeled them by filename (e.g. "Watch.java"), perhaps you should refer to them that way, or alternatively label them below with Watch and WatchCompanion rather than filename.

source/includes/usage-examples/code-snippets/Watch.java
```diff
@@ -0,0 +1,28 @@
...
+
+        // Replace the uri string with your MongoDB deployment's connection string
+        String uri = "mongodb+srv://<user>:<password>@<cluster-url>?retryWrites=true&w=majority";
+
+        try (MongoClient mongoClient = MongoClients.create(uri)) {
+
+            MongoDatabase database = mongoClient.getDatabase("sample_mflix");
+            MongoCollection<Document> collection = database.getCollection("movies");
+
+            ChangeStreamIterable<Document> changeStream = database.watch()
```
ccho-mongodb, line 23: Since we're seeking full document updates and we've prominently (my opinion) mentioned passing a pipeline, I wonder if we should also show how to limit to a type of operation using "operationType"in the pipeline. For examples, see: https://developer.mongodb.com/quickstart/java-change-streams

source/includes/usage-examples/code-snippets/Watch.java
```diff
@@ -0,0 +1,28 @@
...
+                .fullDocument(FullDocument.UPDATE_LOOKUP);
+            changeStream.forEach(event -> System.out.println("received a change to the collection: " + event));
```
ccho-mongodb, line 25: Nitpick: would it make sense to capitalize the start of a phrase here?

source/includes/usage-examples/code-snippets/WatchCompanion.java
```diff
@@ -0,0 +1,34 @@
...
+        // Replace the uri string with your MongoDB deployment's connection string
+        String uri = "mongodb+srv://<user>:<password>@<cluster-url>?retryWrites=true&w=majority";
+
+        try (MongoClient mongoClient = MongoClients.create(uri)) {
+
+            MongoDatabase database = mongoClient.getDatabase("sample_mflix");
+            MongoCollection<Document> collection = database.getCollection("movies");
+
+            try {
+                InsertOneResult result = collection.insertOne(new Document("test", "sample movie document"));
```
ccho-mongodb, line 26: I just noticed that we're doing an insert, but specifying an option for full document on updates in the other app. Perhaps we could leave out the option in Watch.java or change this to add an update after the insert so the user can observe the defined behavior?

source/usage-examples/watch.txt
```diff
@@ -0,0 +1,139 @@
...
+
+.. literalinclude:: /includes/usage-examples/code-snippets/Watch.java
+   :language: java
+
+``WatchCompanion.java``:
+
+.. literalinclude:: /includes/usage-examples/code-snippets/WatchCompanion.java
+   :language: java
+
+If you run the example above, you should see output similar to the
```
ccho-mongodb, line 122: Nitpick: "example" here is slightly ambiguous. Could we say something like this:
"If you run each of the applications in sequence, you should see output from the Watch application that is similar to the following:"
The WatchCompanion app also prints out some information, so you might want to include that here as well.

source/usage-examples/watch.txt
```diff
@@ -0,0 +1,139 @@
...
+
+To configure options for the change stream, use member methods of the
+``ChangeStreamIterable`` object returned by ``watch()``. Refer to the
+links at the end of this section for more information on the settings
+you can configure.
+
+The ``watch()`` method returns an instance of a ``ChangeStreamIterable``.
+You can call ``forEach()`` on the ``ChangeStreamIterable`` to handle
+events as they occur, or you can use the ``iterator()`` method to access
+the underlying ``MongoIterator``. This allows you to call manual methods
```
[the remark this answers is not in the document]
nathan-contino-mongo: Removed this mention.

source/usage-examples/watch.txt
nathan-contino-mongo: Considered removing both headings, but I think this section is too chunky without something to break it up. Taking your suggestion.

nathan-contino-mongo:
resumed here: https://github.com/mongodb/docs-java/pull/22 since github is broken and isn't updating PRs with new commits today.

PR #22, opened 2020-07-15
Docsp 9602

nathan-contino-mongo:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-9602

https://docs-mongodborg-staging.corp.mongodb.com/50be17a/java/docsworker-xlarge/DOCSP-9602/

### DIFF
--- a/source/includes/usage-examples/code-snippets/Watch.java
+++ b/source/includes/usage-examples/code-snippets/Watch.java
@@ -28,7 +28,14 @@ public class Watch {
                                 Arrays.asList("insert", "update"))));
             ChangeStreamIterable<Document> changeStream = database.watch(pipeline)
                 .fullDocument(FullDocument.UPDATE_LOOKUP);
-            changeStream.forEach(event -> System.out.println("Received a change to the collection: " + event));
+            // variables referenced in a lambda must be final; final array gives us a mutable integer
+            final int[] numberOfEvents = {0};
+            changeStream.forEach(event -> {
+            System.out.println("Received a change to the collection: " + event);
+                if (++numberOfEvents[0] >= 2) {
+                  System.exit(0);
+                }
+            });
         }
     }
 }
--- a/source/includes/usage-examples/code-snippets/Watch.java
+++ b/source/includes/usage-examples/code-snippets/Watch.java
@@ -1,0 +1,28 @@
+package usage.examples;
+
+import org.bson.Document;
+
+import com.mongodb.client.ChangeStreamIterable;
+import com.mongodb.client.MongoClient;
+import com.mongodb.client.MongoClients;
+import com.mongodb.client.MongoCollection;
+import com.mongodb.client.MongoDatabase;
+import com.mongodb.client.model.changestream.FullDocument;
+
+public class Watch {
+    public static void main( String[] args ) {
+
+        // Replace the uri string with your MongoDB deployment's connection string
+        String uri = "mongodb+srv://<user>:<password>@<cluster-url>?retryWrites=true&w=majority";
+
+        try (MongoClient mongoClient = MongoClients.create(uri)) {
+
+            MongoDatabase database = mongoClient.getDatabase("sample_mflix");
+            MongoCollection<Document> collection = database.getCollection("movies");
+
+            ChangeStreamIterable<Document> changeStream = database.watch()
+                .fullDocument(FullDocument.UPDATE_LOOKUP);
+            changeStream.forEach(event -> System.out.println("received a change to the collection: " + event));
+        }
+    }
+}
--- a/source/includes/usage-examples/code-snippets/Watch.java
+++ b/source/includes/usage-examples/code-snippets/Watch.java
@@ -1,12 +1,14 @@
 package usage.examples;
 
 import org.bson.Document;
+import org.bson.conversions.Bson;
 
 import com.mongodb.client.ChangeStreamIterable;
 import com.mongodb.client.MongoClient;
 import com.mongodb.client.MongoClients;
 import com.mongodb.client.MongoCollection;
 import com.mongodb.client.MongoDatabase;
+import com.mongodb.client.model.Filters;
 import com.mongodb.client.model.changestream.FullDocument;
 
 public class Watch {
@@ -20,9 +22,13 @@ public class Watch {
             MongoDatabase database = mongoClient.getDatabase("sample_mflix");
             MongoCollection<Document> collection = database.getCollection("movies");
 
-            ChangeStreamIterable<Document> changeStream = database.watch()
+            List<Bson> pipeline = Arrays.asList(
+                Aggregates.match(
+                        Filters.in("operationType",
+                                Arrays.asList("insert", "update", "replace"))));
+            ChangeStreamIterable<Document> changeStream = database.watch(pipeline)
                 .fullDocument(FullDocument.UPDATE_LOOKUP);
-            changeStream.forEach(event -> System.out.println("received a change to the collection: " + event));
+            changeStream.forEach(event -> System.out.println("Received a change to the collection: " + event));
         }
     }
 }
--- a/source/includes/usage-examples/code-snippets/Watch.java
+++ b/source/includes/usage-examples/code-snippets/Watch.java
@@ -25,7 +25,7 @@ public class Watch {
             List<Bson> pipeline = Arrays.asList(
                 Aggregates.match(
                         Filters.in("operationType",
-                                Arrays.asList("insert", "update", "replace"))));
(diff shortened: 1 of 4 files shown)


source/usage-examples/watch.txt
```diff
@@ -0,0 +1,171 @@
...
+   List<Bson> pipeline = Arrays.asList(
+                           Aggregates.match(
+                              Filters.lt("fullDocument.runtime", 15)));
+   ChangeStreamIterable<Document> changeStream = database.watch(pipeline);
+
+The ``watch()`` method returns an instance of ``ChangeStreamIterable``, a class
+that offers several methods to access, organize, and traverse the results.
+``ChangeStreamIterable`` also inherits methods from its parent class,
+``MongoIterable`` which implements the core Java interface ``Iterable``.
+You can call ``forEach()`` on the ``ChangeStreamIterable`` to handle
```
ccho-mongodb, line 33: This seems like a good place for a new paragraph since the prior sentences describe the ``ChangeStreamIterable``and here you're describing the ``forEach()`` method.

source/usage-examples/watch.txt
```diff
@@ -0,0 +1,171 @@
...
+events as they occur, or you can use the ``iterator()`` method which
+returns a ``MongoCursor`` instance that you can use to traverse the results.
+You can call methods on the ``MongoCursor`` such as ``hasNext()`` to check
```
ccho-mongodb, line 36: This also seems like a good place for a new paragraph since you're describing methods you can call on ``MongoCursor`` for the next few sentences.

source/usage-examples/watch.txt
```diff
@@ -0,0 +1,171 @@
...
+whether additional results exist, or ``next()`` to return the next document
+in the collection. Unlike the ``MongoCursor`` returned by other queries,
+a ``MongoCursor`` associated with a change stream will wait until a
+change event arrives before returning a result from ``next()`` and
+``hasNext()`` instead of throwing a ``java.util.NoSuchElementException``.
```
ccho-mongodb, line 41: Could we either separate the description of the exception in a new sentence (hasNext() will never throw that exception)? And if there are more events pending that you have not consumed with `next()` from the cursor, it will not throw it either. 
I think if we're going to describe the exception case, it should be separated and very clear in what circumstance it may happen.

source/usage-examples/watch.txt
```diff
@@ -0,0 +1,171 @@
...
+
+To configure options for processing the change stream, use member methods of the
```
ccho-mongodb, line 43: Nitpick: how about "processing the documents returned from the change stream" as they don't actually modify the behavior of the change stream itself. I think the only way you can modify a change stream itself is with the pipeline or resumeAfter option.

source/includes/usage-examples/code-snippets/Watch.java
```diff
@@ -0,0 +1,34 @@
...
+
+        try (MongoClient mongoClient = MongoClients.create(uri)) {
+
+            MongoDatabase database = mongoClient.getDatabase("sample_mflix");
+            MongoCollection<Document> collection = database.getCollection("movies");
+
+            List<Bson> pipeline = Arrays.asList(
+                Aggregates.match(
+                        Filters.in("operationType",
+                                Arrays.asList("insert", "update", "replace"))));
```
ccho-mongodb, line 28: Nitpick: since the example only performs insert/update/delete, I wonder if we should leave out "replace" to reduce the chance someone might get confused by including it in this situation.

source/usage-examples/watch.txt
rozza, line 41: There is a [`tryNext()`](http://mongodb.github.io/mongo-java-driver/4.1/apidocs/mongodb-driver-sync/com/mongodb/client/MongoCursor.html#tryNext()) method, which doesn't block.

```
@Nullable TResult tryNext()
A special next() case that returns the next element in the iteration if available or null.
Tailable cursors are an example where this is useful. A call to tryNext() may return null, but in the future calling tryNext() would return a new element if a document had been added to the capped collection.

Returns:  the next element in the iteration if available or null.
```

nathan-contino-mongo: Sounds good, I've added mention of this.

source/includes/usage-examples/code-snippets/Watch.java
```diff
@@ -0,0 +1,34 @@
...
+            MongoDatabase database = mongoClient.getDatabase("sample_mflix");
+            MongoCollection<Document> collection = database.getCollection("movies");
+
+            List<Bson> pipeline = Arrays.asList(
+                Aggregates.match(
+                        Filters.in("operationType",
+                                Arrays.asList("insert", "update"))));
+            ChangeStreamIterable<Document> changeStream = database.watch(pipeline)
+                .fullDocument(FullDocument.UPDATE_LOOKUP);
+            changeStream.forEach(event -> System.out.println("Received a change to the collection: " + event));
```
rozza, line 31: The only downside to this approach is there's no clear way to stop / exit the process.  Which will lead to questions similar to: https://developer.mongodb.com/community/forums/t/how-to-unwatch-or-stop-a-running-change-stream-with-mongodb-java-driver/6760

I don't have a clear answer other than also suggesting documenting something like:

```
try (MongoChangeStreamCursor<Document> cursor = collection.find().cursor()) {
     while (cursor.hasNext()) {
         System.out.println("Received a change to the collection: " + cursor.next());

         if (shutdown()) {
            break;
         }
     }
 }
```

nathan-contino-mongo: Unfortunately I feel like `forEach()` is probably the most useful function to demonstrate for our target audience of inexperienced developers -- next/hasNext ends up being a lot of verbosity to accomplish the same thing. What if I added some details about the cursor remaining open instead:

```

    … the cursor remains open until one of the following occurs:

        The cursor is explicitly closed.
        An invalidate event occurs.
        If the deployment is a sharded cluster, a shard removal may cause an open change stream cursor to close, and the closed change stream cursor may not be fully resumable.
```
... so even if it's not clear how to close the stream from code, the circumstances under which it will close are made clear?

I could also add an explicit cursor close in the example if you think that would be helpful. But I wonder if we're making it more complicated than it needs to be.

nathan-contino-mongo: @ccho-mongodb what do you think? While I do want to be comprehensive and cover closing a stream, I worry that it might just overcomplicate the examples which are supposed to be straightforward.

ccho-mongodb: Some background info that I think we've already discussed, but just want to preface my recommendation:
In the [Node.js version of this usage example](https://docs-mongodbcom-staging.corp.mongodb.com/da1c19e/node/docsworker-xlarge/preview-071020/usage-examples/changeStream), we had the same app setup the callback, have it exit after it receives the expected event, and then used timers to ensure it exited regardless of success. In hindsight, I wonder if this example is trying to strictly adhere to the documentation template structure we set for Usage Examples, and perhaps we should reduce this example to setup and iteration, and link to a more complete example to Fundamentals that walks them through events that trigger the callbacks to keep the code clean and more reflective of how you might use them.

So my order of preference without any user study input:
1) Only show code to watch a pipeline and iterate through ChangeStream event and link to Fundamentals where we walk through sending events as well.
2) Add code to exit after N events received, explain that they need to kill the app if it does not get the expected number
3) Use a timer to exit automatically

As far as `hasNext` or `forEach`, I'd prefer the latter *only if we can accomplish what we need using it*. I don't think the inexperience/verbosity is an obstacle, but we're using Java 8 try-with-resources, so we should be able to assume they are also familiar with forEach.

nathan-contino-mongo: I think demonstrating callbacks actually being called/sending events is pretty crucial to understanding how change streams work, otherwise you miss out on the entire point -- changes! I think code to exit after N events is an OK solution to make sure the example exits, but I prefer a timer. I'll try both and see if I can come up with something that doesn't feel too clunky.

ccho-mongodb: Ok, I know that you already know this, but keep in mind that content in the Usage Examples is for showing working code examples (generally based on search results) and providing a quick reference to them. The content in Fundamentals are for understanding how to use different methods to accomplish a specific task. There's been some overlap, but it would be great to keep this separation so there's an expected pattern of visiting Fundamentals for further learning if necessary, and that we're not repeating ourselves in the content too much.

nathan-contino-mongo: The current form is a working example that provides a quick reference. Does that mean we should leave this as is?

PR #24, opened 2020-08-04
(DOCSP-9551): Connection Guide

nathan-contino-mongo:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-9551

[Connection overview](https://docs-mongodbcom-staging.corp.mongodb.com/bb38ab1/java/docsworker-xlarge/DOCSP-9551/fundamentals/connection)
[JNDI](https://docs-mongodbcom-staging.corp.mongodb.com/bb38ab1/java/docsworker-xlarge/DOCSP-9551/fundamentals/connection/jndi)
[Enable TLS/SSL](https://docs-mongodbcom-staging.corp.mongodb.com/bb38ab1/java/docsworker-xlarge/DOCSP-9551/fundamentals/connection/tls)


source/fundamentals/connection.txt
```diff
@@ -0,0 +1,636 @@
...
+.. _connection-options:
+
+Connection Options
+------------------
+
+This section explains MongoDB connection and authentication options
+supported by the Java driver. You can pass the connection options as
+parameters of the connection URI to specify the behavior of the client.
+
+.. list-table::
```
ccho-mongodb, line 424: Nitpick: You can probably leave out mention of "Java" when describing the driver as that is implied on this page and explicitly stated in this section.

source/fundamentals/connection.txt
```diff
@@ -0,0 +1,636 @@
...
+1. In a `Wildfly <http://wildfly.org/>`__ installation, create a new module
+   for MongoDB at `modules/system/layers/base/org/mongodb/main`. Copy the
+   ``mongo-java-driver.jar`` jar file into the module. Add the following
+   ``module.xml`` file into the module:
+
+   .. code-block:: xml
+
+      <module xmlns="urn:jboss:module:1.3" name="org.mongodb">
+         <resources>
+            <resource-root path="mongodb-driver-sync-4.1.0-rc0.jar"/>
```
ccho-mongodb, line 97: I'd like to propose moving code snippets that reference a specific version to an include with filename postfix "-with-version" so that we know to update it whenever a new version of the driver is released (or write a script to do that). How does that sound?

source/fundamentals/connection.txt
```diff
@@ -0,0 +1,636 @@
...
+   * - **journal**
+     - boolean
+     - Specifies that the driver must wait for the connected MongoDB
+       instance to group commit to the journal file on disk for all writes.
+
+   * - **w**
+     - string or integer
+     - Specifies the write concern. For more information on values, see
+       the server documentation for the :manual:`w option
+       </reference/write-concern/#w-option>`. Implies **safe=true**.
```
ccho-mongodb, line 531: Question: what does "Implies safe=true" mean?

source/fundamentals/connection.txt
```diff
@@ -0,0 +1,636 @@
...
+.. literalinclude:: /includes/fundamentals/code-snippets/srv.java
+   :language: java
+
+.. _jndi:
+
+Java Naming and Directory Interface (JNDI)
+------------------------------------------
+
+MongoClientFactory includes a `JNDI <http://docs.oracle.com/javase/8/docs/technotes/guides/jndi/index.html>`__
+ObjectFactory implementation that returns ``MongoClient`` instances
```
ccho-mongodb, line 80: Nitpick: Use monospace for class names like ObjectFactory.

source/fundamentals/connection.txt
```diff
@@ -0,0 +1,636 @@
...
+     - integer
+     - Specifies the maximum amount of time, in milliseconds, the Java
+       driver will continue to use a pooled connection before closing the
+       connection.
+
+   * - **safe**
+     - boolean
+     - Specifies that the driver ensures that all writes are acknowledged
+       by the connected MongoDB instance. Throws an exception on any write
+       that does not receive acknowledgement from the connected MongoDB
```
ccho-mongodb, line 519: acknowledgment

source/fundamentals/connection.txt
```diff
@@ -0,0 +1,636 @@
...
+       the server documentation for the :manual:`readPreferenceTags option
+       </reference/connection-string/#urioption.readPreferenceTags>`.
+
+   * - **maxStalenessSeconds**
+     - integer
+     - Specifies, in seconds, how stale a secondardy can be before the
+       Java driver stops communicating with that secondary. Not providing
+       a parameter or explicitly setting it to ``-1`` indicates that there
+       should be no staleness check for secondaries. The minimum value is
+       either 90 seconds or the hearbeat frequency plus 10 seconds, whichever
```
ccho-mongodb, line 559: heartbeat

source/fundamentals/connection.txt
```diff
@@ -0,0 +1,636 @@
...
+.. _mongoclient:
+
+MongoClient
+-----------
+
+Use the ``MongoClients.create()`` method to instantiate an instance of a
+``MongoClient``. As each ``MongoClient`` represents a threadsafe *pool* of
+connections to the database, most applications only require a single
+instance of a ``MongoClient``, even across multiple threads. All resource
+usage limits apply to individual ``MongoClient`` instances.
```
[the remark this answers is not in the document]
nathan-contino-mongo: Added an example here!

source/fundamentals/connection.txt
```diff
@@ -0,0 +1,636 @@
...
+          Container
+        </res-auth>
+      </resource-ref>
+
+This will make a ``MongoClient`` instance accessible via the JNDI name
+``mongodb/MyMongoClient`` in the ``java:comp/env`` context.
+
+.. _tls:
+
+TLS/SSL
```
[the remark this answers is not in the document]
nathan-contino-mongo: Done.

source/fundamentals/connection.txt
```diff
@@ -0,0 +1,636 @@
...
+
+To enable TLS/SSL with :java-sync-api:`MongoClientSettings
+<com/mongodb/MongoClientSettings.html>`, set the ``enabled``
+property of the builder to ``true`` using the
+:java-sync-api:`applyToSslSettings()
+<com/mongodb/MongoClientSettings.Builder.html#applyToSslSettings(com.mongodb.Block)>`
+builder block:
+
+.. code-block:: java
+   :emphasize-lines: 2-3
```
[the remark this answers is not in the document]
nathan-contino-mongo: Ugh. Maybe someday we'll have sphinx parity.

source/fundamentals/connection.txt
```diff
@@ -0,0 +1,636 @@
...
+   * - **replicaSet**
+     - string
+     - Specifies that the :ref:`connection string <connection-uri>`
+       provided includes multiple hosts. When specified, the Java driver
+       attempts to find all members of that set.
+
+   * - **ssl**
+     - boolean
+     - Specifies that all communication with MongoDB instances should
+       use TLS.
```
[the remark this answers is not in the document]
nathan-contino-mongo: I've only included it since it's mentioned by name in the **tls** option, but I've added an extra note that this is superseded by **tls** to encourage users to use tls instead.

source/fundamentals/connection.txt
```diff
@@ -0,0 +1,636 @@
...
+<http://docs.oracle.com/javase/8/docs/technotes/guides/security/jsse/JSSERefGuide.html>`__.
+
+.. _tls-custom-sslContext:
+
+Customize TLS/SSL Configuration with an SSLContext
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+If your TLS/SSL configuration requires additional customization, you can
+set the ``sslContext`` property of your ``MongoClient`` by
+passing an ``SSLContext`` object to the builder in the
```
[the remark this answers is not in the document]
nathan-contino-mongo: I would probably classify this in the "if you need to use this, you ought to already know why" category (at the very least, I've never encountered a use case for this)... but this is a valid point, so I'll add a link to javax.net.ssl.SSLContext.

source/fundamentals/connection.txt
```diff
@@ -0,0 +1,636 @@
...
+
+   MongoClientSettings settings = MongoClientSettings.builder()
+        .applyToSslSettings(builder -> 
+            builder.enabled(true))
+        .build();
+   MongoClient client = MongoClients.create(settings);
+
+.. _tls-disable-hostname-verification:
+
+Disable Hostname Verification
```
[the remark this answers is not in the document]
nathan-contino-mongo: Moving it afterward and adding a warning.

source/fundamentals/connection.txt
```diff
@@ -0,0 +1,636 @@
...
+        .build();
+
+.. _tls-configure-jvm-truststore:
+
+Configure the JVM TrustStore
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Most applications need to configure two JVM system properties to ensure
+that the client is able to validate the TLS/SSL certificate presented
+by a connected MongoDB instance.
```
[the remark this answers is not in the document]
nathan-contino-mongo: Agreed, added one for the key store as well (and split up the two sections, now that TLS/SSL is its own page).

source/fundamentals/connection.txt
```diff
@@ -0,0 +1,636 @@
...
+                    builder.invalidHostNameAllowed(true);
+                })
+        .build();
+
+.. _tls-configure-jvm-truststore:
+
+Configure the JVM TrustStore
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Most applications need to configure two JVM system properties to ensure
```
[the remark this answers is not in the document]
nathan-contino-mongo: With the restructured two-section layout we have now, I'm going to leave the numbers there -- store/StorePassword is a pattern that's remained unchanged for a long time in the JVM, and I doubt it'll change any time soon. In this case I prefer to be straightforward with numbers since omitting them always makes me (as a reader) wonder how much more I'll have to do when I scroll down further!

source/fundamentals/connection.txt
```diff
@@ -0,0 +1,636 @@
...
+   the TLS 1.2 protocol, you may need to upgrade to a later release to
+   connect using TLS 1.2.
+
+.. _compression:
+
+Compression
+-----------
+
+You can compress messages passing between your MongoDB instance and the
+Java driver. MongoDB drivers support up to three different algorithms
```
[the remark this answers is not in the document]
nathan-contino-mongo: I haven't found any version compatibility information for the driver side of things, so I assume that the driver compression algorithm compatibility probably matches the MongoDB version compatibility that supports that version. Seems like a good question to ask during tech review.

source/fundamentals/connection.txt
```diff
@@ -0,0 +1,636 @@
...
+.. _connect-replica-set:
+
+Connect to a Replica Set
+------------------------
+
+A MongoDB replica set deployment is a group of connected instances that
+store the same set of data. This configuration of instances provides data
+redundancy and high data availability.
+
+To connect to a replica set deployment, specify the hostname and port numbers
```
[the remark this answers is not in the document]
nathan-contino-mongo: Adding "(or IP address)" in parentheticals since English doesn't have and/or precedence that's as clear as Java.

source/fundamentals/connection.txt
```diff
@@ -0,0 +1,636 @@
...
+
+   * - **readPreferenceTags**
+     - string
+     - Specifies the read preference tags. For more information on values, see
+       the server documentation for the :manual:`readPreferenceTags option
+       </reference/connection-string/#urioption.readPreferenceTags>`.
+
+   * - **maxStalenessSeconds**
+     - integer
+     - Specifies, in seconds, how stale a secondardy can be before the
```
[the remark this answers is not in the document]
nathan-contino-mongo: Ah, you can see where I started to run out of steam when reviewing the options...

source/fundamentals/connection.txt
nathan-contino-mongo: Seems like a good idea to me.

source/fundamentals/connection.txt
nathan-contino-mongo: Reworded to make this clearer; let me know if you have any further questions.

source/fundamentals/connection.txt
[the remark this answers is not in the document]
ccho-mongodb: The issue is that you mention "Most applications need to configure two properties" and later "the typical application also needs to set two properties", not that I think that keystore and truststore are going to change in the JVM. I know you're addressing"typical application", doesn't "most" imply some overlap with "typical" and therefore four are usually required?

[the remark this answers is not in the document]
nathan-contino-mongo: I've pushed my changes, and I feel that separating the key/trust store sections (under one larger Configure Certificates section) has made things a little more goal-oriented and hopefully addressed the issue. If you feel strongly about it I can remove the numbers entirely, though.

source/fundamentals/connection.txt
```diff
@@ -0,0 +1,510 @@
...
+deployment using the driver. For information about authenticating
+with a MongoDB instance, see :ref:`Authentication <authentication>`.
+
+.. _mongoclient:
+
+MongoClient
+-----------
+
+Use the ``MongoClients.create()`` method to construct a
+``MongoClient``. As each ``MongoClient`` represents a threadsafe *pool* of
```
ccho-mongodb, line 25: Nitpick: if you provide emphasis on a term, I would expect an explanation of the term to follow. I would either:
1) omit the emphasis or 
2) explain the term

source/fundamentals/connection.txt
```diff
@@ -0,0 +1,510 @@
...
+connections to the database, most applications only require a single
+instance of a ``MongoClient``, even across multiple threads. All resource
+usage limits, such as max connections, apply to individual
```
ccho-mongodb, line 28: Good example! 👍

source/fundamentals/connection.txt
```diff
@@ -0,0 +1,510 @@
...
+``MongoClient`` instances.
+
+.. note::
+
+   Always call ``MongoClient.close()`` to clean up resources when an
+   instance is no longer needed.
```
ccho-mongodb, line 34: Good to mention. Since our examples use try with resources, I wonder if that might be appropriate to mention as well as an alternative to close(), and link to: https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html

source/fundamentals/connection.txt
```diff
@@ -0,0 +1,510 @@
...
+``retryWrites=true`` and ``w=majority``. For more information on connection
+options, skip to the :ref:`connection options <connection-options>` section.
+
+The code below shows how you can use the sample connection URI in a client to
+connect to MongoDB.
+
+.. literalinclude:: /includes/fundamentals/code-snippets/srv.java
+   :language: java
+
+.. _jndi:
```
ccho-mongodb, line 84: I think moving JNDI to a separate section as the current Java reference docs are structured could be helpful in separating concerns. I also think connecting to a replica set should be directly after Connection URI since it consists of passing different values to the URI that was just introduced. What do you think?

source/fundamentals/connection.txt
```diff
@@ -0,0 +1,510 @@
...
+                  <property name="connectionString" value="mongodb+srv://<user>:<password>@<cluster-url>?retryWrites=true&w=majority"/>
+               </environment>
+            </object-factory>
+         </bindings>
+         <remote-naming/>
+      </subsystem>
+
+   .. note::
+
+      Replace the placeholder connection value URI here with a value that
```
ccho-mongodb, line 127: Nitpick: rather than "here", it could be useful to specify "in the property tag" since we don't have emphasize-lines to give them a direct cue.

source/fundamentals/connection.txt
```diff
@@ -0,0 +1,510 @@
...
+         auth="Container"
+         type="com.mongodb.MongoClient"
+         closeMethod="close"
+         factory="com.mongodb.client.MongoClientFactory"
+         singleton="true"
+         connectionString="mongodb+srv://<user>:<password>@<cluster-url>?retryWrites=true&w=majority"/>
+
+   .. note::
+
+      Replace the placeholder connection value URI here with a value that
```
ccho-mongodb, line 157: Similar comment on specifying where: "in the connectionString attribute"

source/fundamentals/connection.txt
```diff
@@ -0,0 +1,510 @@
...
+redundancy and high data availability.
+
+To connect to a replica set deployment, specify the hostname (or IP address)
+and port numbers of one or more members of the replica set separated by commas.
+Use the :manual:`replica set option
+</reference/connection-string/#urioption.replicaSet>` to specify the name
+of the replica set with which your application is connecting. When the
+replica set option is specified, the driver assumes that your
+connection string uses a seed list of multiple hosts instead of a single
+``host:port`` configuration.
```
ccho-mongodb, line 196: I like to describe examples immediately before presenting them. In this case, I might write:
"The connection string below specifies three hosts in the cluster and the replica set named 'myRs'"

source/fundamentals/connection/tls.txt
```diff
@@ -0,0 +1,184 @@
...
+<com/mongodb/MongoClientSettings.html>` objects.
+
+.. _tls-connectionstring:
+
+Enable TLS/SSL via Connection String
+------------------------------------
+
+To enable TLS/SSL on a connection with a :java-sync-api:`ConnectionString
+<com/mongodb/ConnectionString.html>`, assign the connection string
+parameter ``ssl`` a value of ``true`` in the connection string passed to
```
ccho-mongodb, line 28: Nitpick: the parameter name doesn't match the example

source/fundamentals/connection/tls.txt
```diff
@@ -0,0 +1,184 @@
...
+
+   MongoClientSettings settings = MongoClientSettings.builder()
+        .applyToSslSettings(builder -> 
+            builder.enabled(true))
+        .build();
+   MongoClient client = MongoClients.create(settings);
+
+.. _tls_configure-certificates:
+
+Configure Certificates
```
ccho-mongodb, line 56: Great, I like this grouping.

source/includes/fundamentals/code-snippets/wildfly-module-4.0.0.xml
```diff
@@ -0,0 +1,10 @@
+<module xmlns="urn:jboss:module:1.3" name="org.mongodb">
+   <resources>
+      <resource-root path="mongodb-driver-sync-4.0.0.jar"/>
```
ccho-mongodb, line 3: No need to change at the moment; 4.1.0 is the latest driver.
Also for the filename, I meant the postfix name would be "-with-version" so we could find filenames that included "-with-version". No need to change this or implement any naming scheme in this PR.

source/fundamentals/connection.txt
nathan-contino-mongo: sounds good

source/includes/fundamentals/code-snippets/wildfly-module-4.0.0.xml
nathan-contino-mongo: Gotcha, updated.

nathan-contino-mongo:
Some pending questions for the technical reviewer:
1. Are the compression algorithm dependencies explicitly required to be included for users who want to use those compression algorithms who are not using the uber jar?
2. What driver versions first supported each compression algorithm (zlib/zstd/snappy)?

jyemin:
1. Explicit dependencies must be included for zstd and snappy. zlib support is part of the JVM so we don't need a dependency for it.  http://mongodb.github.io/mongo-java-driver/4.1/driver/tutorials/compression/#dependencies
2. Not sure, but do you need to include that?  My understanding is that we're documenting the 4.1 driver here, and don't need to look backwards to see when support was added.  If you do decide to include this info, it's easy to find.  Just check the fix version on the Jira ticket.  Snappy and Zlib were added in JAVA-2527.  Zstandard in JAVA-3145

nathan-contino-mongo:
@jyemin Thanks for the answers, sounds like my best guess for both of these was right on the money. Please take a look at the rest of the PR when you get the chance to give this a proper review!

jyemin:
@rozza can you take the first pass on this one?

source/fundamentals/connection.txt
```diff
@@ -0,0 +1,434 @@
...
+``MongoClient``. As each ``MongoClient`` represents a threadsafe pool of
+connections to the database, most applications only require a single
+instance of a ``MongoClient``, even across multiple threads. All resource
+usage limits, such as max connections, apply to individual
+``MongoClient`` instances.
+
+.. note::
+
+   Always call ``MongoClient.close()`` to clean up resources when an
+   instance is no longer needed. Alternatively, in Java SE 7 and later,
```
rozza, line 41: 4.0.0 requires Java 8 or higher. So perhaps just mention its best practice to use try with resources.

source/fundamentals/connection.txt
```diff
@@ -0,0 +1,434 @@
...
+   you can use a `try-with-resources statement
+   <https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html>`__
+   to automatically close the ``MongoClient`` when done.
+
+.. _connection-uri:
+
+Connection URI
```
rozza, line 48: Connection URI or Connection String? Is there an editorial policy?

source/fundamentals/connection.txt
```diff
@@ -0,0 +1,434 @@
...
+To connect to a replica set deployment, specify the hostname (or IP address)
+and port numbers of one or more members of the replica set separated by commas.
+Use the :manual:`replica set option
+</reference/connection-string/#urioption.replicaSet>` to specify the name
+of the replica set with which your application is connecting. When the
+replica set option is specified, the driver assumes that your
+connection string uses a seed list of multiple hosts instead of a single
+``host:port`` configuration.
+
+The connection string below specifies three hosts in the cluster and
```
rozza, line 112: Do we need the replicaSet name? The driver doesn't it will work it out from the hosts.  The replicaSet parameter seems like its an old option that serves no real purpose.

source/fundamentals/connection.txt
```diff
@@ -0,0 +1,434 @@
...
+.. _compression:
+
+Compression
+-----------
+
+You can compress messages passing between your MongoDB instance and the
+driver. MongoDB drivers support up to three different algorithms
+depending on release version:
+
+1. `Snappy <https://google.github.io/snappy/>`__: available in MongoDB 3.4 and later.
```
rozza, line 134: Perhaps make a note here that snappy & zstandard require an extra dependency?

source/fundamentals/connection.txt
```diff
@@ -0,0 +1,434 @@
...
+       driver will allow a pooled connection to idle before closing the
+       connection.
+
+   * - **maxLifeTimeMS**
+     - integer
+     - Specifies the maximum amount of time, in milliseconds, the Java
+       driver will continue to use a pooled connection before closing the
+       connection.
+
+   * - **safe**
```
rozza, line 311: I'm not sure we need to document this,  writeconcern is a better parameter for safety.

ccho-mongodb:
Just noticed this:

Screen Shot 2020-08-05 at 11 37 05 AM

source/fundamentals/connection/jndi.txt
```diff
@@ -0,0 +1,100 @@
...
+              <res-type>
+                com.mongodb.MongoClient
+              </res-type>
+              <res-auth>
+                Container
+              </res-auth>
+            </resource-ref>
+
+      This will make a ``MongoClient`` instance accessible via the JNDI name
+      ``mongodb/MyMongoClient`` in the ``java:comp/env`` context.
```
rozza, line 100: @jyemin do you think its worth adding a note that MongoClientFactory's are simple wrappers and for any non support JNDI implementations users can easily create their own factory?

source/fundamentals/connection/tls.txt
```diff
@@ -0,0 +1,191 @@
...
+<http://docs.oracle.com/javase/8/docs/technotes/tools/unix/keytool.html>`__
+or `openssl <https://www.openssl.org/docs/apps/openssl.html>`__ command
+line tools.
+
+For more information on configuring a Java application to use TLS/SSL,
+please refer to the `JSSE Reference Guide
+<http://docs.oracle.com/javase/8/docs/technotes/guides/security/jsse/JSSERefGuide.html>`__.
+
+.. _tls-disable-hostname-verification:
+
```
rozza, line 129: Can we add a debugging tls/ssl section or note block with a link to : https://docs.oracle.com/javase/8/docs/technotes/guides/security/jsse/ReadDebug.html to help users debugging any issues.  Tls on java is a real pain so mentioning setting the system property: `-Djavax.net.debug=all` to generate debugging output will really help users out.

source/fundamentals/connection.txt
nathan-contino-mongo: Changing this to Connection String as that seems to be the more popular term. Added a parenthetical to clarify that connection uri == connection string.

source/fundamentals/connection.txt
nathan-contino-mongo: There is a subsection that explains this below -- do you think we need more than that?

source/fundamentals/connection.txt
nathan-contino-mongo: Agreed. I'd like to remove this, but it is mentioned in quite a few other option descriptions. I worry that those descriptions would feel incomplete without that mention (or without a definition for the **safe** option). The same thing happened with the **ssl** option -- do you think we should hide both of these entirely, or do you think it's OK to keep them around since they have historical value?

source/fundamentals/connection/tls.txt
nathan-contino-mongo: Great suggestion, added a note at the top of the TLS page!

source/fundamentals/connection.txt
nathan-contino-mongo: I'll leave the option/description in the options section (with a note about the fact that it's redundant these days) but I'll remove the bits from "connect to a replica set".

nathan-contino-mongo:
@rozza Some fixes and some questions for you above. Please take a look when you get a chance!

source/fundamentals/connection/jndi.txt
nathan-contino-mongo: Open to this, if @jyemin thinks it's a good idea.

jyemin: Yes, it's a good idea.  Though I haven't heard anyone ask about JNDI in ages.

source/fundamentals/connection.txt
rozza: Regarding ssl - I think that might be useful as not all users know the difference between ssl and tls.

I'm not sure where else safe is used, so happy to leave the decision to you. To note: there is a link at the end to all options via the java doc.

source/fundamentals/connection.txt
rozza: I'm probably being over zealous that someone will miss that.

nathan-contino-mongo:
@rozza thanks for taking another look! Added a note to the JNDI doc -- please take a look and let me know if it looks good to you!

source/fundamentals/connection.txt
```diff
@@ -0,0 +1,435 @@
...
+``MongoClient``. As each ``MongoClient`` represents a threadsafe pool of
+connections to the database, most applications only require a single
+instance of a ``MongoClient``, even across multiple threads. All resource
+usage limits, such as max connections, apply to individual
+``MongoClient`` instances.
+
+.. note::
+
+   Always call ``MongoClient.close()`` to clean up resources when an
+   instance is no longer needed. The `try-with-resources statement
```
jyemin, line 41: `try-with-resources` is not that useful in practice, as it requires that the lifetime of the MongoClient is scoped to a single method.  In a typical web application, where you want to cache the MongoClient for the life of that application, this is not generally the case.  In fact, if you're doing that, you're usually doing something wrong, because you are creating a MongoClient per web request!

jyemin: I suggest removing this.

source/fundamentals/connection.txt
```diff
@@ -0,0 +1,435 @@
...
+Connect to a Replica Set
+------------------------
+
+A MongoDB replica set deployment is a group of connected instances that
+store the same set of data. This configuration of instances provides data
+redundancy and high data availability.
+
+To connect to a replica set deployment, specify the hostname (or IP address)
+and port numbers of one or more members of the replica set separated by commas.
+Use the  to specify the name
```
jyemin, line 106: `Use the  to`?

source/fundamentals/connection.txt
```diff
@@ -0,0 +1,435 @@
...
+of the replica set with which your application is connecting. When the
+replica set option is specified, the driver assumes that your
```
jyemin, line 108: Not sure what you mean here.  What is the assumption?  

I think what you're referring to is that:

`mongodb://<hostname:port>` will create a direct connection to a single server, even if that server is a member of a replica set.  whereas `mongodb://<hostname:port>/?replicaSet=rs0` will create a replica set connection, and discover all the members.  

If so, you should clarify

source/fundamentals/connection.txt
jyemin: I think you're missing the `replicaSet` query parameter name here.

source/fundamentals/connection.txt
```diff
@@ -0,0 +1,435 @@
...
+.. note::
+
+   Although it is possible to connect to a replica set deployment by
+   providing only a single host, you should provide the driver with the
+   full list to ensure that it is able to connect even if one host fails.
+
+.. note::
+
+   The :manual:`replicaSet option
+   </reference/connection-string/#urioption.replicaSet>` is no longer
```
jyemin, line 128: `no longer`?  It was never required in the Java driver

source/fundamentals/connection.txt
```diff
@@ -0,0 +1,435 @@
...
+                                            MongoCompressor.createZstdCompressor()))
+              .build();
+         MongoClient client = MongoClients.create(settings);
+
+.. _compression-dependencies:
+
+Compression Algorithm Dependencies
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The JDK supports `Zlib <https://zlib.net/>`__ compression natively, but
```
jyemin, line 210: You should mention that applications that require snappy or zstd need to add their own explicit dependency on those libraries.

source/fundamentals/connection.txt
jyemin: I also think that `safe` should be removed.  Especially as it just means `w=1`, which is the default anyway.

jyemin: It also implies something that is not true: w: 1 writes can be lost on a rollback

source/fundamentals/connection.txt
```diff
@@ -0,0 +1,435 @@
...
+     - Specifies, in seconds, how stale a secondary can be before the
+       driver stops communicating with that secondary. Not providing
+       a parameter or explicitly setting it to ``-1`` indicates that there
+       should be no staleness check for secondaries. The minimum value is
+       either 90 seconds or the heartbeat frequency plus 10 seconds, whichever
+       is greater. For more information, see the server documentation for the
+       :manual:`maxStalenessSeconds option
+       </reference/connection-string/#urioption.maxStalenessSeconds>`.
+
+   * - **authMechanism**
```
jyemin, line 369: I don't think we say anything about how to specify username and password, do we?

source/fundamentals/connection/tls.txt
```diff
@@ -0,0 +1,199 @@
...
+Configure the JVM Trust Store
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The JVM trust store saves certificates that securely identify other
+applications with which your Java application interacts. Using these
+certificates, your application can prove that the connection to another
+application is genuine and secure from tampering by third parties.
+
+An application that serves TLS/SSL requests needs to configure two
```
jyemin, line 90: The driver is not serving requests, it's initiating them.

source/fundamentals/connection/tls.txt
```diff
@@ -0,0 +1,199 @@
...
+
+Configure the JVM Key Store
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The JVM key store saves certificates that securely identify your Java
+application to other applications. Using these certificates, other
+applications can prove that the connection to your application is
+genuine and secure from tampering by third parties.
+
+An application that serves TLS/SSL requests needs to set two JVM system
```
jyemin, line 117: Same as above

source/fundamentals/connection/tls.txt
```diff
@@ -0,0 +1,199 @@
...
+
+.. code-block:: console
+
+   keytool -importcert -trustcacerts -file <path to certificate authority file>
+            -keystore <path to trust store> -storepass <password>
+
+Configure the JVM Key Store
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The JVM key store saves certificates that securely identify your Java
```
jyemin, line 112: We should note that by default the server does not require the client to prove it's identity.  For example, to connect to Atlas this is not necessary.

source/fundamentals/connection/tls.txt
```diff
@@ -0,0 +1,199 @@
...
+
+Java applications that serve TLS/SSL requests require access to
+cryptographic certificates that prove identity for the application
+itself as well as other applications with which the application
+interacts. To configure access to these certificates in your application,
+you should use the JVM Trust Store and the JVM Key Store.
+
+.. _tls-configure-jvm-truststore:
+
+Configure the JVM Trust Store
```
jyemin, line 82: We should note that by default the JRE includes most commonly used public certificates from signing authorities, e.g. Let's Encrypt.  So by default you don't have to do anything here.  e.g. to connect to an Atlas server it's enough to just enable TLS and it will just work.

source/fundamentals/connection.txt
nathan-contino-mongo: Seems fair -- try-with-resources makes sense for usage examples, doesn't make sense for enterprise applications.

source/fundamentals/connection.txt
nathan-contino-mongo: We don't, but I've added a link to the auth section so we don't have to repeat this.

PR #25, opened 2020-08-04
DOCSP-9552 auth mechanisms

ccho-mongodb:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-9552

https://docs-mongodborg-staging.corp.mongodb.com/8e2a078/java/docsworker-xlarge/DOCSP-9552-auth-mechanisms/fundamentals/auth


source/fundamentals/auth.txt
```diff
@@ -0,0 +1,384 @@
...
+Authentication Mechanisms
+=========================
+
+.. default-domain:: mongodb
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
```
nathan-contino-mongo, line 11: Given the fact that you list each of the auth mechanisms below, I'm not sure if we need the "on this page" widget here. It feels repetitive to me.

source/fundamentals/auth.txt
```diff
@@ -0,0 +1,384 @@
...
+mechanism:
+
+.. tabs::
+
+   .. tab::
+      :tabid: Connection String
+
+      To specify the ``SCRAM-SHA-256`` authentication mechanism using a
+      connection string, pass the ``SCRAM-SHA-256`` value as a parameter
+      of ``authMechanism`` in your connection string. Your code to instantiate
```
nathan-contino-mongo, line 167: "pass the ``SCRAM-SHA-256`` value as a parameter of ``authMechanism`` in your connection string" sounds strange to me, though I can't quite say why. I would probably word this more like: "assign the ``authMechanism`` parameter a value of ``SCRAM-SHA-256`` in your connection string. What do you think? This applies to all of the connection string tabs across the page.

source/fundamentals/auth.txt
```diff
@@ -0,0 +1,384 @@
...
+using ``SCRAM-SHA-1``. If the instance does not support that mechanism either,
+the driver attempts to authenticate using ``MONGODB-CR``.
+
+The following code snippets show how to specify the connection mechanism,
+using the following placeholders:
+
+* ``username`` - your MongoDB username
+* ``password`` - your MongoDB user's password
+* ``hostname`` - network address of your MongoDB server, accessible by your client
+* ``port`` - port number of your MongoDB server
```
nathan-contino-mongo, line 96: You specify the default value of "admin" for ``authenticationDb`` -- should we also mention the default value of 27017 for the Mongo access port?

source/fundamentals/auth.txt
```diff
@@ -0,0 +1,384 @@
...
+using the following placeholders:
+
+* ``username`` - value of your ``AWS_ACCESS_KEY_ID``
+* ``password`` - value your ``AWS_SECRET_ACCESS_KEY``
+* ``hostname`` - network address of your MongoDB server, accessible by your client
+* ``port`` - port number of your MongoDB server
+* ``authenticationDb`` - MongoDB database that contains your user's
+  authentication data. If you omit this parameter, the driver uses the
+  default value ``admin``.
+* ``awsSessionToken`` - value your ``AWS_SESSION_TOKEN`` *(optional)*
```
nathan-contino-mongo, line 276:
```suggestion
* ``awsSessionToken`` - value of your ``AWS_SESSION_TOKEN`` *(optional)*
```

source/fundamentals/auth.txt
```diff
@@ -0,0 +1,384 @@
...
+      .. code-block:: java
+
+         MongoClient mongoClient = MongoClients.create("mongodb://<username>:<password>@<hostname>:<port>/?authSource=<authenticationDb>&authMechanism=MONGODB-AWS");
+
+      If you need to specify an AWS session token, include it in the
+      ``authMechanismProperties`` parameter as follows:
+
+      .. code-block:: java
+
+         MongoClient mongoClient = MongoClients.create("mongodb://<username>:<password>@<hostname>:<port>/?authSource=<authenticationDb>&authMechanism=MONGODB-AWS&authMechanismProperties=AWS_SESSION_TOKEN:<awsSessionToken>");
```
nathan-contino-mongo, line 301: previous snippets have trained me to look for a key-value pairing (i would expect `AWS_SESSION_TOKEN=<token>`), do you think you can add an explicit mention that the ``authMechanismProperties``query parameter is used to declare and define the AWS_SESSION_TOKEN?

source/fundamentals/auth.txt
```diff
@@ -0,0 +1,384 @@
...
+:wikipedia:`TLS <Transport_Layer_Security>` with X.509 certificates to
+authenticate your user, identified by the distinguished name (DN) of your
+client certificate. For more information on determining the *subject* name from
+the X.509 certificate, see the MongoDB manual
+:manual:`X.509 tutorial </tutorial/configure-x509-client-authentication/#add-x-509-certificate-subject-as-a-user>`.
+
+The following code snippets show how to specify the connection mechanism,
+using the following placeholders:
+
+* ``username`` - your MongoDB username, derived from the certificate subject
```
nathan-contino-mongo, line 341: I think your previous mention of how to define the username:
```
For more information on determining the *subject* name from
the X.509 certificate, see the MongoDB manual
:manual:`X.509 tutorial </tutorial/configure-x509-client-authentication/#add-x-509-certificate-subject-as-a-user>`.
```
is somewhat easy to glance over by mistake. I think the link might be more discoverable if you linked the text "derived from the certificate subject" to that tutorial instead.

source/includes/fundamentals/code-snippets/auth-credentials-aws-session.rst
```diff
@@ -0,0 +1,11 @@
+.. code-block:: java
```
nathan-contino-mongo, line 1: I noticed that all of your code snippets are actually .rst files containing only a code block. Is there a reason you would prefer this to a .java file (even if it isn't a complete program) and a literalinclude? This isn't so much a comment or suggestion as it is just curiosity if I ought to follow this pattern as well for consistency.

source/fundamentals/auth.txt
ccho-mongodb: Thanks for bringing this up. It's a tough call, but I think omitting instruction about the placeholders that are more related to connection than authentication is justified for keeping the description concise. I think this is a good place given there isn't any mention of the authentication db parameter in the Connection Guide (which is appropriate IMHO).

source/fundamentals/auth.txt
ccho-mongodb: I think the description prior to this mentions that they'll need to include the AWS session token in that property and the code example shows how it is formatted using the placeholders. Are you suggesting rewording it? Should this be in an admonition instead?

source/includes/fundamentals/code-snippets/auth-credentials-aws-session.rst
ccho-mongodb: Yes, I'd like to stick to rst for snippets unless it's a full class file (includes imports, class declaration, etc.). If you have an objection, please let me know!

source/fundamentals/auth.txt
ccho-mongodb: I think it's important for a lengthy page like this one to have a ToC at the top since we establish this as an expected pattern in our docs. I made the decision to add links into the bulleted list (not present in the Node driver docs) because the examples are much longer and don't fit in the typical viewport. If you strongly feel this is repetitive, I can remove the links from the bulleted list as it was a late refinement.

source/fundamentals/auth.txt
ccho-mongodb: Great suggestion, will update.

source/fundamentals/auth.txt
ccho-mongodb: Ok, would you mind if I linked it from both? I want to retain mention of what information they provide to use the specific mechanism which I did with SCRAM and AWS.

source/fundamentals/auth.txt
nathan-contino-mongo: Ah sorry, I didn't quite express myself clearly here. I meant to say that maybe we should mention that it's not merely a value for the `authMechanismProperties` parameter, but actually another way of expressing values. I could see a user pasting their `<awsSessionToken>` directly into the value of `authMechanismProperties` otherwise.

source/fundamentals/auth.txt
```diff
@@ -0,0 +1,383 @@
...
+mechanism:
+
+.. tabs::
+
+   .. tab::
+      :tabid: Connection String
+
+      To specify the ``X.509`` authentication mechanism using a connection
+      string, pass the ``MONGODB-X509`` value to the ``authMechanism``
+      parameter and enable TLS by passing a ``true`` value to the ``tls``
```
ccho-mongodb, line 359: Note to self to update.

ccho-mongodb:
> LGTM with some small suggestions/thoughts. Two other thoughts that you may want to consider:
> 
> 1. I know when I first tried to connect with the Java driver using X509 it was a painful and long process because I didn't use exactly the correct console snippet in the maual to extract the subject from the certificate. While the other methods are fairly self-explanatory for user-password, I wonder if we ought to provide a very small example of a certificate (presumably with the body itself truncated) and the resulting username. I know that would have helped me a lot.
> 2. This is probably going above and beyond, but my immediate thought (and I'm sure the immediate thought of many a developer) after reading this was: but how do I get my password into this securely? What's the best practice? An example is probably far too complicated and far too much, but a push in the right direction could be very helpful for a lot of developers here.

Sorry didn't respond to this earlier.
For the first item, do you feel that the link to the manual that shows them how to extract their username from the certificate using openssl isn't sufficient? I think certificates can take different formats, so they should be using the tool to extract it rather than looking at a certificate.

For the second item, I think this is more of a language and environment issue rather than a driver one. The driver doesn't provide any mechanism for this, but maybe there's an article we can point to that explains some options. What if we created an admonition that mentions that we recommend sourcing the value of the password from an external source to avoid exposing it in the source code? The issue I see is there are too many different development environments, levels of security, and options to provide a quality recommendation. It could also be a good FAQ question if this comes up frequently in the forums.

nathan-contino-mongo:
1) I mostly feel like seeing an concrete example makes it a lot easier to understand what you need to use as your subject. The problem I experienced was that depending on what tool you use to extract the subject, you can get slightly different results, and we require an exact string match for the username. The manual link is extremely helpful, but that page in general is really intimidating and long and it's easy to open the tab and forget about it. Considering how prone to issues X509 can be, I think a little nudge in the right direction would go a long way.

2) I agree that we can't give a comprehensive breakdown, but maybe pointing to best practices for a couple of common java frameworks could give another "nudge in the right direction." I actually think this could be a perfect space for a DevHub article that we could link to or, depending on maintainability, pull into our docs as a guide. For now I think an admonition with a vague recommendation and maybe a helpful link or two is our best bet.

source/fundamentals/auth.txt
ccho-mongodb: (Good place for the authenticationDb, not so much for the default port)

source/fundamentals/auth.txt
ccho-mongodb: Ok, will clarify.

ccho-mongodb:
> 1. I mostly feel like seeing an concrete example makes it a lot easier to understand what you need to use as your subject. The problem I experienced was that depending on what tool you use to extract the subject, you can get slightly different results, and we require an exact string match for the username. The manual link is extremely helpful, but that page in general is really intimidating and long and it's easy to open the tab and forget about it. Considering how prone to issues X509 can be, I think a little nudge in the right direction would go a long way.

Ok, I will rephrase the content of the manual link. I think this is probably content that is potentially outside of our target personas as it's generally the DBA's role to create users and credentials, and pass them onto the users.

nathan-contino-mongo:
>  I think this is probably content that is potentially outside of our target personas as it's generally the DBA's role to create users and credentials, and pass them onto the users.

Agreed that creating the user is outside of the scope of what our users will probably be doing (though plenty of fullstack developers these days have to wear the DBA hat, hence lots of DB security issues!). *However*... even if a DBA provides the developer with a certificate, the user will probably still have to extract the username from the certificate, so they'll need to run a snippet of code from that link.

ccho-mongodb:
> > I think this is probably content that is potentially outside of our target personas as it's generally the DBA's role to create users and credentials, and pass them onto the users.
> 
> Agreed that creating the user is outside of the scope of what our users will probably be doing (though plenty of fullstack developers these days have to wear the DBA hat, hence lots of DB security issues!). _However_... even if a DBA provides the developer with a certificate, the user will probably still have to extract the username from the certificate, so they'll need to run a snippet of code from that link.

They can provide the username because their user needs to be created in the authentication database.  See step 2 of the manual link.

source/fundamentals/auth.txt
```diff
@@ -0,0 +1,388 @@
...
+      .. include:: /includes/fundamentals/code-snippets/auth-credentials-aws-session.rst
+
+.. _x509-auth-mechanism:
+
+``X.509``
+~~~~~~~~~
+
+.. note::
+   The X.509 authentication mechanism is only available in MongoDB versions
+   2.6 and later.
```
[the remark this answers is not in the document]
ccho-mongodb: I'll get rid of it, thanks.

source/fundamentals/auth.txt
```diff
@@ -0,0 +1,384 @@
...
+to connect and authenticate to your MongoDB cluster. To specify your
+authentication mechanism using a connection string, include your settings in
+your connection string and pass it to the ``MongoClients.create()`` method
+to instantiate your ``MongoClient``. For descriptions of how to use a
+connection string to specify your authentication method, navigate to the
+relevant mechanism section and select the :guilabel:`Connection String` tab.
+
+The ``MongoCredential`` class contains static factory methods that allow you
+to specify your authentication mechanism and credentials. You can use this
+helper class to keep your code organized. If you choose to use the
```
jyemin, line 52: `to keep your code organized`?  Not sure what you mean. Can you clarify?

source/fundamentals/auth.txt
```diff
@@ -0,0 +1,384 @@
...
+when constructing your ``MongoClient``. For descriptions of how to use the
+``MongoCredential`` class to specify your authentication method, navigate
+to the relevant mechanism section and select the :guilabel:`MongoCredential`
+tab.
+
+For more information on the aforementioned classes and methods, refer to the
+following API documentation:
+
+- :java-sync-api:`MongoClients.create() </com/mongodb/client/MongoClients.html#create(com.mongodb.ConnectionString)>`
+- :java-sync-api:`MongoClient </com/mongodb/client/MongoClient.html>`
```
jyemin, line 64: Is MongoClient relevant here?

source/fundamentals/auth.txt
```diff
@@ -0,0 +1,384 @@
...
+
+The default authentication mechanism is a fallback setting that instructs
+the driver to negotiate the first authentication mechanism supported by the
+server in the following order of preference:
+
+#. ``SCRAM-SHA-256``
+#. ``SCRAM-SHA-1``
+#. ``MONGODB-CR``
+
+When you use the default authentication mechanism, the driver first attempts
```
jyemin, line 84: I don't think this is accurate.  Where did you get this information from?

source/fundamentals/auth.txt
```diff
@@ -0,0 +1,384 @@
...
+tab below for instructions and sample code for specifying this authentication
+mechanism:
+
+.. tabs::
+
+   .. tab::
+      :tabid: Connection String
+
+      To specify the default authentication mechanism using a connection
+      string, explicitly omit the mechanism. Your code to instantiate
```
jyemin, line 111: It's weird to say to explicitly omit something.  Seems like an oxymoron!.  Can we just remove the word `explicitly`?

source/fundamentals/auth.txt
```diff
@@ -0,0 +1,384 @@
...
+
+To retrieve the X.509 certificate *subject* which is used as your
+*username* for authentication, see the MongoDB server :manual:`X.509
+tutorial </tutorial/configure-x509-client-authentication/#add-x-509-certificate-subject-as-a-user>`
+for more information.
+
+The following code snippets show how to specify the connection mechanism,
+using the following placeholders:
+
+* ``username`` - your MongoDB username, :manual:`derived from the certificate subject </tutorial/configure-x509-client-authentication/#add-x-509-certificate-subject-as-a-user>`
```
jyemin, line 342: As of MongoDB 3.4, it's no longer required (or recommended) to explicitly provide the username.  The server will pull it out of the certificate for you.  You can just use `MongoCredential.createMongoX509Credential()` programmatically, and just leave out the username on the connection string.  It's much easier that way!

source/fundamentals/auth.txt
ccho-mongodb: Ah, what's the behavior then? I was working from the spec: https://github.com/mongodb/specifications/blob/master/source/auth/auth.rst#defaults

source/fundamentals/auth.txt
ccho-mongodb: Sure, makes sense!

source/fundamentals/auth.txt
ccho-mongodb: Awesome, I'll update the documentation to describe the easier way.

source/fundamentals/auth.txt
jyemin: Right.  It's hard to understand, and what happens depends on the server version that the driver is connected to.

It would be accurate to say just that the driver will default to a mechanism depending on the server version it's connected to.  If you want to go into more detail, you can say that it will default to MONOGDB-CR for 3.6 and below, and for 4.0 and up it will default based on what mechanisms the server advertises support for.

The only retry of auth has to do with the speculative authentication support added in 4.4, but I don't think it's worth mentioning that, as it's just an optimization to try and save one round trip to the server by combining the isMaster from the handshake with the saslStart for the auth conversation.

source/fundamentals/auth.txt
ccho-mongodb: By "organized", I was referring to keeping specific pieces of logic in separate places that are identified by the code that wraps them. I labeled "organized" for lack of my ability to find a better word. 

I'm probably missing the advantages of using the helper class other than ensuring you're providing parameters that fit the method signature. If there any, can you explain them to me?

source/fundamentals/auth.txt
ccho-mongodb: I agree that it's easer to understand by mentioning only that it defaults to the mechanisms the server supports, will update.

source/fundamentals/auth.txt
ccho-mongodb: While it's not tied to the authentication classes/methods, I thought it belonged here because it is the end product of the process of establishing a connection with authentication. I can remove this and the ``MongoClients.create()`` method if you think it's only relevant to include API links to the authentication-specific methods.

source/fundamentals/auth.txt
jyemin: You have it basically right.  The connection string is a grab bag of properties about everything from servers to credentials to timeouts to write concern.  It may not make sense to externalize all of them together in a single string (especially when the credentials are the one piece that have to remain secure).   As one example, some applications use an external credential provider which will set up temporary one-time credentials to use with MongoDB, that are provided to the application on startup via some sort of REST service.  So in that case you could have a connection string indicating the server names, and combine it with credentials that are retrieved at run time., e.g. 

```
   MongoClientSettings.builder()
       .applyConnectionString(...)
       .credential(<get credential at runtime>)
      .build()
```

source/fundamentals/auth.txt
jyemin: I don't have strong feelings about it.

source/fundamentals/auth.txt
ccho-mongodb: Ok, I'll give summarizing this a try.

ccho-mongodb:
@jyemin Here's the new PR https://github.com/mongodb/docs-java/pull/27
Sorry again for not waiting for your review!

PR #39, opened 2020-11-19
DOCSP-12158: Fundamentals Documents

ccho-mongodb:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-12158

https://docs-mongodbcom-staging.corp.mongodb.com/b067e65/java/docsworker-xlarge/DOCSP-12158-fundamentals-documents/fundamentals/data-formats/documents

### Self-Review Checklist


source/fundamentals/data-formats/documents.txt
```diff
@@ -14,34 +14,351 @@ Documents
...
 --------
 
-- Description
-- Chart that shows mapping between BSON and Java types (similar to the one here https://mongodb.github.io/mongo-java-driver/4.1/bson/documents/)
-- Code example of declaring one and appending fields
-- Code example of get() for various types
-- Link to Fundamentals > CRUD for more information
-- Link to API documentation
+The ``Document`` class offers a flexible representation of a BSON document.
+You can access and manipulate document data using Java types with this
```
nathan-contino-mongo, line 67:
```suggestion
You can access and manipulate fields using Java types with this
```
Maybe this is a good opportunity to use the "field" term to familiarize readers? (see previous comment)

source/fundamentals/data-formats/documents.txt
```diff
@@ -14,34 +14,351 @@ Documents
...
+     - ``org.bson.Document``
+   * - Double
+     - ``java.lang.Double``
+   * - Int32
+     - ``java.lang.Integer``
+   * - Int64
+     - ``java.lang.Long``
+   * - Null
+     - ``null``
+   * - ObjectId``
```
nathan-contino-mongo, line 95:
```suggestion
   * - ObjectId
```
I think you accidentally a monospace!

source/fundamentals/data-formats/documents.txt
```diff
@@ -14,34 +14,351 @@ Documents
...
+     - ``org.bson.types.ObjectId``
+   * - String
+     - ``java.lang.String``
+
+You can also customize the type mapping by specifying a custom codec.
+For more information on customizing mapped types, see our guide on using
+:doc:`Codecs </fundamentals/data-formats/codecs>`.
+
+In the following code snippet, we show how to instantiate and build a sample
+``Document`` instance that contains a variety of data types:
```
nathan-contino-mongo, line 105: Thought: Is "contains a variety of data types" the best descriptor we can offer here? I wonder if this glosses over the fact that a document actually contains a collection of string keys that map to values with a variety of data types. Perhaps we can add some more detail about that fact here (and, my favorite bone to pick at this point, add the term "field" here)?

source/fundamentals/data-formats/documents.txt
```diff
@@ -14,34 +14,351 @@ Documents
...
 
-- Description
-- Chart that shows mapping between BSON and Java types (similar to the one here https://mongodb.github.io/mongo-java-driver/4.1/bson/documents/)
-- Code example of declaring one and appending fields
-- Code example of get() for various types
-- Link to Fundamentals > CRUD for more information
-- Link to API documentation
+The ``Document`` class offers a flexible representation of a BSON document.
+You can access and manipulate document data using Java types with this
+class. See the following table for mappings between frequently-used BSON
```
nathan-contino-mongo, line 68: nit: should "frequently-used" be "frequently used"? I tried to look it up in the meta style guide but I couldn't find it there.

source/fundamentals/data-formats/documents.txt
```diff
@@ -14,34 +14,351 @@ Documents
...
 
-- Description
-- Link to Fundamentals > Document Data Formats: BSON for more information
-- Code example of declaring one and appending fields
-- Link to Fundamentals > CRUD for more information
-- Link to API documentation
+The ``BsonDocument`` class provides a type-safe API to access and manipulate
+a BSON document. You need to specify the BSON type from the Java BSON
+library for each field. See the following table for mappings between
+frequently-used BSON and Java BSON library types:
```
nathan-contino-mongo, line 173: duplicate of an earlier nit: should "frequently-used" be "frequently used"? I tried to look it up in the meta style guide but I couldn't find it there.

source/fundamentals/data-formats/documents.txt
```diff
@@ -14,34 +14,351 @@ Documents
...
+     - ``org.bson.BsonInt64``
+   * - Null
+     - ``null``
+   * - ObjectId
+     - ``org.bson.BsonObjectId``
+   * - String
+     - ``org.bson.BsonString``
+
+
+You can also customize the type mapping by specifying a custom codec.
```
nathan-contino-mongo, line 205:
```suggestion
You can customize the type mapping by specifying a custom codec.
```
Nit: It seems to me like "also" is unnecessary here. What do you think?

source/fundamentals/data-formats/documents.txt
```diff
@@ -14,34 +14,351 @@ Documents
...
+   * - Int64
+     - ``java.lang.Long``
+   * - Null
+     - ``null``
+   * - ObjectId``
+     - ``org.bson.types.ObjectId``
+   * - String
+     - ``java.lang.String``
+
+You can also customize the type mapping by specifying a custom codec.
```
nathan-contino-mongo, line 100: Nit: It seems like "also" is unnecessary here. What do you think?
```suggestion
You can customize the type mapping by specifying a custom codec.
```

source/fundamentals/data-formats/documents.txt
```diff
@@ -14,34 +14,351 @@ Documents
...
+           new Document("title", "Love in the Time of Cholera").append("yearPublished", 1985)));
+
+You can insert this document into a collection using the following code:
+
+.. code-block:: java
+   :copyable: true
+
+  // MongoClient mongoClient = ...
+
+  MongoDatabase database = mongoClient.getDatabase("test_data");
```
nathan-contino-mongo, line 125: Because Document is the default representation for database documents, it seems like you're just relying on the defaults here. Should we call this out explicitly above?

source/fundamentals/data-formats/documents.txt
```diff
@@ -14,34 +14,351 @@ Documents
...
+
+You can insert this document into a collection using the following code:
+
+.. code-block:: java
+   :copyable: true
+
+   // MongoClient mongoClient = ...
+
+   MongoCollection<BsonDocument> collection = database.getCollection("authors", BsonDocument.class);
+   MongoDatabase database = mongoClient.getDatabase("fundamentals_data");
```
nathan-contino-mongo, line 233: I might be going crazy but I *think* these lines are in the wrong order!

source/fundamentals/data-formats/documents.txt
```diff
@@ -14,34 +14,351 @@ Documents
...
+       .append("novels", new BsonArray(Arrays.asList(
+           new BsonDocument().append("title", new BsonString("One Hundred Years of Solitude")).append("yearPublished", new BsonInt32(1967)),
+           new BsonDocument().append("title", new BsonString("Chronicle of a Death Foretold")).append("yearPublished", new BsonInt32(1981)),
+           new BsonDocument().append("title", new BsonString("Love in the Time of Cholera")).append("yearPublished", new BsonInt32(1985))
+      )));
+
+You can insert this document into a collection using the following code:
+
+.. code-block:: java
+   :copyable: true
```
nathan-contino-mongo, line 228: While you have specified that these blocks should be copyable, I don't see a copy button in your staged version. Nextgen bug?

source/fundamentals/data-formats/documents.txt
```diff
@@ -14,34 +14,351 @@ Documents
...
+Once you perform a successful insert, you can retrieve the sample document
+data from the collection using the following code:
+
+.. code-block:: java
+   :copyable: true
+
+   import com.mongodb.client.model.Filters;
+   ...
+
+   BsonDocument doc = collection.find(Filters.eq("novels.yearPublished", 1967)).first();
```
nathan-contino-mongo, line 246: I wonder if this query could confuse users who aren't hugely familiar with MongoDB, since we're using some fairly advanced ideas here (querying an embedded document with dot syntax, matching a document based off a match to one field in a single document in a collection). On one hand, this does demonstrate some interesting MongoDB edge cases, and a lot of readers might be curious about dot syntax. On the other hand... when I started out with MongoDB, I might have had some trouble following this. Perhaps it would be better to use a simpler filter here and instead cover these more advanced concepts in the CRUD section?

source/fundamentals/data-formats/documents.txt
```diff
@@ -14,34 +14,351 @@ Documents
...
+
+   _id: 5fb5fad05f734e3794741a35, name: Gabriel García Márquez, lastUpdated: Thu Mar 26 00:00:00 EDT 2020
+   title: One Hundred Years of Solitude, yearPublished: 1967
+   title: Chronicle of a Death Foretold, yearPublished: 1981
+   title: Love in the Time of Cholera, yearPublished: 1985
+
+For more information on retrieving and manipulating MongoDB data, see our
+:doc:`CRUD guide </fundamentals/crud>`.
+
+For more information on this class, see the
```
nathan-contino-mongo, line 271: Duplicate: perhaps a good use for a see also directive?

source/fundamentals/data-formats/documents.txt
```diff
@@ -14,34 +14,351 @@ Documents
...
 -------------
 
-- Description
-- Code example of declaring one and appending fields
-- Link to Fundamentals > CRUD for more information
-- Link to API documentation
+The ``BasicDBObject`` class allows you to access and manipulate document data
+using Java types. We recommend that you avoid using this class unless you
+are migrating an application from that uses an older driver version. In
+addition to deprecation, this class has the following limitations:
```
nathan-contino-mongo, line 280: Instead of the mildly awkward "in addition to deprecation" phrasing, I wonder if we could:
1. Add (Deprecated) after the section header, akin to the (BETA) label for Realm Sync
2. Mention "deprecated" in the first sentence ("The deprecated ``BasicDBObject`` class") to make the deprecation even more obvious.

We could then just change this sentence to "This class has the following limitations" and, if we really want to hammer in the deprecation idea, add "This class is deprecated" as a new first bullet point in the limitations.

source/fundamentals/data-formats/documents.txt
```diff
@@ -14,34 +14,351 @@ Documents
...
+
+- This class implements the ``DBObject`` interface, so you cannot extend its
+  API without breaking binary compatibility.
```
nathan-contino-mongo, line 283: Perhaps I'm just not familiar with "binary compatibility" or API extension constraints, but it's unclear to me how the first clause relates to the second clause (I'm assuming a causal relationship). It could also be helpful to define "binary compatibility" here (are we talking about binaries as in programs? Or binary representation of objects in process memory?

source/fundamentals/data-formats/documents.txt
```diff
@@ -14,34 +14,351 @@ Documents
...
+       .append("name", "Gabriel García Márquez")
+       .append("lastUpdated", new Date(120, 2, 26))
+       .append("novels", Arrays.asList(
+           new Document("title", "One Hundred Years of Solitude").append("yearPublished", 1967),
+           new Document("title", "Chronicle of a Death Foretold").append("yearPublished", 1981),
+           new Document("title", "Love in the Time of Cholera").append("yearPublished", 1985)));
+
+You can insert this document into a collection using the following code:
+
+.. code-block:: java
```
nathan-contino-mongo, line 120: For some reason, syntax highlighting doesn't seem to be working for this example.

source/fundamentals/data-formats/documents.txt
nathan-contino-mongo: The equivalent code snippets for other document representations use the "test_fundamentals" database instead of "test_data". Should we standardize the database used?

source/fundamentals/data-formats/documents.txt
```diff
@@ -14,34 +14,351 @@ Documents
...
+
+       BasicDBList novels = (BasicDBList) doc.get("novels");
+       if (novels != null) {
+           BasicDBObject[] novelArr = novels.toArray(new BasicDBObject[0]);
+           for (BasicDBObject novel : novelArr) {
+               System.out.println("title: " + novel.getString("title")
+               + ", yearPublished: " + novel.getInt("yearPublished"));
+           }
+       }
+   }
```
nathan-contino-mongo, line 336: Similar thoughts as I had for the Document class -- should we explain the usage and constraints of the BasicDBObject type getters?

source/fundamentals/data-formats/documents.txt
```diff
@@ -14,34 +14,351 @@ Documents
...
+Summary
+-------
+
+In this guide, we covered the following topics on Java document classes:
+
+- Described three Java classes you can use to work with MongoDB documents and
+  why you might prefer one over the other.
+- Provided usage examples for each class on building documents containing
+  multiple types, inserting them into a collection, and
+  retrieving/accessing their typed fields.
```
nathan-contino-mongo, line 363: Nice summary, I really like this.

nathan-contino-mongo (commented):
I think this is a great start -- a lot of comments from me, but mostly suggestions of how we can enhance this guide to make it even better. This is a tricky concept to cover and I think we're definitely moving in the right direction!

source/fundamentals/data-formats/documents.txt
```diff
@@ -14,34 +14,351 @@ Documents
...
 
-- Definition - what is a document, what format is it, where to find more info 
-- Table comparison of classes (columns for: implements Map, provides type safety, is an interface or class, compatibility with server versions
-- Recommend Document for general use and why.
-- Duplicate field name admonition (I forgot what this is)
+A MongoDB **document** is a data structure that contains key/value pairs in
+binary JSON (BSON) format and is the base unit in the MongoDB data hierarchy.
+
+Each **document** belongs to a single **collection** and each **collection**
+belongs to a single **database**. For more information, see our MongoDB manual
```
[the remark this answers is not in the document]
ccho-mongodb: Thanks for bringing this up.

1) I think I could replace this with `:term:` (see Glossary terms in https://docs.mongodb.com/meta/style-guide/style/text-formatting/).
Italics could be used for introducing new terms (see "New terms" in https://docs.mongodb.com/meta/style-guide/style/text-formatting/ ) explained immediately afterwards

2) I did think about creating a graphic and searched through server/manual documentation and wasn't able to find anything suitable to link to or re-use. Maybe I can't find the right page which probably means our users can't find it either. Instead of attempting to explain it from scratch, I just wanted to introduce repetition of terms and what they are as a learning aid.

That said, if this was a language-specific concept, I would probably devote a section for it.

source/fundamentals/data-formats/documents.txt
```diff
@@ -14,34 +14,351 @@ Documents
...
+
+   * - ``BsonDocument``
+     - ``org.bson``
+     - Yes, implements ``Map<String, BsonValue>``
+     - When you need a type-safe API.
+
+   * - ``BasicDBObject``
+     - ``com.mongodb``
+     - No
+     - When you are migrating your application from a legacy driver version.
```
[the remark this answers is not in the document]
ccho-mongodb: Thanks, I wonder if this should be a pattern we reuse for pages that present multiple similar topics.

source/fundamentals/data-formats/documents.txt
ccho-mongodb: Whoops, good catch. Even the autobuilder did not warn me about that.

source/fundamentals/data-formats/documents.txt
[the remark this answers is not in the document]
nathan-contino-mongo: I think this is a perfect opportunity for creating a glossary, so definitely think about that! Just check to make sure it's supported properly by nextgen first :)

I hoped that the linked server page would include such a graphic, too, but sadly it doesn't. Might be a good idea to ask in the #docs chat to see if anybody knows of such a graphic (or would be willing to create one for you, if they have a LucidChart license and some degree of artistic skills)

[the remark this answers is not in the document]
nathan-contino-mongo: +1 to repetition as a learning aid, I think that works quite well here.

source/fundamentals/data-formats/documents.txt
```diff
@@ -14,34 +14,351 @@ Documents
...
+           new Document("title", "One Hundred Years of Solitude").append("yearPublished", 1967),
+           new Document("title", "Chronicle of a Death Foretold").append("yearPublished", 1981),
+           new Document("title", "Love in the Time of Cholera").append("yearPublished", 1985)));
+
+You can insert this document into a collection using the following code:
+
+.. code-block:: java
+   :copyable: true
+
+  // MongoClient mongoClient = ...
```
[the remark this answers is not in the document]
ccho-mongodb: It was designed to shows the type with the instantiation as the variable is used later on. I can certainly add more explanation.

source/fundamentals/data-formats/documents.txt
```diff
@@ -14,34 +14,351 @@ Documents
...
+  InsertOneResult result = collection.insertOne(author);
+
+Once you perform a successful insert, you can retrieve the sample document
+data from the collection using the following code:
+
+.. code-block:: java
+   :copyable: true
+
+   import com.mongodb.client.model.Filters;
+   ...
```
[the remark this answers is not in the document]
ccho-mongodb: From [this commen](https://github.com/mongodb/docs-java/pull/39#discussion_r527092414)t, I need to add some explanation of the dots.

source/fundamentals/data-formats/documents.txt
```diff
@@ -14,34 +14,351 @@ Documents
...
+   _id: 5fb5fad05f734e3794741a35, name: Gabriel García Márquez, lastUpdated: Thu Mar 26 00:00:00 EDT 2020
+   title: One Hundred Years of Solitude, yearPublished: 1967
+   title: Chronicle of a Death Foretold, yearPublished: 1981
+   title: Love in the Time of Cholera, yearPublished: 1985
+
+For more information on retrieving and manipulating MongoDB data, see our
+:doc:`CRUD guide </fundamentals/crud>`.
+
+For more information on this class, see the
+:java-docs:`Document API documentation </apidocs/bson/org/bson/Document.html>`.
```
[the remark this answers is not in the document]
ccho-mongodb: Thanks for the suggestion. I think you mean `seealso`. I've been avoiding using it, but I think it doesn't really change the experience either way. I'll keep it in mind though since the change in formatting it produces could be a good signal to readers if used consistently.

source/fundamentals/data-formats/documents.txt
ccho-mongodb: I think usually you use hyphen for adverb + past participle verb, but maybe open for debate.
There's some discussion here:
https://english.stackexchange.com/questions/225433/should-i-used-a-hyphen-for-often-used-and-well-documented

I'm going to leave it _as-is_ for now, unless there's a pressing need to change it.

nathan-contino-mongo: Sounds good to me!

source/fundamentals/data-formats/documents.txt
ccho-mongodb: Thanks!

source/fundamentals/data-formats/documents.txt
ccho-mongodb: I was also waffling with the wording of this sentence. I think I might need to provide more detail as I was hoping I could keep it concise and that the reader could infer that the table showed default mappings and that they could override them by specifying or implementing a custom codec.

source/fundamentals/data-formats/documents.txt
ccho-mongodb: Will investigate. When I last checked, the first code blocks I created and previewed were working fine, but maybe I introduced some syntax issue along the way

source/fundamentals/data-formats/documents.txt
ccho-mongodb: Yeah, good copypasta catch.

source/fundamentals/data-formats/documents.txt
```diff
@@ -14,34 +14,351 @@ Documents
 Overview
 --------
 
-- Definition - what is a document, what format is it, where to find more info 
-- Table comparison of classes (columns for: implements Map, provides type safety, is an interface or class, compatibility with server versions
-- Recommend Document for general use and why.
-- Duplicate field name admonition (I forgot what this is)
+A MongoDB **document** is a data structure that contains key/value pairs in
+binary JSON (BSON) format and is the base unit in the MongoDB data hierarchy.
```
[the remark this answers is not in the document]
ccho-mongodb: I'll replace this with something similar.

source/fundamentals/data-formats/documents.txt
```diff
@@ -14,34 +14,351 @@ Documents
...
+While you can use any of the document classes in your application, we
+recommend that you use the ``Document`` class since it can concisely
+represent dynamically structured documents of any complexity. It
+implements the ``Map<String, Object>`` interface which enables it to use
+loosely-typed values.
+
+.. warning:: Duplicate Field Names
+
+   Although a document is valid if it contains two duplicate keys, the driver
+   only sets the value of the key to the value in the last instance.
```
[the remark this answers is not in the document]
ccho-mongodb: I don't have any qualms with adding this info, but "field" is included in the linked doc: https://docs.mongodb.com/manual/core/document/#field-names

I'll make sure I explain that the link has important info about field-related terminology and limitations, and remove this warning. I think the reason it is called out in the current version of the driver docs is because they don't link to the manual entry: https://mongodb.github.io/mongo-java-driver/4.1/bson/documents/

source/fundamentals/data-formats/documents.txt
ccho-mongodb: Appending the fix to another commit.

source/fundamentals/data-formats/documents.txt
ccho-mongodb: > Because Document is the default representation for database documents, it seems like you're just relying on the defaults here. Should we call this out explicitly above?

I'm guessing you meant this on the next line for `getCollection`. I'll add a mention and link to the API docs.

source/fundamentals/data-formats/documents.txt
```diff
@@ -14,34 +14,351 @@ Documents
...
+   if (doc != null) {
+       System.out.println("_id: " + doc.getObjectId("_id")
+           + ", name: " + doc.getString("name")
+           + ", lastUpdated: " + doc.getDate("lastUpdated"));
+
+       doc.getList("novels", Document.class).forEach((novel) -> {
+           System.out.println("title: " + novel.getString("title")
+               + ", yearPublished: " + novel.getInteger("yearPublished"));
+           });
+   }
```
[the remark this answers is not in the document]
ccho-mongodb: >  include a sentence or two clarifying how you access field values from an Instance of Document

Could you clarify what you mean by this?  I included methods called on the Document instance to access the fields such as getObjectId, getString, and getDate. I also showed how to access other types within an embedded Document.

For the second part:
>what happens if you use the wrong type of get (for instance, getString on an ObjectId field)

I'll think about it how I might incorporate that info.

source/fundamentals/data-formats/documents.txt
ccho-mongodb: Thought it was deprecated, but I guess not.

source/fundamentals/data-formats/documents.txt
ccho-mongodb: I think you might have been expecting a description of a Document class. I was actually attempting to communicate that the example intentionally uses field values of different types to demonstrate how to set them. I'll clarify that here. 

Regarding data structures contained within the Document class: I want to limit the scope of what is discussed in reference documentation since the reader can refer to API documentation for those details.

source/fundamentals/data-formats/documents.txt
ccho-mongodb: I copied this from the existing docs, but I'm assuming it is referring to https://wiki.eclipse.org/Evolving_Java-based_APIs_2#Evolving_API_packages -- see example 4. (A more recent doc: https://docs.oracle.com/javase/specs/jls/se7/html/jls-13.html )

As I understand it, "extend its API" means add a method to a class that implements it such as BasicDBObject. Suppose you had a compiled app that uses driver v4.1, but v4.2 came out with a new method, BasicDBObject#coolNewMethod. Normally, you could just upgrade the jar in the classpath and restart the app (or hot-reload) and it would work fine.  However, since you broke binary compatibility with v4.2 by adding a method, you need to recompile your app for it to work with v4.2. However, if there was some non-breaking change such as a private method change, your app would continue to work without recompilation because it's not visible to your binary.

Admittedly, it is a little unclear which object the pronoun "its" refers to so I can reword. It's also probably less frequently encountered than the next item, so I'll re-order it in the list.

source/fundamentals/data-formats/documents.txt
ccho-mongodb: Addressed in a note directive

source/fundamentals/data-formats/documents.txt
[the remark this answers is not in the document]
ccho-mongodb: Didn't realize you need a glossary to use `:term:`. I removed formatting for the time being. I'm a little confused by the inconsistent formatting here, so probably something to address sometime later: https://github.com/mongodb/docs-java/blob/master/source/fundamentals/databases-collections.txt

source/fundamentals/data-formats/documents.txt
```diff
@@ -14,34 +14,379 @@ Documents
...
+on using :doc:`Codecs </fundamentals/data-formats/codecs>`.
+
+In the following code snippet, we show how to instantiate and build a sample
+``Document`` instance representing a document containing several
+different field types:
+
+.. code-block:: java
+
+   Document author = new Document("_id", new ObjectId())
+       .append("name", "Gabriel García Márquez")
```
MongoCaleb, line 107: Nice author reference!

source/fundamentals/data-formats/documents.txt
```diff
@@ -14,34 +14,379 @@ Documents
...
+
+   The code sample above uses helper methods that check the returned type
+   and throw a :java-docs:`BsonInvalidOperationException </apidocs/bson/org/bson/BsonInvalidOperationException.html>`
+   if it is unable to cast the field value. You can call the
+   :java-docs:`get() </apidocs/bson/org/bson/BsonDocument.html#get(java.lang.Object)>`
+   method specified by the ``Map`` interface to retrieve field values as
+   type :java-docs:`BsonValue </apidocs/bson/org/bson/BsonValue.html>` and to
+   skip type checking.
+
+The output should resemble the following:
```
MongoCaleb, line 275: "will" instead of "should"?

source/fundamentals/data-formats/documents.txt
```diff
@@ -14,34 +14,379 @@ Documents
...
+   }
+
+.. note::
+
+   The code sample above uses helper methods that check the returned type
+   and throw an exception if it is unable to cast the field value.
+   You can call the :java-docs:`get() </apidocs/bson/org/bson/BasicBSONObject.html?is-external=true#get(java.lang.String)>`
+   method to retrieve values as type type ``Object`` and to skip type checking.
+
+The output should resemble the following:
```
MongoCaleb, line 366: "will" instead of "should"?

source/fundamentals/data-formats/documents.txt
```diff
@@ -14,34 +14,379 @@ Documents
...
+
+.. note::
+
+   The code sample above uses helper methods that check the returned type
+   and throw an exception if it is unable to cast the field value.
+   You can call the :java-docs:`get() </apidocs/bson/org/bson/Document.html#get(java.lang.Object)>`
+   method specified by the ``Map`` interface to retrieve field values as
+   type ``Object`` and to skip type checking.
+
+The output should resemble the following:
```
MongoCaleb, line 156: "will" instead of "should"?

MongoCaleb (approved):
LGTM with one suggestion that occurs 3x

source/fundamentals/data-formats/documents.txt
ccho-mongodb: I avoid using "will" across docs because it often leads to passive voice and I think we prefer active when possible based on some feedback I received last year. Do you think that I should change it anyway?

ccho-mongodb:
Hi @jyemin and @rozza , ready for your review whenever you get the chance. Thanks!

source/fundamentals/data-formats/documents.txt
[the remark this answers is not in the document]
nathan-contino-mongo: I very much like the note that you added after these examples -- that's precisely what I was suggesting above. Resolved!

source/fundamentals/data-formats/documents.txt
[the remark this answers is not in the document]
nathan-contino-mongo: I think that your solution is solid here -- it gives just enough of a hint that a reader can google for a solution (or search on the docs, or navigate within the docs) if they don't already know how to instantiate a mongoClient, mongoCollection, etc.

nathan-contino-mongo (approved):
LGTM -- this page is really coming together! Ready for tech review.

rozza (approved):
LGTM.

I'm not sold on having DBObjects listed but I suppose people are still using them! :cry:

source/fundamentals/data-formats/documents.txt
```diff
@@ -14,34 +14,379 @@ Documents
 Overview
 --------
 
-- Definition - what is a document, what format is it, where to find more info 
-- Table comparison of classes (columns for: implements Map, provides type safety, is an interface or class, compatibility with server versions
-- Recommend Document for general use and why.
-- Duplicate field name admonition (I forgot what this is)
+A MongoDB **document** is a data structure that contains key/value pairs in
+binary JSON (BSON) format. The document is the base unit of the MongoDB
+data hierarchy: each document belongs to a single collection
```
jyemin, line 19: > The document is the base unit of the MongoDB	data hierarchy

I feel like you're inventing new terminology here: I've never heard a document described this way in our documentation.  And it's not accurate.  The base unit is a BSON _value_ (e.g. int64, int32, boolean), not a BSON document.

And also, documents are used for other purposed besides storage in collections: e.g. commands, queries, sort criteria, etc

source/fundamentals/data-formats/documents.txt
```diff
@@ -14,34 +14,379 @@ Documents
...
 --------
 
-- Description
-- Chart that shows mapping between BSON and Java types (similar to the one here https://mongodb.github.io/mongo-java-driver/4.1/bson/documents/)
-- Code example of declaring one and appending fields
-- Code example of get() for various types
-- Link to Fundamentals > CRUD for more information
-- Link to API documentation
+The ``Document`` class offers a flexible representation of a BSON document.
+You can access and manipulate fields using Java types with this class. See
```
jyemin, line 63: "using Java types from the standard library"  would be more accurate.  BsonDocument also uses Java types, but ones that are defined in our own bson library

source/fundamentals/data-formats/documents.txt
```diff
@@ -14,34 +14,379 @@ Documents
...
+
+In the following code snippet, we show how to instantiate and build a sample
+``Document`` instance representing a document containing several
+different field types:
+
+.. code-block:: java
+
+   Document author = new Document("_id", new ObjectId())
+       .append("name", "Gabriel García Márquez")
+       .append("lastUpdated", new Date(120, 2, 26))
```
jyemin, line 108: That's a strange date... The year 120?  

Also, that constructor is deprecated.  Horribly, the way to do this in Java is:

    Date.from(LocalDate.of(2000, 1, 1).atStartOfDay(ZoneId.systemDefault()).toInstant())

But since you're just modeling a lastUpdated field, why not just `new Date())`?

source/fundamentals/data-formats/documents.txt
```diff
@@ -14,34 +14,379 @@ Documents
...
+   * - Document
+     - ``org.bson.BsonDocument``
+   * - Double
+     - ``org.bson.BsonDouble``
+   * - Int32
+     - ``org.bson.BsonInt32``
+   * - Int64
+     - ``org.bson.BsonInt64``
+   * - Null
+     - ``null``
```
jyemin, line 203: `null` is actually not allowed.  Use the `org.bson.BsonNull` type

source/fundamentals/data-formats/documents.txt
```diff
@@ -14,34 +14,379 @@ Documents
...
+   * - ObjectId
+     - ``org.bson.BsonObjectId``
+   * - String
+     - ``org.bson.BsonString``
+
+The mapping table above shows the default mapping when working with the
```
jyemin, line 209: You should not really customize the mapping for BsonDocument.  Let's remove this paragraph.

source/fundamentals/data-formats/documents.txt
```diff
@@ -14,34 +14,379 @@ Documents
...
+In the following code snippet, we show how to instantiate and build a sample
+``BsonDocument`` instance representing a document containing several
+different field types:
+
+.. code-block:: java
+
+   BsonDocument author = new BsonDocument()
+       .append("_id", new BsonObjectId())
+       .append("name", new BsonString("Gabriel García Márquez"))
+       .append("lastUpdated", new BsonDateTime(new Date(120, 2, 26).getTime()))
```
jyemin, line 223: As above: just use `new Date()`

source/fundamentals/data-formats/documents.txt
```diff
@@ -14,34 +14,379 @@ Documents
...
+   method specified by the ``Map`` interface to retrieve field values as
+   type :java-docs:`BsonValue </apidocs/bson/org/bson/BsonValue.html>` and to
+   skip type checking.
+
+The output should resemble the following:
+
+.. code-block:: none
+   :copyable: false
+
+   _id: 5fb5fad05f734e3794741a35, name: Gabriel García Márquez, lastUpdated: Thu Mar 26 00:00:00 EDT 2020
```
jyemin, line 280: I'm surprised you got 2020 given the date you used above

source/fundamentals/data-formats/documents.txt
```diff
@@ -14,34 +14,379 @@ Documents
...
+
+In the following code snippet, we show how to instantiate and build a sample
+``BasicDBObject`` instance representing a document containing several
+different field types:
+
+.. code-block:: java
+
+   BasicDBObject author = new BasicDBObject("_id", new ObjectId())
+       .append("name", "Gabriel García Márquez")
+       .append("lastUpdated", new Date(120, 2, 26))
```
jyemin, line 315: As above

source/fundamentals/data-formats/documents.txt
```diff
@@ -14,34 +14,379 @@ Documents
...
+For more information on retrieving and manipulating MongoDB data, see our
+:doc:`CRUD guide </fundamentals/crud>`.
+
+For more information on this class, see the
+:java-docs:`BasicDBObject API documentation </apidocs/mongodb-driver-core/com/mongodb/BasicDBObject.html>`.
+
+Summary
+-------
+
+In this guide, we covered the following topics on Java document classes:
```
jyemin, line 385: `Java document classes` is kind of awkward.  It's Java driver classes, not Java classes.  And they're not document classes, they're classes used to model/represent BSON documents.

Can you rephrase?

jyemin (changes requested):

source/fundamentals/data-formats/documents.txt
ccho-mongodb: Good point, I'll remove the phrase that mentions it's the base unit.

source/fundamentals/data-formats/documents.txt
ccho-mongodb: Thanks, will update.

source/fundamentals/data-formats/documents.txt
ccho-mongodb: Thanks, haven't kept up with Java 8+ dates. I'll change the field name and purpose since I want to show an example of setting a specific date.

source/fundamentals/data-formats/documents.txt
ccho-mongodb: Thanks, didn't see that type. Will update.

source/fundamentals/data-formats/documents.txt
ccho-mongodb: Will remove.

source/fundamentals/data-formats/documents.txt
ccho-mongodb: Will update for new field.

source/fundamentals/data-formats/documents.txt
ccho-mongodb: Will update for new field.

source/fundamentals/data-formats/documents.txt
ccho-mongodb: Will update for new field.

source/fundamentals/data-formats/documents.txt
ccho-mongodb: Sure, found a couple other similar mentions in this PR, will update.

ccho-mongodb:
Ready for re-review @jyemin 
-Updated Date instantiation
-Updated date field name in the sample document and output
-Removed text about customizing codec for BsonDocument
-Updated null to BsonNull type
-Clarified Java types from standard library
-Removed mention of document classes
-Removed mention of document as base unit in data hierarchy

source/fundamentals/data-formats/documents.txt
```diff
@@ -15,15 +15,14 @@ Overview
 --------
 
 A MongoDB **document** is a data structure that contains key/value pairs in
-binary JSON (BSON) format. The document is the base unit of the MongoDB
-data hierarchy: each document belongs to a single collection
+binary JSON (BSON) format. Each document belongs to a single collection
```
jyemin, line 18: I still think this is misleading since we use documents for queries, commands, etc.  Those don't belong to a collection.

ccho-mongodb: Ok thanks, I'll update the sentence. My intent is to build a little more context beyond the first sentence without having the user follow the link. If you think that's extraneous or if there's some other essential info you'd prefer to include instead, let me know.

ccho-mongodb:
Addressed https://github.com/mongodb/docs-java/pull/39#discussion_r530652834 
Ready for re-review.

jyemin (approved):

PR #45, opened 2020-12-03
DOCSP-9557: Fundamentals > CRUD Operations > Read Operations > Sort Results

ksadoff:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-9557

https://docs-mongodborg-staging.corp.mongodb.com/960a4dd/java/docsworker-xlarge/DOCSP-9557/fundamentals/crud/sort

### Self-Review Checklist

### DIFF
--- a/source/fundamentals/crud/sort.txt
+++ b/source/fundamentals/crud/sort.txt
@@ -12,9 +12,6 @@ To sort in descending (greatest first) order instead, use the static method ``So
 If you do not specify a sort, MongoDB does not guarantee the order of
 query results.
 
-For more information on the ``sort()`` method, 
-see the :docs-java:`API documentation </apidocs/mongodb-driver-sync/com/mongodb/client/FindIterable.html#sort(org.bson.conversions.Bson)>`.
-
 Follow the instructions in the examples below to insert data into
 a collection and perform a sort on the results of a query.
 Consider a collection containing documents that describe books. To
@@ -113,3 +110,10 @@ the documents matching the query, sorting "A Dance with Dragons" before
     Document{{_id=6, title=A Dance with Dragons, length=1104, author=Martin}}
     Document{{_id=4, title=Infinite Jest, length=1104, author=Wallace}}
     Document{{_id=2, title=Les Misérables, length=1462, author=Hugo}}
+
+For additional information on the classes and methods mentioned on this
+page, see the following API documentation:
+
+- :java-sync-api:`FindIterable <com/mongodb/client/FindIterable.html>`
+- :java-sync-api:`MongoCursor <com/mongodb/client/MongoCursor.html>`
+- :java-sync-api:`Sorts <com/mongodb/client/model/Sorts.html>`
--- a/source/fundamentals/crud/sort.txt
+++ b/source/fundamentals/crud/sort.txt
@@ -73,11 +73,11 @@ Sometimes, the order of two or more documents is ambiguous using a
 specified sort. Consider the case of sorting by length: in the above case, 
 both "A Dance with Dragons" and "Infinite Jest" have ``1104`` pages, so 
 the order in which ``sort()`` returns them is not guaranteed. To resolve ties in 
-your sorted results in a repeatable way, add additional fields by which to sort to the 
+your sorted results in a repeatable way, add additional fields to the 
 ``Sorts.ascending()`` function call. 
 Call the ``sort()`` method, specifying ``length`` as 
 the first field by which to sort and the ``title`` field as the second field. In the event 
-that the length field comparison results in a tie, the operation compares the tied results 
+that the ``length`` field comparison results in a tie, the operation compares the tied results 
 using the ``title`` field:
 
 
@@ -89,8 +89,8 @@ using the ``title`` field:
         .sort(ascending("length", "title")).iterator();
 
 This sample code prints your query results in ascending order by length. 
-When there are two documents with the same value, it resolves the tie using the 
-title field you specified as the second parameter.
+When there are two documents with the same ``length`` value, it resolves the tie using the 
+``title`` field you specified as the second parameter.
 
 .. code-block:: javascript
 
--- a/source/fundamentals/crud/sort.txt
+++ b/source/fundamentals/crud/sort.txt
@@ -116,4 +116,4 @@ page, see the following API documentation:
 
 - :java-sync-api:`FindIterable <com/mongodb/client/FindIterable.html>`
 - :java-sync-api:`MongoCursor <com/mongodb/client/MongoCursor.html>`
-:java-docs:`Sorts </apidocs/mongodb-driver-core/com/mongodb/client/model/Sorts.html>`
+- :java-docs:`Sorts </apidocs/mongodb-driver-core/com/mongodb/client/model/Sorts.html>`
--- a/source/fundamentals/crud/sort.txt
+++ b/source/fundamentals/crud/sort.txt
@@ -67,7 +67,7 @@ on "title" and an ascending order:
 .. note::
 
    These results are in JSON format due to the ``toJson()`` method used in the example above.
-   Without ``toJson()``, they would be returned in a Java string format, 
+   Without ``toJson()``, they are returned in a Java string format, 
    like this: ``Document{{_id=6, title=A Dance with Dragons, author=Martin, length=1104}}``.
 
 Sometimes, the order of two or more documents is ambiguous using a
@@ -95,8 +95,8 @@ your sorted results in a repeatable way, add additional parameters to the
 With the addition of the ``title`` field to the sort document, the read
 operation sorts matching documents first by ``length`` and, in the event
 of a tie, then by ``title``. Matched document fields are compared in
-the same order as fields are specified in the sort document. The ``find()``
-operation returns the following ordering of documents when this sort is used on
+the same order as fields are specified in the sort document. The ``sort``
+operation returns the following ordering of documents when used on
 the documents matching the query, sorting "A Dance with Dragons" before 
 "Infinite Jest" for the two books with the same length:
 
--- a/source/fundamentals/crud/sort.txt
+++ b/source/fundamentals/crud/sort.txt
@@ -47,7 +47,7 @@ operation returns books in alphabetical order by title:
     // print out items in alphabetical order
     try {
       while (cursor.hasNext()) {
-        System.out.println(cursor.next().toJson());
+        System.out.println(cursor.next());
       }
     // close the cursor
     } finally {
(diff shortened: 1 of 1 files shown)


source/fundamentals/crud/sort.txt
```diff
@@ -0,0 +1,110 @@
...
+
+.. code-block:: java
+
+   collection.insertMany([
+        new Document().append("_id", 1)
+            .append("title", "The Brothers Karamazov").append("length", 824).append("author", "Dostoyevksy"),
+        new Document().append("_id", 2)
+            .append("title", "Les Misérables").append("length", 1462).append("author", "Hugo"),
+        new Document().append("_id", 3)
+            .append("title", "Atlas Shrugged).append("length", 1088).append("author", "Rand"),
```
terakilobyte, line 28:
```suggestion
            .append("title", "Atlas Shrugged").append("length", 1088).append("author", "Rand"),
```
Missing a closing quote. Copying and pasting this makes IntelliJ very mad.

source/fundamentals/crud/sort.txt
```diff
@@ -0,0 +1,110 @@
...
+        new Document().append("_id", 4)
+            .append("title", "Infinite Jest").append("length", 1104).append("author", "Wallace"),
+        new Document().append("_id", 5)
+            .append("title", "Cryptonomicon").append("length", 918).append("author", "Stephenson"),
+        new Document().append("_id", 6)
+            .append("title", "A Dance with Dragons").append("length", 1104).append("author", "Martin"),
+   ]);
```
terakilobyte, line 35:
```suggestion
        collection.insertMany(new ArrayList<Document>(Arrays.asList(
                new Document().append("_id", 1)
                        .append("title", "The Brothers Karamazov").append("length", 824).append("author", "Dostoyevksy"),
                new Document().append("_id", 2)
                        .append("title", "Les Misérables").append("length", 1462).append("author", "Hugo"),
                new Document().append("_id", 3)
                        .append("title", "Atlas Shrugged").append("length", 1088).append("author", "Rand"),
                new Document().append("_id", 4)
                        .append("title", "Infinite Jest").append("length", 1104).append("author", "Wallace"),
                new Document().append("_id", 5)
                        .append("title", "Cryptonomicon").append("length", 918).append("author", "Stephenson"),
                new Document().append("_id", 6)
                        .append("title", "A Dance with Dragons").append("length", 1104).append("author", "Martin")
        )));
```
I'm unsure this is valid Java syntax for array declarations? Also, I was unable to use this code to insert documents even with adding the missing quote. `insertMany` appears to want a List like item.

terakilobyte (changes requested):
The content is good, but the code doesn't work.

source/fundamentals/crud/sort.txt
nathan-contino-mongo, line 28:
```suggestion
            .append("title", "Atlas Shrugged").append("length", 1088).append("author", "Rand"),
```
The syntax highlighting for this code block looked absolutely crazy to me, until I realized that there was a missing quote 🤭

terakilobyte (approved):
LGTM. I do not know if my `Arrays.asList` approach is the most idiomatic for doing this insert but everything  works for me and is understandable.

ksadoff:
> LGTM. I do not know if my `Arrays.asList` approach is the most idiomatic for doing this insert but everything works for me and is understandable.

Looks like ``Arrays.asList`` is how it's done in the ``insertMany`` [code example](https://github.com/mongodb/docs-java/blob/master/source/includes/usage-examples/code-snippets/InsertMany.java), so presumably it's okay to use for this example as well.

ccho-mongodb (commented):
This looks pretty good, better than my first two DevEd PRs. Here are a few items that would be good to address before moving to an external review:

Always strongly prefer active voice over passive voice (here's an explanation: https://www.grammarly.com/blog/passive-voice/?gclid=CjwKCAiAwrf-BRA9EiwAUWwKXueo3fRQODjAloka31wFcqWB0uTNblside8QzxtgY5aq342X89Y7JBoCzhYQAvD_BwE&gclsrc=aw.ds). I think if you use a grammar checker like writegood, you should be able to identify potential instances quickly. One example of this in your current PR is "when sort is used with a query" on line 58 and "Matched document fields are compared (line 95).

-When referring to a method/function with preformatted text, add the "()" at the end to stay consistent with the rest of the Java documentation (e.g. source/usage-examples/find.txt). Also, unless it has been mentioned prior, specify whether it's a static or instance method.

Make sure to link to the API documentation for the methods that you reference. For example, :java-docs:`sort() </apidocs/mongodb-driver-sync/com/mongodb/client/FindIterable.html#sort(org.bson.conversions.Bson)>`
https://mongodb.github.io/mongo-java-driver/4.1/apidocs/mongodb-driver-sync/com/mongodb/client/FindIterable.html#sort(org.bson.conversions.Bson)

Make sure to explain new terms such as "sort document" (line 93). Is it actually a sort document or a call to the sort method?

Stay consistent with text formatting (e.g. line 59 "title") to prevent confusion.

source/fundamentals/crud/sort.txt
```diff
@@ -0,0 +1,119 @@
...
+    Document{{_id=4, title=Infinite Jest, length=1104, author=Wallace}}
+    Document{{_id=2, title=Les Misérables, length=1462, author=Hugo}}
+    Document{{_id=1, title=The Brothers Karamazov, length=824, author=Dostoyevsky}}
+
+Sometimes, the order of two or more documents is ambiguous using a
+specified sort. Consider the case of sorting by length: in the above case, 
+both "A Dance with Dragons" and "Infinite Jest" have ``1104`` pages, so 
+the order in which ``sort()`` returns them is not guaranteed. To resolve ties in 
+your sorted results in a repeatable way, add additional parameters to the 
+``Sorts.ascending()`` function call:
```
ccho-mongodb, line 77: Make this parallel in structure to the prior sample description:

```suggestion
``Sorts.ascending()`` function call.

Call the `sort()``method, specifying the length as the first field by which to sort and the length field as the second field. In the event that the length field comparison results in a tie, the operation compares the tied results using the title field:
```

source/fundamentals/crud/sort.txt
```diff
@@ -0,0 +1,119 @@
...
+      cursor.close();
+    }
+
+With the addition of the ``title`` field to the ``sort()`` call, the
+operation sorts matching documents first by ``length`` and, in the event
+of a tie, then by ``title``. The sort operation compares matched document fields in
+the same order as they are specified in the function call. The ``sort()``
+operation returns the following ordering of documents when used on
+the documents matching the query, sorting "A Dance with Dragons" before 
+"Infinite Jest" for the two books with the same length:
```
ccho-mongodb, line 103:
```suggestion
This sample code prints your query results in ascending order by length. When there are two documents with the same value, it resolves the tie using the title field you specified as the second parameter.
```

ccho-mongodb (commented):
I've added some suggestions which I think may help clarify some of the parts that aren't as easy to explain. I also made suggestions to make the explanatory text more concise. Let me know what you think.

ccho-mongodb (approved):
LGTM!

ksadoff:
Added @jyemin and @rozza as reviewers, but only one review is necessary

source/fundamentals/crud/sort.txt
```diff
@@ -0,0 +1,109 @@
...
+            .append("author", "Martin")
+    ));
+
+Call the ``sort()`` function and specify the order with ``ascending()`` to ensure that the
+operation returns books in alphabetical order by title:
+
+.. code-block:: java
+  :emphasize-lines: 5
+
+    import com.mongodb.client.model.Sorts;
```
jyemin, line 46: I don't think this is enough to make `ascending` work below.  You either have to [statically import](https://docs.oracle.com/javase/8/docs/technotes/guides/language/static-import.html) `Sorts.ascending` or use `Sorts.ascending` in the code.

source/fundamentals/crud/sort.txt
```diff
@@ -0,0 +1,109 @@
...
+    Document{{_id=6, title=A Dance with Dragons, length=1104, author=Martin}}
+    Document{{_id=3, title=Atlas Shrugged, length=1088, author=Rand}}
+    Document{{_id=5, title=Cryptonomicon, length=918, author=Stephenson}}
+    Document{{_id=4, title=Infinite Jest, length=1104, author=Wallace}}
+    Document{{_id=2, title=Les Misérables, length=1462, author=Hugo}}
+    Document{{_id=1, title=The Brothers Karamazov, length=824, author=Dostoyevsky}}
+
+Sometimes, the order of two or more documents is ambiguous using a
+specified sort. Consider the case of sorting by length: in the above case, 
+both "A Dance with Dragons" and "Infinite Jest" have ``1104`` pages, so
```
jyemin, line 74: Nice examples!

source/fundamentals/crud/sort.txt
```diff
@@ -0,0 +1,109 @@
...
+that the ``length`` field comparison results in a tie, the operation compares the tied results 
+using the ``title`` field:
+
+
+.. code-block:: java
+  :emphasize-lines: 3
+    
+    // sort first by length, then title
+    MongoCursor<Document> cursor = collection.find()
+        .sort(ascending("length", "title")).iterator();
```
jyemin, line 89: You might want to show an example where one field is ascending and another is descending.

jyemin (changes requested):

source/fundamentals/crud/sort.txt
```diff
@@ -101,6 +101,31 @@ When there are two documents with the same ``length`` value, it resolves the tie
...
+You can also sort by multiple fields in different orders, using ``Sorts.orderBy()``. 
+Consider this example where you first sort on ``length`` in an ascending order, and then on
+``title`` in a descending order:
+
+.. code-block:: java
+  :emphasize-lines: 3
+    
+    // sort first by length ascending, then title descending
+    MongoCursor<Document> cursor = collection.find()
+        .sort(orderBy(ascending("length"), descending("title")))
```
jyemin, line 113: Should we have imports for this example, to be consistent?

ksadoff: My thoughts for that example were to only show the code that has changed from the previous examples (so, just the sort call), but I can add the import if you think it would clarify things. I can also add a sentence in explaining that any omitted code stays constant from one example to the next.

jyemin: Right.  But the previous example statically imports just `import static com.mongodb.client.model.Sorts.ascending` and not any other static method from `Sorts`.  One option is to change the import to a wildcard, like `import static com.mongodb.client.model.Sorts.*`.   

@ccho-mongodb do we do that elsewhere for builders?

jyemin (changes requested):

source/fundamentals/crud/sort.txt
ccho-mongodb: We haven't documented builders yet, but I used explicit static method imports for a snippet in this data formats example:

https://github.com/mongodb/docs-java/blob/master/source/fundamentals/data-formats/document-data-format-pojo.txt#L169-L172

I prefer listing them individually over using "*" since it's more clear what package the method comes from, but am flexible.

jyemin (approved):

PR #47, opened 2020-12-04
DOCSP-9558: Fundamentals > CRUD Operations > Read Operations > Limit the Number of Returned Results

ksadoff:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-9558

https://docs-mongodborg-staging.corp.mongodb.com/d2fdb0a/java/docsworker-xlarge/DOCSP-9558/fundamentals/crud/read/limit

### Self-Review Checklist

### DIFF
--- a/source/fundamentals/crud/read/limit.txt
+++ b/source/fundamentals/crud/read/limit.txt
@@ -110,6 +110,29 @@ longest books:
 You can combine ``skip()`` and ``limit()`` in this way to implement paging for your
 collection, returning only small subsets of the collection at one time.
 
+.. note::
+
+  In order to ensure stable sorts across multiple queries, you must sort
+  using a unique key (such as ``_id``). Otherwise, a call to ``skip()``
+  and ``limit()`` may produce unpredictable results when combined with
+  ``sort()``.
+
+  For example, consider the following data:
+
+  .. code_block:: java
+      
+      { type: "computer", data: "1", serial_no: 235235 }
+      { type: "computer", data: "2", serial_no: 235237 }
+      { type: "computer", data: "3", serial_no: 235239 }
+      { type: "computer", data: "4", serial_no: 235241 }
+
+  If you sorted by ``type`` alone, ``sort()`` does not guarantee the same order
+  upon return. Appending ``skip()`` and ``limit()`` to the ``sort()``
+  could return different documents for different queries. In this case, sorting
+  by ``data`` or ``serial_no`` would guarantee a stable sort, as both are unique keys.
+
+
+
 For additional information on the classes and methods mentioned on this
 page, see the following API documentation:
 
--- a/source/fundamentals/crud/read/limit.txt
+++ b/source/fundamentals/crud/read/limit.txt
@@ -14,7 +14,7 @@ the skip.
 
 The examples below demonstrate, respectively, how to insert data into
 a collection, how to use ``limit()`` to restrict the number of returned documents,
-and how to combine ``limit()`` with ``skip()`` to further narrow the results returned from that query. 
+and how to combine ``limit()`` with ``skip()`` to further narrow the results returned from a query. 
 
 The following operation inserts documents representing books into a collection:
 
--- a/source/fundamentals/crud/read/limit.txt
+++ b/source/fundamentals/crud/read/limit.txt
@@ -89,23 +89,13 @@ To see the next three longest books, append the ``skip()`` method to your
 how many documents the find operation returns:
 
 .. code-block:: java
-  :emphasize-lines: 5, 6
-
-    import com.mongodb.client.model.Sorts;
+  :emphasize-lines: 3, 4
 
     MongoCursor<Document> cursor = collection.find()
         .sort(ascending("length"))
         .limit(3)
         .skip(3)
         .iterator();
-    try {
-        while(cursor.hasNext()) {
-            System.out.println(cursor.next());
-        }
-    } 
-    finally {
-        cursor.close();
-    }
 
 This operation returns the documents that describe the fourth through sixth
 longest books:
--- a/source/fundamentals/crud/read/limit.txt
+++ b/source/fundamentals/crud/read/limit.txt
@@ -43,7 +43,7 @@ a ``sort`` on the ``length`` field to return books with longer lengths before
 books with shorter lengths, and applies a ``limit`` to return only ``3`` results:
 
 .. code-block:: java
-  :emphasize-lines: 5
+  :emphasize-lines: 6
 
     import com.mongodb.client.model.Sorts;
 
@@ -89,7 +89,6 @@ the number of documents to skip over to the previous code snippet's call to
 ``find()``:
 
 .. code-block:: java
-
   :emphasize-lines: 5
 
     import com.mongodb.client.model.Sorts;
--- a/source/fundamentals/crud/read/limit.txt
+++ b/source/fundamentals/crud/read/limit.txt
@@ -5,9 +5,9 @@ Limit the Number of Returned Results
 .. default-domain:: mongodb
 
 Use ``limit`` to cap the number of documents that can be returned from a
-read operation. ``limit`` functions as a cap on the maximum number of
-documents that the operation can return, but the operation can return
-a smaller number of documents if there are not enough documents present
+read operation. The ``limit`` operation functions as a cap on the maximum number of
+documents that the operation can return, but it can return
+a smaller number of documents if there are not enough present
 to reach the limit. If ``limit`` is used with the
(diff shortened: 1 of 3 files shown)


source/fundamentals/crud/read/limit.txt
```diff
@@ -0,0 +1,117 @@
...
+   { "_id": 1, "title": "The Brothers Karamazov", "author": "Dostoyevsky", "length": 824 }
+
+.. note::
+
+   These results are in JSON format due to the ``toJson()`` method used in the example above.
+   Without ``toJson()``, they are returned in a Java string format, 
+   like this: ``Document{{_id=6, title=A Dance with Dragons, author=Martin, length=1104}}``.
+
+You can combine ``skip`` and ``limit`` in this way to implement paging for your
+collection, returning only small "slices" of the collection at once.
```
nathan-contino-mongo, line 117: Should we add a link to this page in a toctree somewhere, or are we holding off on that until we create the CRUD pages that will sit above this page in the sidebar hierarchy?

source/fundamentals/crud/read/limit.txt
```diff
@@ -0,0 +1,117 @@
...
+The following example queries the collection to return the top three
+longest books. It matches all the documents with the query, applies
+a ``sort`` on the ``length`` field to return books with longer lengths before
+books with shorter lengths, and applies a ``limit`` to return only ``3`` results:
+
+.. code-block:: java
+
+    // define a cursor that will return the first 3 sorted items
+    MongoCursor<Document> cursor = collection.find()
+        .sort(descending("length"))
```
nathan-contino-mongo, line 49: Vague method names like this can sometimes cause issues when users try to copy/paste the code, but can't quite get the imports right (particularly since method names like this tend to have overlaps with standard library methods). Consider following the pattern that @ccho-mongodb has used on other docs-java pages to demonstrate imports at the top of an example: https://github.com/mongodb/docs-java/blob/master/source/fundamentals/data-formats/documents.txt#L134

source/fundamentals/crud/read/limit.txt
```diff
@@ -0,0 +1,117 @@
...
+      while (cursor.hasNext()) {
+        System.out.println(cursor.next().toJson());
+      }
+    // close the cursor
+    } finally {
+      cursor.close();
+    }
+
+
+The code example above outputs the following three documents, sorted by
```
nathan-contino-mongo, line 63: nitpick: "prints out" might be a more descriptive phrase than "outputs"

source/fundamentals/crud/read/limit.txt
```diff
@@ -0,0 +1,117 @@
...
+            .append("title", "A Dance with Dragons").append("length", 1104)
+            .append("author", "Martin")
+    ));
+
+The following example queries the collection to return the top three
+longest books. It matches all the documents with the query, applies
+a ``sort`` on the ``length`` field to return books with longer lengths before
+books with shorter lengths, and applies a ``limit`` to return only ``3`` results:
+
+.. code-block:: java
```
nathan-contino-mongo, line 45: It looks like the Node.js equivalent of this page uses emphasize-lines to highlight the line containing the sort call. I think the code samples is just long enough that this highlight is helpful, but it's a bit of a subjective call. What do you think?

source/fundamentals/crud/read/limit.txt
```diff
@@ -0,0 +1,117 @@
...
+
+.. code-block:: javascript
+
+   { "_id": 3, "title": "Atlas Shrugged", "author": "Rand", "length": 1088 }
+   { "_id": 5, "title": "Cryptonomicon", "author": "Stephenson", "length": 918 }
+   { "_id": 1, "title": "The Brothers Karamazov", "author": "Dostoyevsky", "length": 824 }
+
+.. note::
+
+   These results are in JSON format due to the ``toJson()`` method used in the example above.
```
nathan-contino-mongo, line 112: Couple of things here.
1. Passive voice, suggest reword to active
2. "returned in a Java string format" isn't strictly true here, because we're actually printing the documents and they're turned into a string either way -- toJson() just turns the objects into a different format first, so they use a different toString().
3. Is there a reason why we use toJson() here anyway? I suppose it is a bit more readable than the default toString method, but I wonder if we're better off displaying the results that way instead if we need to include a note anyway to explain what's going on with toJson().

I'd suggest a rewording to something like this, but feel free to reword however you see fit -- this is just a suggestion I came up with on the fly:
```
These results appear in JSON format in this example because
we use the ``toJson()`` method to transform the results into
a JSON format that prints as a JSON string. If you omit the
``toJson()`` method call, the results instead print using the
``toString()`` method of the
:doc:`data format </fundamentals/data-formats>` used to
represent objects in your ``MongoCollection``, which
defaults to ``Document``.
```

nathan-contino-mongo (approved):
LGTM with some small nitpicks and suggestions.

source/fundamentals/crud/read/limit.txt
ksadoff: I can add it, I agree that it might be helpful.

source/fundamentals/crud/read/limit.txt
ksadoff: To your 3rd point, I used it because of readability, but it seems like it might simplify things a bit to simply omit ``toJson()`` and this note altogether, and display the documents as they're originally printed.

nathan-contino-mongo (commented):
This looks good, though if we're going to phase out ``toJson()`` calls, I think it's a good idea to also remove the calls from the Sort page.

ksadoff:
> This looks good, though if we're going to phase out `toJson()` calls, I think it's a good idea to also remove the calls from the Sort page.

I had the same thought! I'll make sure to go through that PR and remove those calls (as well as add the relevant ``toctree`` lines in there)

nathan-contino-mongo:
For the sake of merge conflict sanity, it might be a good idea to just include the toctree changes in this PR, then merge the sort PR, then merge master into this PR -- that way, you don't have to fix a merge conflict including the exact same changes in both branches. But honestly just do whatever you feel comfy with.

ccho-mongodb (commented):
Let's use 4-space indentation for all Java code examples which is the common formatting convention.

Make sure the antecedent is clear whether in the same sentence or in a larger scope:
E.g. "The limit operation functions as a cap..." vs "The limit operation caps..."
Another e.g. "passing the number of documents to skip over... to find()"
Larger scope example: "Follow the instructions..." but after the insert, the imperative language (do/run this) is omitted.

Try to limit number of details in each sentence (part of our style guide https://docs.mongodb.com/meta/style-guide/writing/write-clear-concise-sentences-paragraphs/).

Similar to the other PR, make sure to consistently use `()` after method names. I see that you've used that syntax (e.g. line 87 and 89), but not in all instances (such as line 42 and 43).

Avoid introducing terms without immediately explaining them. E.g. line 120 "slices".

ccho-mongodb (changes requested):
Looks pretty good, just added a couple suggestions and comments.

ccho-mongodb (approved):
LGTM!

ksadoff:
Added @jyemin and @rozza as reviewers, but only one review is necessary

source/fundamentals/crud/read/limit.txt
```diff
@@ -0,0 +1,119 @@
...
+   because the driver reorders the calls to apply the sort first and the
+   limit after it. The following two calls are equivalent:
+
+   .. code-block:: java
+
+      collection.find().sort(descending("length")).limit(3);
+      collection.find().limit(3).sort(descending("length"));
+
+
+To see the next three longest books, append the ``skip()`` method to your
```
jyemin, line 88: Given the recent activity on DOCS-9994  and DOCSP-13487 , I'm not sure we should be recommending this sort of usage of skip/limit.  Do we do this in other places?

source/fundamentals/crud/read/limit.txt
```diff
@@ -110,6 +110,29 @@ longest books:
 You can combine ``skip()`` and ``limit()`` in this way to implement paging for your
 collection, returning only small subsets of the collection at one time.
 
+.. note::
```
jyemin, line 113: @ccho-mongodb this is probably something you should review, since it's the sort of thing we should standardize on across all of our documentation.

ccho-mongodb: Sure, talked about this with Katie earlier today. Thanks for bringing it up, I didn't know about this before then. I created a ticket for DevEd to update Node.js documentation and will need to discuss with the driver engineers for the rest of the documentation. I'll have to look into how Server Docs team wants to provide messaging for this as well.

jyemin (approved):
LGTM

PR #49, opened 2020-12-17
Docsp 9556

terakilobyte:
## Pull Request Info

This adds the landing "drawer" for crud operations. I created some placeholder pages to make it look better.

Additionally, there was a build warning that I fixed in [`limit.txt`](https://github.com/mongodb/docs-java/blob/3b41410d27fae8ebc53e0e639b220838511c6ca8/source/fundamentals/crud/read/limit.txt#L122).

### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-9556

https://docs-mongodbcom-staging.corp.mongodb.com/a61cf71/java/docsworker-xlarge/DOCSP-9556/fundamentals/crud

### Self-Review Checklist


source/fundamentals/crud.txt
```diff
@@ -1,13 +1,26 @@
...
 .. toctree::
-   :titlesonly:
-   :maxdepth: 1
+   :caption: CRUD Operations
+
+   /fundamentals/crud/read-operations
+   /fundamentals/crud/write-operations
+
+CRUD (Create, Read, Update, Delete) operations allow you to work with
+the data stored in MongoDB.
```
ccho-mongodb, line 14: I wonder if it might be useful to mention what they impact rather than what they "allow you to [do]".
 E.g.:

You can specify CRUD (Create, Read, Update, Delete) operations to read and modify data in your MongoDB instance.

or (might be too passive):

CRUD (...) operations enable you to read and modify data in MongoDB.

source/fundamentals/crud.txt
```diff
@@ -1,13 +1,26 @@
...
+
+The CRUD operation documentation is categorized in two sections:
```
ccho-mongodb, line 16: I think the preposition for categorized should be into if it's acting on the subject (CRUD operation documentation). If the subject was "two sections", you could "categorize [each] in two sections", but I don't think that is the correct description.

```suggestion
The CRUD operation documentation is categorized into two sections:
```

source/fundamentals/crud.txt
```diff
@@ -1,13 +1,26 @@
...
+
+- :doc:`Read Operations </fundamentals/crud/read-operations>` find and return
+  documents stored within your MongoDB database.
```
ccho-mongodb, line 19: Nitpick:
Not certain, but I think we usually talk about data stored "in" a database rather than within. In either case, it would be good to stay consistent (see line 21).

source/fundamentals/crud.txt
```diff
@@ -1,13 +1,26 @@
...
+- :doc:`Write Operations </fundamentals/crud/write-operations>` insert, modify,
+  or delete documents in your MongoDB database.
 
-   /fundamentals/crud/read
-   
\ No newline at end of file
+..
+  Some operations combine aspects of read and write operations. See our
+  guide on :doc:`compound operations </fundamentals/crud/compound-operations>`
+  to learn more about these hybrid methods.
```
ccho-mongodb, line 26: Question: is hybrid the term we use for compound methods? It seems like terminology to me, so perhaps we should avoid it if it's not the official term and we're not explaining what makes them hybrid.

ccho-mongodb (approved):
LGTM, a couple nitpicks

source/fundamentals/crud.txt
terakilobyte: I'm not aware of any approved term. This is commented out and will only be included in once we have a page or section for compound operations, and I agree that at that time we should look at making this more approachable with as few "jargon" words as possible.

source/fundamentals/crud.txt
terakilobyte: I've reworded this and removed some wording that I felt was adding nothing to the page.

ccho-mongodb (approved):

PR #52, opened 2021-01-22
DOCSP-9550 POJO customization

ccho-mongodb:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-9550

https://docs-mongodbcom-staging.corp.mongodb.com/e69ddf4/java/docsworker-xlarge/DOCSP-9550-pojo-customization/fundamentals/data-formats/pojo-customization

### Assumptions
- The :java-docs: directive points to 4.2 rather than 4.1

### Self-Review Checklist

### DIFF
--- a/source/fundamentals/data-formats/pojo-customization.txt
+++ b/source/fundamentals/data-formats/pojo-customization.txt
@@ -26,7 +26,7 @@ about more specific customization from the section on :ref:`Advanced Configurati
 We also show you how to use helpers such as :ref:`Conventions <conventions>`
 and :ref:`Annotations <annotations>` to specify common serialization actions.
 
-See the section on :ref:`Discriminators <discriminators>` if you want to
+See the section on :ref:`Discriminators <pojo-discriminators>` if you want to
 serialize multiple POJO classes to documents in the same collection.
 
 If you need to implement conditional serialization, or if you need to use
@@ -84,7 +84,7 @@ ClassModel
 A ``ClassModel`` instance stores data conversion information about a specific
 POJO class. It contains a list of ``PropertyModel`` instances which describe
 the property fields of the POJO, whether to convert fields, and optionally,
-``Codecs`` to convert them with.
+``Codecs`` to convert the fields.
 
 A ``ClassModel`` contains the following fields:
 
--- a/source/fundamentals/data-formats/pojo-customization.txt
+++ b/source/fundamentals/data-formats/pojo-customization.txt
@@ -23,7 +23,14 @@ We show how to specify your data conversion using the :ref:`ClassModel <classmod
 and :ref:`PropertyModel <property-model>` classes. You can also learn
 about more specific customization from the section on :ref:`Advanced Configuration <pojo-advanced-configuration>`
 
-If you only need to use the pre-defined behavior to convert data between
+We also show you how to use helpers such as :ref:`Conventions <conventions>`
+and :ref:`Annotations <annotations>` to specify common serialization actions.
+
+If you need to implement conditional serialization, or if you need to use
+enums, generics, interface, or abstract types, see the section on
+:ref:`Advanced Configuration <pojo-advanced-configuration>`.
+
+If you only need to use the predefined behavior to convert data between
 BSON and POJOs, you can use the **automatic** setting for the
 ``PojoCodecProvider`` shown in the :doc:`Document Data Formats: POJOs guide </fundamentals/data-formats/document-data-format-pojo>`.
 
@@ -40,10 +47,10 @@ You can create a ``PojoCodecProvider`` instance using the
 ``PojoCodecProvider.builder()`` method. You can also chain methods to the
 builder to register any of the following:
 
-- Individual POJO classes
-- Package names that contain POJO classes
-- ``ClassModel`` instances that describe conversion logic for a specific POJO
-  class
+- individual POJO classes
+- package names that contain POJO classes
+- instances of ``ClassModel`` that describe conversion logic for a specific
+  POJO class
 
 The following example shows how you can specify the POJOs in a package named
 "org.example.pojos" and add the ``PojoCodecProvider`` to a ``CodecRegistry``:
@@ -65,9 +72,6 @@ The following example shows how you can specify the POJOs in a package named
 
 See the :java-docs:`PojoCodecProvider.Builder API documentation </apidocs/bson/org/bson/codecs/pojo/PojoCodecProvider.Builder.html>`
 for more information.
-
-
-TODO: (transition) the following sections...
 
 .. _classmodel:
 
@@ -199,11 +203,10 @@ from the following static fields in the ``Conventions`` class:
        package for your POJO.
 
    * - ``CLASS_AND_PROPERTY_CONVENTION``
-     - Sets the following default values for the ``ClassModel`` and
-       ``PropertyModel`` instances:
-       - Discriminator key to ``_t``
-       - Discriminator value to the ``ClassModel`` simple type name
-       - Id field to ``_id`` for each ``PropertyModel``.
+     - | Sets the following default values for the ``ClassModel`` and ``PropertyModel`` instances:
+       | - Discriminator key to ``_t``
+       | - Discriminator value to the ``ClassModel`` simple type name
+       | - Id field to ``_id`` for each ``PropertyModel``.
 
    * - ``DEFAULT_CONVENTIONS``
      - | Enables the following Conventions:
@@ -257,22 +260,23 @@ package:
 
    * - ``BsonCreator``
      - Marks a public constructor or a public static method as the creator for
-     new instances of the class. Can combine with the BsonProperty annotation to link the parameters with properties.
+       new instances of the class. You can combine this with the
+       ``BsonProperty`` annotation to link the parameters with properties.
 
    * - ``BsonDiscriminator``
-     - Specifies a class uses a discriminator. It also allows you to set a
-     custom discriminator key and value.
+     - Specifies a class uses a discriminator. You can set a custom
+       discriminator key and value.
 
    * - ``BsonId``
      - Marks a property to serialized as the _id property.
 
    * - ``BsonIgnore``
-     - Marks a property to ignore. Can configure whether to serialized and/or
-       deserialize a property.
+     - Marks a property to ignore. You can configure whether to serialized
+       and/or deserialize a property.
 
    * - ``BsonProperty``
      - Allows for an alternative document key name when converting the POJO
-       field to BSON. Also, allows a field to turn on using a discriminator
+       field to BSON. Also allows a field to turn on using a discriminator
        when storing a POJO value.
 
 The following code snippet shows a sample POJO called ``Product`` that uses
@@ -329,11 +333,10 @@ The annotations in the example POJO specify the following behavior:
   on write operations
 - convert between the POJO ``name`` field and value and the BSON ``modelName``
   field and value in the document
-- convert between the POJO ``serialNumber`` field and value the BSON docuemnt
+- convert between the POJO ``serialNumber`` field and value the BSON document
   ``_id`` field and value in the document
 - omit the ``relatedItems`` field and value when converting data
 - use the ``Product(String name)`` constructor when instantiating the POJO
-
 
 .. _pojo-discriminators:
 
@@ -355,8 +358,8 @@ following:
 
 - use the ``@BsonDiscriminator`` annotation to specify the discriminator for
   the POJO class
-- or call ``enableDiscriminator(true)`` on the ``ClassModelBuilder`` for the
-  POJO class
+- or call ``enableDiscriminator(true)`` on the ``ClassModelBuilder``
+  associated with the POJO class
 
 See the following example POJO classes that contain ``@BsonDiscriminator``
 annotations and example documents that contain the discriminator fields:
@@ -507,7 +510,7 @@ resemble the following:
 .. code-block:: json
    :copyable: false
 
-   { "_id" : ObjectId("..."), "eventDateTime" : ..., "name" : "Robinette" }
+   { "_id" : ObjectId("..."), "eventDateTime" : ..., "name" : "Galadriel" }
 
 Generics Support
 ^^^^^^^^^^^^^^^^
@@ -519,8 +522,8 @@ properties if they meet the following criteria:
 - the top-level POJO does not contain any type parameters
 
 The ``ClassModelBuilder`` inspects and saves concrete type parameters to work
-around type erasure. Classes that contain generic properties without concrete
-type parameters cannot be serialized since the JVM removes the type
+around type erasure. It cannot serialize classes that contain generic
+properties without concrete type parameters since the JVM removes the type
 parameter information.
 
 Instead, you can implement the ``PropertyCodecProvider`` interface to specify
@@ -542,7 +545,7 @@ Suppose you wanted to serialize the following POJO:
 
 You can use the following implementation of ``PropertyCodecProvider`` to
 retrieve the your custom Codec. This implementation uses the
-``TypeWithParameters`` interface to access the type information.
+``TypeWithTypeParameters`` interface to access the type information.
 
 .. code-block:: java
    :caption: Implementation of PropertyCodecProvider
@@ -605,6 +608,10 @@ and the package that contains your POJO as follows:
            .register(new OptionalPropertyCodecProvider())
            .build();
 
+See the API documentation for more information on the following:
+
+- :java-docs:`PropertyCodecProvider <apidocs/bson/org/bson/codecs/pojo/PropertyCodecProvider.html>`
+- :java-docs:`TypeWithTypeParamaters <apidocs/bson/org/bson/codecs/pojo/TypeWithTypeParameters.html>`
 
 Enum Type Support
 ^^^^^^^^^^^^^^^^^
@@ -628,7 +635,7 @@ field:
        private Integer height;
        private Family family;
 
-       // ... the rest of the Class definitiion
+       // ...
   }
 
 The following snippets show an example instantiation of the ``Penguin``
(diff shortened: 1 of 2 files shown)


source/fundamentals/data-formats/pojo-customization.txt
```diff
@@ -10,78 +10,653 @@ POJO Customization
...
+   * - InstanceCreatorFactory
+     - Contains a new instance factory that creates new instances of the POJO.
+       By default, it requires the POJO to have an empty constructor.
+
+   * - PropertyModels
+     - Contains a list of ``PropertyModel`` instances that specify how to
+       convert data to and from BSON for a field in the POJO.
+
+   * - IdPropertyModelHolder
+     - Specifies the POJO field that corresponds to the document id field. Optional.
```
nathan-contino-mongo, line 107:
```suggestion
     - Specifies the POJO field that corresponds to the document ``_id`` field. Optional.
```
I assume this is talking about the MongoDB default _id field. Would this be more correct, or does the Java driver use a different convention?

source/fundamentals/data-formats/pojo-customization.txt
```diff
@@ -10,78 +10,653 @@ POJO Customization
...
+
+   * - Discriminator Key
+     - Specifies the name of the discriminator field. Optional.
```
nathan-contino-mongo, line 110: Perhaps this is just a blind spot of mine, but what is a discriminator field? Should we link to something that explains this?

source/fundamentals/data-formats/pojo-customization.txt
```diff
@@ -10,78 +10,653 @@ POJO Customization
...
+       specified in the checker.
 
-- Description
-- Table of properties
-- Link to Conventions section for example of PropertyModelBuilder
+   * - Property accessor
+     - Method used to access the value of the property from the POJO
+
+   * - useDiscriminator
+     - Specifies whether to use the discriminator.
```
nathan-contino-mongo, line 171: Same discriminator comment as above.

source/fundamentals/data-formats/pojo-customization.txt
```diff
@@ -10,78 +10,653 @@ POJO Customization
...
+
+   * - Serialization checker
+     - Determines whether to serialize a value by using the criteria
+       specified in the checker.
 
-- Description
-- Table of properties
-- Link to Conventions section for example of PropertyModelBuilder
+   * - Property accessor
+     - Method used to access the value of the property from the POJO
```
nathan-contino-mongo, line 168:
```suggestion
     - Method used to access the value of the property from the POJO.
```

source/fundamentals/data-formats/pojo-customization.txt
```diff
@@ -10,78 +10,653 @@ POJO Customization
...
+
+   * - useDiscriminator
+     - Specifies whether to use the discriminator.
+
+To create a ``PropertyModel`` use a ``PropertyModelBuilder`` which you
+can instantiate by calling the ``PropertyModel.builder()`` method.
+
+For more information, see the
+:java-docs:`PropertyModel.Builder API documentation <apidocs/bson/org/bson/codecs/pojo/PropertyModelBuilder.html>`.
```
nathan-contino-mongo, line 177:
```suggestion
:java-docs:`PropertyModel.Builder <apidocs/bson/org/bson/codecs/pojo/PropertyModelBuilder.html>` API documentation.
```
The other links to API docs don't include "API documentation" in the link itself.

source/fundamentals/data-formats/pojo-customization.txt
```diff
@@ -10,78 +10,653 @@ POJO Customization
...
+.. list-table::
+   :header-rows: 1
+   :stub-columns: 1
+
+   * - Field Name
+     - Description
+
+   * - ``ANNOTATION_CONVENTION``
+     - Enables the annotations defined in the ``org.bson.codecs.pojo.annotations``
+       package for your POJO.
```
nathan-contino-mongo, line 203: I know that the Driver docs tend to leave links to the end of a section, but I wonder if it would be useful to link to the list of those annotations here (since this description isn't hugely helpful without knowing the annotations you can use!). It seems like a link to the next section would be adequate? Up to you.

source/fundamentals/data-formats/pojo-customization.txt
```diff
@@ -10,78 +10,653 @@ POJO Customization
...
 Annotations
 -----------
 
-- Description
-- Table of Annotations
-- Example class with annotations
-- Example output
+You can apply annotations to the getter and setter methods of a POJO
+class. These annotations configure the ``ClassModel`` and ``PropertyModel``
+behavior for a specific field, method, or class.
```
nathan-contino-mongo, line 248: It might be helpful to mention how conflicts between annotations and Conventions are resolved. I assume that annotations override conventions if annotations are enabled, but it might be nice to have an explicit callout of that fact (or, if there are exceptions to that behavior, to warn users).

source/fundamentals/data-formats/pojo-customization.txt
```diff
@@ -10,78 +10,653 @@ POJO Customization
...
+
+       @BsonId()
+       @BsonRepresentation(BsonType.OBJECT_ID)
+       private String serialNumber;
+
+       @BsonIgnore
+       private List<Product> relatedItems;
+
+       @BsonCreator
+       public ProductPojo(@BsonProperty("modelName") String name) {
```
nathan-contino-mongo, line 314: Big fan of this example. One thing that I notice you don't mention is the usage of this ``@BsonProperty`` annotation in the constructor parameters -- should we add this to the BsonProperty bullet point?

source/fundamentals/data-formats/pojo-customization.txt
```diff
@@ -10,78 +10,653 @@ POJO Customization
...
+- convert between the POJO ``name`` field and value and the BSON ``modelName``
+  field and value in the document
+- convert between the POJO ``serialNumber`` field and value the BSON document
+  ``_id`` field and value in the document
+- omit the ``relatedItems`` field and value when converting data
+- use the ``Product(String name)`` constructor when instantiating the POJO
+
+.. _pojo-discriminators:
+
+Discriminators
```
nathan-contino-mongo, line 349: Discriminators are mentioned an awful lot in the preceding sections. I would consider either moving this section above or explicitly linking to this section when you mention discriminators, otherwise reading this page in a linear format can make you think that you're missing prerequisite knowledge about discriminators. Or maybe a link/mention in the overview would be enough?

source/fundamentals/data-formats/pojo-customization.txt
```diff
@@ -10,78 +10,653 @@ POJO Customization
...
+
+  @BsonDiscriminator(value="AnonymousUser", key="_cls")
+  public class AnonymousUser {
+      // class code
+  }
+
+  @BsonDiscriminator(value="RegisteredUser", key="_cls")
+  public class RegisteredUser {
+      // class code
+  }
```
nathan-contino-mongo, line 384: For some reason, I don't see syntax highlighting for this code sample. Not a huge problem, but a big jarring since the other examples all have syntax highlighting, which is particularly obvious for annotations.

source/fundamentals/data-formats/pojo-customization.txt
```diff
@@ -10,78 +10,653 @@ POJO Customization
...
+   CodecRegistry pojoCodecRegistry = fromRegistries(getDefaultCodecRegistry(), fromProviders(pojoCodecProvider));
+
+If you insert a POJO that contains a value greater than 29 in the ``age``
+field, the serialized document omits it. The resulting document should
+resemble the following:
+
+.. code-block:: json
+   :copyable: false
+
+   { "_id" : ObjectId("..."), "eventDateTime" : ..., "name" : "Galadriel" }
```
nathan-contino-mongo, line 521: This example is both education and hilarious. I would suggest introducing a sample full Galadriel document (including an age) before we show this resulting document, though, in case users want to confirm their understanding of the example.

source/fundamentals/data-formats/pojo-customization.txt
```diff
@@ -10,78 +10,653 @@ POJO Customization
...
 
 Generics Support
 ^^^^^^^^^^^^^^^^
 
-- Description of when a POJO is not serializable
-- Example showing how to use PropertyCodecProviders to provide concrete class information
+You can use the POJO ``Codec`` to serialize classes that contain generic
+properties if they meet the following criteria:
+
+- contain only bounded concrete type parameters
```
nathan-contino-mongo, line 529: I would suggest adding a subject like "the top-level POJO and all subclasses must" for consistency with the next bullet point

nathan-contino-mongo: Additionally, a link to what "concrete type parameters" (or an explanation below) would help this be more understandable to readers.

source/fundamentals/data-formats/pojo-customization.txt
```diff
@@ -10,78 +10,653 @@ POJO Customization
...
+- the top-level POJO does not contain any type parameters
+
+The ``ClassModelBuilder`` inspects and saves concrete type parameters to work
+around type erasure. It cannot serialize classes that contain generic
+properties without concrete type parameters since the JVM removes the type
+parameter information.
+
+Instead, you can implement the ``PropertyCodecProvider`` interface to specify
```
nathan-contino-mongo, line 537: Instead of using the default ClassModelBuilder? I would consider ditching the "Instead" here or maybe elaborate more about the options before we descend into the details in these two paragraphs.

source/fundamentals/data-formats/pojo-customization.txt
```diff
@@ -10,78 +10,653 @@ POJO Customization
...
+
+       // ...
+   }
+
+You can use the following implementation of ``PropertyCodecProvider`` to
+retrieve the your custom Codec. This implementation uses the
+``TypeWithTypeParameters`` interface to access the type information.
+
+.. code-block:: java
+   :emphasize-lines: 3
```
nathan-contino-mongo, line 558: Do you think this emphasize-lines is useful? Consider either removing it or highlighting the entire ``get()`` method and its annotations.

nathan-contino-mongo (approved):
LGTM. Just a few musings, nitpicks, and suggestions to help improve the page, but I think it's shaping up to be brilliant already.

source/fundamentals/data-formats/pojo-customization.txt
ccho-mongodb: I thought it might be useful to highlight the overridden method declaration but not all the custom logic it contains, but maybe it's jarring -- will update.

source/fundamentals/data-formats/pojo-customization.txt
ccho-mongodb: Will update the bullet point and I'll add a link to the Oracle docs for info on type parameters at the end of the section.

source/fundamentals/data-formats/pojo-customization.txt
ccho-mongodb: I'll add emphasize-lines if syntax highlighting isn't working. I think the implementation of syntax highlighting for Java is poor because it doesn't pick up anything unless wrapped in "public class" except some specific keywords. I'll probably open a DOP ticket if I can find something better.

source/fundamentals/data-formats/pojo-customization.txt
ccho-mongodb: Since it can't be used by itself in this context, I described under the BsonCreator item https://github.com/mongodb/docs-java/pull/52/files#diff-81770bd4e69d34b6329eacade6c66cffd45edab4a1eb84310dc644364e1660deR261-R264

I think it might be a little confusing or extraneous to talk about how you need to use it only if you decide to use a custom constructor which requires @BsonCreator, and @BsonCreator already mentions it.

source/fundamentals/data-formats/pojo-customization.txt
```diff
@@ -10,78 +10,653 @@ POJO Customization
...
+
+.. list-table::
+   :header-rows: 1
+   :stub-columns: 1
+
+   * - Field Name
+     - Description
+
+   * - Name
+     - The POJO class name to associate with the ``ClassModel``.
```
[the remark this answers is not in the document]
ccho-mongodb: While I thought about it, I think it would be too detailed and cause information overload as it's intended only to give the reader an overview of the structure to see how it fits together in the POJO serialization ecosystem. I'll consult external reviewers to see if we can link to the test cases containing them.

source/fundamentals/data-formats/pojo-customization.txt
ccho-mongodb: I'll link to the section in the guide

source/fundamentals/data-formats/pojo-customization.txt
ccho-mongodb: Thanks, I'll link to the next section.

source/fundamentals/data-formats/pojo-customization.txt
ccho-mongodb: Thanks, I'll link to the next section. To reiterate why API documentation goes to the end of the section is because we received both writer and reader feedback that opening links external to the doc distract from the material.

source/fundamentals/data-formats/pojo-customization.txt
ccho-mongodb: Not sure what you mean, but maybe the sentence that explains that conventions are applied in order https://github.com/mongodb/docs-java/pull/52/files#diff-81770bd4e69d34b6329eacade6c66cffd45edab4a1eb84310dc644364e1660deR188-R189 needs to be more prominient. If the ANNOTATION_CONVENTION convention is not applied, the annotations do not get applied. If the convention is applied, it overrides whatever was specified earlier.

source/fundamentals/data-formats/pojo-customization.txt
ccho-mongodb: Will add a sentence and link in the overview, thanks for the suggestion.

source/fundamentals/data-formats/pojo-customization.txt
nathan-contino-mongo: Perhaps I wasn't clear here -- i meant situations where ANNOTATION_CONVENTION is enabled, and a specific annotation instance specifies something that conflicts with another convention (for instance, a situation where a user specifies a custom discriminator field/value but has the default discriminator convention enabled as well. I assume that the annotation for a custom discriminator field/value would "win" in that case, but it could be helpful to specify that here in a sentence.

ccho-mongodb: Hm, I think there might be some confusion and I want to clarify to make sure I'm not explaining something incorrectly before moving to external review.

First, could you clarify whether you're talking about:
1) "default values" - something specified only if nothing previously exists
2)  "DEFAULT_CONVENTIONS" - enables three pre-defined Conventions
3) Nothing to do with defaults - different behavior from two or more different Conventions in which case the latest one in the list overrides any conflicting value

I think i was explaining #3 in my prior comment in this thread.

For #1, suppose we are talking about CLASS_AND_PROPERTY_CONVENTION which provides those (description says "Sets the following default values..."). These defaults only apply if no other value is specified at that time. And, if any discriminator key/value is added to the class via PojoCodecProvider.Builder.conventions() or ClassModelBuilder.conventions() after the CLASS_AND_PROPERTY_CONVENTION is specified, it overwrites that value. If I first defined CLASS_AND_PROPERTY_CONVENTION, my "Product" class would have a "default" discriminator key "_t" and discriminator "Product". If my annotation specified @BsonDiscriminator(key="_cls", value="AnnotatedProduct"), that takes precedence whether CLASS_AND_PROPERTY_CONVENTION is defined before or after the ANNOTATION_CONVENTION.

For #2, I don't think any of the three conventions conflict. You can have situations such as #1 (in which an annotation overrides the defaults provided in CLASS_AND_PROPERTY_CONVENTION) or #3 (in which additional Conventions are specified to PojoCodecProvider.Builder.conventions() or ClassModelBuilder.conventions()).

Let me know if one of these answers your questions. Happy to chat over Zoom if that helps us get to the issue more quickly.

ccho-mongodb: Please ignore links generated by github markdown to "#1", "#2", and "#3"

nathan-contino-mongo: #1 answered my question -- specifically, "If my annotation specified @BsonDiscriminator(key="_cls", value="AnnotatedProduct"), that takes precedence whether CLASS_AND_PROPERTY_CONVENTION is defined before or after the ANNOTATION_CONVENTION." I was just suggesting that we could add a sentence explicitly calling out that relationship (between @BsonDiscriminator and CLASS_AND_PROPERTY_CONVENTION) but it might be that I missed a previous callout.

ccho-mongodb:
Hi @rozza and @jyemin, this PR is ready for your reviews.

One question from a teammate: do you think it would be helpful to the reader to link to test case classes in the repo that provide coverage of the different fields in the ClassModel and PropertyModel?

source/fundamentals/data-formats/pojo-customization.txt
```diff
@@ -10,78 +10,679 @@ POJO Customization
...
-- Summary of sections
+In this guide, you can learn how to define custom data conversions between
+BSON and POJOs. In our guide on :doc:`POJOs </fundamentals/data-formats/document-data-format-pojo>`,
+we show how to specify a ``PojoCodecProvider`` which contains classes that
+provide instructions on how to convert data for one or more POJO classes
+and their properties.
+
+We show how to specify your data conversion using the :ref:`ClassModel <classmodel>`
+and :ref:`PropertyModel <property-model>` classes. You can also learn
+about more specific customization from the section on :ref:`Advanced Configuration <pojo-advanced-configuration>`
```
rozza, line 24: Missing full stop

rozza (approved):
One nit: but after that LGTM.

jyemin:
I removed myself from the reviewers list on this one.

PR #53, opened 2021-01-25
Docsp 9569

terakilobyte:
## Pull Request Info

There is a link further down in the page that links to the Aggregation builders, a page yet to be created.

### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-9569

https://docs-mongodbcom-staging.corp.mongodb.com/d79eef7/java/docsworker-xlarge/DOCSP-9569/fundamentals/aggregation/

### Self-Review Checklist


source/fundamentals/aggregation.txt
```diff
@@ -0,0 +1,192 @@
...
+                 Aggregates.group("$stars", Accumulators.sum("count", 1))
+         )
+   ).forEach(printBlock);
+
+Use Aggregation Expressions
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The Java driver provides builders for :java-sync-api:`accumulator expressions<com/mongodb/client/model/Accumulators.html>`
+for use with :java-sync-api:`$group <com/mongodb/client/model/Aggregates.html#group(TExpression,java.util.List)>`. All
+other expressions must be manually built.
```
ccho-mongodb, line 167: Nitpick: I would prefer to be specific as to who performs what rather than mentioning the end-state. I think this helps avoid passive voice too.  In this sentence, we could reword "X must be [done]" as something like:

```suggestion
for use with :java-sync-api:`$group <com/mongodb/client/model/Aggregates.html#group(TExpression,java.util.List)>`. For all other expressions, you must manually declare the operators in JSON or compatible document format.
```

source/fundamentals/aggregation.txt
```diff
@@ -0,0 +1,192 @@
...
+  helper class.
+
+.. code-block:: java
+
+   collection.aggregate(
+         Arrays.asList(
+                 Aggregates.match(Filters.eq("categories", "Bakery")),
+                 Aggregates.group("$stars", Accumulators.sum("count", 1))
+         )
+   ).forEach(printBlock);
```
ccho-mongodb, line 160: I expected to see output after all of the examples that included printBlock. You might consider showing sample output or removing the forEach call to make that apparent.

source/fundamentals/aggregation.txt
```diff
@@ -0,0 +1,192 @@
...
+
+The Java driver provides builders for :java-sync-api:`accumulator expressions<com/mongodb/client/model/Accumulators.html>`
+for use with :java-sync-api:`$group <com/mongodb/client/model/Aggregates.html#group(TExpression,java.util.List)>`. All
+other expressions must be manually built.
+
+In the following example, the aggregation pipeline uses a
+`$project <https://mongodb.github.io/mongo-java-driver/4.2/apidocs/mongodb-driver-core/com/mongodb/client/model/Aggregates.html#project(org.bson.conversions.Bson)>`__
+stage and various
+`Projections <https://mongodb.github.io/mongo-java-driver/4.2/apidocs/mongodb-driver-core/com/mongodb/client/model/Projections.html>`__
+to return only the ``name`` field, and the calculated field ``firstCategory`` whose value is the first element in the
```
ccho-mongodb, line 173: Would it be worthwhile linking to the "arrayElemAt" aggregation operator (in an API links section at the end)? Also, I wonder if it would be helpful to explain the "$" field path for categories.

ccho-mongodb (changes requested):
Visual aid and examples look good though I didn't check the syntax closely.
I would have presented the subject a little differently. After explaining the animation, I would have assumed a dataset existed, shown an aggregation example using the builder and without it as well as sample output which I think is in-line with what we did for Node.js docs and how we have been proceeding with Java fundamentals sections. Let me know what you think.

source/fundamentals/aggregation.txt
```diff
@@ -0,0 +1,192 @@
...
+In a car factory, for example, one station might shape and bend the frame using a hydraulic press and a welder, another
+might affix the doors with a drill, and so on.
+
+Find and Aggregation Comparison
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Using ``find`` operations, you can:
+
+- select **what** documents to return
+- select **what** fields to return
```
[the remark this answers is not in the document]
terakilobyte: I had gone back and forth on this since italics can be missed, but they are probably more appropriate here. I want to leave the emphasis on the current words as is as that's idea.

source/fundamentals/aggregation.txt
```diff
@@ -0,0 +1,192 @@
...
+.. cssclass:: bordered-figure
+.. figure:: /figures/agg_factory.gif
+   :alt: Animation showing how the aggregation pipeline works, conceptually.
+
+Using the :ref:`Builders <builders>` API, the same operation may look something like:
+
+.. code-block:: java
+
+   Arrays.asList(
+       Aggregates.match(Filters.in("color", "red", "blue")), // only allow red and blue shapes to pass
```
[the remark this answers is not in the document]
terakilobyte: It is 4-space indentation, and shows up as such for me. I'll switch to a literalinclude to see if that helps

source/fundamentals/aggregation.txt
```diff
@@ -0,0 +1,237 @@
...
+
+Overview
+--------
+
+Aggregation operations process data in your MongoDB collections and return computed results. The **aggregation pipeline**
+is like a factory assembly line. Along the line are assembly stations, called **aggregation stages** . Like an assembly
+station configured with specific tooling to perform its task, each ``stage`` uses **expressions** to perform some
+operation or combination of operations.
+
+Aggregation Does More
```
ccho-mongodb, line 21: Nitpick: This should probably specify what you're comparing Aggregation to here. E.g.
"Aggregation is More Flexible than the Find Operation"
Currently, it doesn't match the specificity and organization of information of the rest of the documentation. E.g. sections that describe what it is, show how you use it, or what you need to set up to use it

source/fundamentals/aggregation.txt
```diff
@@ -0,0 +1,237 @@
...
+   :depth: 2
+   :class: singlecol
+
+Overview
+--------
+
+Aggregation operations process data in your MongoDB collections and return computed results. The **aggregation pipeline**
+is like a factory assembly line. Along the line are assembly stations, called **aggregation stages** . Like an assembly
+station configured with specific tooling to perform its task, each ``stage`` uses **expressions** to perform some
+operation or combination of operations.
```
ccho-mongodb, line 19: The sentences switch between extended metaphor and simile which makes it a little more difficult to read. I would stick to one of the two throughout the paragraph. E.g.

Extended metaphor:
The aggregation pipeline is a factory assembly line.  The sequentially-arranged assembly stations are aggregation stages. Each assembly station at which a specific task is performed is an aggregation stage with a specific operation or combination of operations. 

Simile (compare two things):
Aggregation pipelines are like a factory assembly line. Aggregation stages are like the assembly stations that sequentially put together the finished product. Similar to assembly stations, aggregation stages use specific tools called expressions to perform a specific operation or combination of operations.

source/fundamentals/aggregation.txt
```diff
@@ -0,0 +1,237 @@
...
+
+Using ``find`` operations, you can:
+
+- select *what* documents to return
+- select *what* fields to return
+- sort the results
+
+Using ``aggregation`` operations, you can:
+
+- perform all ``find`` operations
```
ccho-mongodb, line 32: I think we should take the approach of being objective and list the drawbacks as well such as memory constraints rather than advocating only for aggregation over find.

source/fundamentals/aggregation.txt
```diff
@@ -0,0 +1,237 @@
...
+Using the **Aggregation builders** API, the same operation may look something like:
+
+.. literalinclude:: /includes/fundamentals/code-snippets/agg-factory.java
+   :language: java
+
+.. code-block:: json
+
+   {"_id": 0, "red": 3, "blue": 2}
+
+Terms
```
ccho-mongodb, line 64: Would it be possible to add a sentence under this heading to explain that these are links at which the ? We don't have any similarly-named sections across docs-java so the purpose of this section might not be easily understood.

source/fundamentals/aggregation.txt
```diff
@@ -0,0 +1,237 @@
...
+- :manual:`Aggregation stages </meta/aggregation-quick-reference/#stages>`
+- :manual:`Operator expressions </meta/aggregation-quick-reference/#operator-expressions>`
+- :ref:`Aggregation Builders <builders>`
+
+Runnable Examples
+-----------------
+
+The following examples require a ``restaurants`` collection in the ``test`` database. To create and populate the collection,
+follow the `directions <https://github.com/mongodb/docs-assets/tree/drivers>`_ to download and load the dataset. If you
+prefer to load the data directly and have both ``mongoimport`` and ``curl`` in your path:
```
ccho-mongodb, line 77:
```suggestion
prefer to load the data directly and have both ``mongoimport`` and ``curl`` in your path, you can run the following instead:
```

source/fundamentals/aggregation.txt
```diff
@@ -0,0 +1,237 @@
...
+
+.. tabs::
+
+   .. tab::
+      :tabid: Atlas
+
+      .. code-block:: sh
+
+         mongoimport --uri mongodb+srv://<your-atlas-uri>/test --collection restaurants --drop --file <(curl https://raw.githubusercontent.com/mongodb/docs-assets/drivers/restaurants.json)
```
ccho-mongodb, line 86: Nitpick: what happens here when curl fails? Maybe we can instead split it into two commands so that the user would see the related error from curl before attempting mongoimport? Same comment for localhost tab.

source/fundamentals/aggregation.txt
```diff
@@ -0,0 +1,237 @@
...
+
+Terms
+~~~~~
+
+- :manual:`Aggregation pipeline </core/aggregation-pipeline/>`
+- :manual:`Aggregation stages </meta/aggregation-quick-reference/#stages>`
+- :manual:`Operator expressions </meta/aggregation-quick-reference/#operator-expressions>`
+- :ref:`Aggregation Builders <builders>`
+
+Runnable Examples
```
ccho-mongodb, line 72: Would you consider adding sub-headings under this for "Setup" and then "Aggregation Example"? When I read this heading, I expect to see the sample code rather than the initial setup so it could be helpful to identify what's happening in each paragraph.

source/fundamentals/aggregation.txt
```diff
@@ -0,0 +1,192 @@
...
+.. code-block:: java
+
+   public class AggTourOne {
+
+     public static void main(String[] args) {
+       MongoClient mongoClient = MongoClients.create();
+       MongoDatabase database = mongoClient.getDatabase("test");
+       MongoCollection<Document> collection = database.getCollection("restaurants");
+
+
```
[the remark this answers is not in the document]
ccho-mongodb: Just wanted to make sure this was visible since it looks like you gave it thumbs, but doesn't look like any changes were made.

source/fundamentals/aggregation.txt
```diff
@@ -0,0 +1,237 @@
...
+   }
+
+.. tip::
+
+   For information on connecting to MongoDB, see the :ref:`Connection Guide <mongoclient>`
+
+Perform Aggregation
+~~~~~~~~~~~~~~~~~~~
+
+To perform an aggregation, pass a list of :manual:`aggregation stages </meta/aggregation-quick-reference/#stages>` to
```
ccho-mongodb, line 142: If this is under the Terms section, does it need to be linked again?

source/fundamentals/aggregation.txt
```diff
@@ -0,0 +1,237 @@
...
+
+   collection.aggregate(
+         Arrays.asList(
+                 Aggregates.match(Filters.eq("categories", "Bakery")),
+                 Aggregates.group("$stars", Accumulators.sum("count", 1))
+         )
+   ).forEach(printBlock);
+
+.. code-block:: json
+
```
ccho-mongodb, line 176: Just to clean up the UI:
```suggestion
    :copyable: false
```

source/fundamentals/aggregation.txt
```diff
@@ -0,0 +1,237 @@
...
+.. code-block:: java
+
+   collection.aggregate(
+         Arrays.asList(
+                 Aggregates.match(Filters.eq("categories", "Bakery")),
+                 Aggregates.group("$stars", Accumulators.sum("count", 1))
+         )
+   ).forEach(printBlock);
+
```
ccho-mongodb, line 174: I like to separate the output from sample code by explaining that running the code should produce the following output. Random example (authored by Katie) https://github.com/mongodb/docs-java/blob/master/source/fundamentals/crud/read-operations/limit.txt#L67
Would you agree with this or otherwise, is it not a good idea?

source/fundamentals/aggregation.txt
```diff
@@ -0,0 +1,237 @@
...
+   {"_id": 5, "count": 1}
+
+Use Aggregation Expressions
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The Java driver provides builders for :java-sync-api:`accumulator expressions<com/mongodb/client/model/Accumulators.html>`
+for use with :java-sync-api:`$group <com/mongodb/client/model/Aggregates.html#group(TExpression,java.util.List)>`. You
+must declare all other expressions in JSON format or compatible document format.
+
+.. example::
```
ccho-mongodb, line 187: To me, the formatting is confusing when using this in the "Runnable Examples" section -- this example is within the "Example" box, but the other examples are not. I wonder if we should identify each example with a sub-heading that includes "Example" similar to Kafka docs here https://docs.mongodb.com/kafka-connector/current/kafka-sink-postprocessors.

source/fundamentals/aggregation.txt
```diff
@@ -0,0 +1,237 @@
...
+
+   .. code-block:: java
+
+      new Document("$arrayElemAt", Arrays.asList("$categories", 0))
+
+   .. code-block:: java
+
+      Document.parse("{ $arrayElemAt: ['$categories', 0] }")
+
+In the following example, the aggregation pipeline uses a
```
ccho-mongodb, line 201: Would it be helpful to explain all the Projections static functions called or provide links to the api docs for them?

ccho-mongodb (changes requested):

source/fundamentals/aggregation.txt
```diff
@@ -0,0 +1,237 @@
...
+The following examples require a ``restaurants`` collection in the ``test`` database. To create and populate the collection,
+follow the `directions <https://github.com/mongodb/docs-assets/tree/drivers>`_ to download and load the dataset. If you
+prefer to load the data directly and have both ``mongoimport`` and ``curl`` in your path:
+
+.. tabs::
+
+   .. tab::
+      :tabid: Atlas
+
+      .. code-block:: sh
```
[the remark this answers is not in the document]
terakilobyte: thanks for the info

source/fundamentals/aggregation.txt
terakilobyte: This was meant to stand out.

source/fundamentals/aggregation.txt
terakilobyte: I don't think so in this case. I think it will we be more user friendly to ultimately link to the builders pages we create as part of the docs site.

source/fundamentals/aggregation.txt
terakilobyte: I felt it was self-evident, but perhaps that's an incorrect position

source/fundamentals/aggregation.txt
terakilobyte: I'm not convinced the value separating the two would provide in the unlikely scenario the curl fails for a small 10 document demo collection is worth worrying over. Open to more discussion on this in private channels.

source/fundamentals/aggregation.txt
```diff
@@ -0,0 +1,276 @@
...
+- perform all ``find`` operations
+- rename fields
+- calculate fields
+- summarize data
+- group values
+
+Aggregation operations have some :manual:`limitations <core/aggregation-pipeline-limits/>` you must keep in mind:
+
+- Returned documents must not violate the :manual:`BSON Document Size limit </reference/limits/#BSON-Document-Size>`,
+  currently 16 megabytes.
```
ccho-mongodb, line 45: Thought it was strange to mix title case and normal case within the same term. Also thought "currently" signaled that it's a value that changes often.

```suggestion
- Returned documents must not violate the :manual:`BSON document size limit </reference/limits/#BSON-Document-Size>`
  of 16 megabytes.
```

source/fundamentals/aggregation.txt
```diff
@@ -0,0 +1,276 @@
...
+.. code-block:: java
+
+   public class AggTour {
+
+       public static void main(String[] args) {
+           MongoClient mongoClient = MongoClients.create();
+           MongoDatabase database = mongoClient.getDatabase("test");
+           MongoCollection<Document> collection = database.getCollection("restaurants");
+
+           // aggregation here
```
ccho-mongodb, line 155: This was confusing to me initially, but think i figured it out by looking at the example. To make this a little more clear, I might say "// placeholder for your aggregation" and then either "// your connection setup code" or omit completely in the example on line 193.

source/fundamentals/aggregation.txt
```diff
@@ -0,0 +1,276 @@
...
+- Uses  a :manual:`$group <reference/operator/aggregation/group/>` stage to group the matching documents by the ``stars``
+  field, accumulating a count of documents for each distinct value of ``stars``.
+
+.. note::
+
+   The example uses `Aggregates.group <https://mongodb.github.io/mongo-java-driver/4.2/apidocs/mongodb-driver-core/com/mongodb/client/model/Aggregates.html#group(org.bson.conversions.Bson)>`__
+   to build the ``$group`` stage and `Accumulators.sum <https://mongodb.github.io/mongo-java-driver/4.2/apidocs/mongodb-driver-core/com/mongodb/client/model/Accumulators#sum(java.lang.String,TExpression)>`__
+   to build the accumulator expression. The :manual:`accumulator expressions <reference/operator/aggregation/#accumulators-group>`
+   used within the ``$group`` stage are built with the `Accumulators <https://mongodb.github.io/mongo-java-driver/4.2/apidocs/mongodb-driver-core/com/mongodb/client/model/Accumulators.html>`__
+   helper class.
```
ccho-mongodb, line 189: I think we should use active voice to clearly identify the who and how. E.g. "You can build the accumulator expressions used within the $group stage using the Accumulators helper class. "

source/fundamentals/aggregation.txt
```diff
@@ -0,0 +1,276 @@
...
+contains builders for aggregation stages.
+
+In the following example, the aggregation pipeline:
+
+- Uses a :manual:`$match <reference/operator/aggregation/match/>` stage to filter for documents whose
+  ``categories`` array field contains the element ``Bakery``. The example uses
+  `Aggregates.match <https://mongodb.github.io/mongo-java-driver/4.2/apidocs/mongodb-driver-core/com/mongodb/client/model/Aggregates.html#match(org.bson.conversions.Bson)>`__
+  to build the ``$match`` stage.
+
+- Uses  a :manual:`$group <reference/operator/aggregation/group/>` stage to group the matching documents by the ``stars``
```
ccho-mongodb, line 180:
```suggestion
- Uses a :manual:`$group <reference/operator/aggregation/group/>` stage to group the matching documents by the ``stars``
```

source/fundamentals/aggregation.txt
```diff
@@ -0,0 +1,276 @@
...
+   import org.bson.Document;
+
+   import java.util.Arrays;
+
+Connect to a MongoDB Deployment
++++++++++++++++++++++++++++++++
+
+Connect to a MongoDB deployment and declare and define ``MongoDatabase`` and ``MongoCollection`` instances.
+
+For example, include the following code to connect to a standalone MongoDB deployment running on localhost on port
```
ccho-mongodb, line 142:
```suggestion
Include the following code to connect to a standalone MongoDB deployment running on localhost on port
```

source/fundamentals/aggregation.txt
```diff
@@ -0,0 +1,276 @@
...
+- :manual:`Aggregation pipeline </core/aggregation-pipeline/>`
+- :manual:`Aggregation stages </meta/aggregation-quick-reference/#stages>`
+- :manual:`Operator expressions </meta/aggregation-quick-reference/#operator-expressions>`
+- :ref:`Aggregation Builders <builders>`
+
+Runnable Examples
+-----------------
+
+The following examples require a ``restaurants`` collection in the ``test`` database. To create and populate the collection,
+follow the `directions <https://github.com/mongodb/docs-assets/tree/drivers>`_ to download and load the dataset. If you
```
ccho-mongodb, line 96:
```suggestion
follow `these directions <https://github.com/mongodb/docs-assets/tree/drivers>`_ to download and load the dataset. If you
```

source/includes/fundamentals/code-snippets/agg-factory.java
```diff
@@ -0,0 +1,17 @@
+Aggregates.match(
```
ccho-mongodb, line 1: Not a big deal, but the filename standard I was using was to use ".rst" for code snippets (e.g. auth-credentials-default.rst") and ".java" (using camelcase to adhere to Java class naming standards, but is not visible in the srv.java file) for full classes.

source/fundamentals/aggregation.txt
terakilobyte: Okay, I can see how there can be disconnect.

ccho-mongodb (approved):
LGTM, left a few individual comments.

source/includes/fundamentals/code-snippets/agg-factory.java
terakilobyte: Quite right

source/fundamentals/aggregation.txt
terakilobyte: I've removed most of this entirely, since it's getting too deep in the weeds and there are going to be other pages that cover these in more detail.

terakilobyte:
@rozza @jyemin ready for review

source/fundamentals/aggregation.txt
```diff
@@ -0,0 +1,233 @@
...
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+Overview
+--------
+
+Aggregation operations process data in your MongoDB collections and return computed results.
+
+A powerful way to think of aggregation is like a car factory. Within the car factory is an assembly line, along which
```
jyemin, line 18: Editorial nit: I would not claim powerful.  That's in the eye of the beholder, and may turn off some readers who don't find it powerful, or are still confused.  Perhaps just "One way to think of..."

source/fundamentals/aggregation.txt
```diff
@@ -0,0 +1,233 @@
...
+   :start-after: begin insert
+   :end-before: end insert
+
+Basic Aggregation Example
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+To perform an aggregation, pass a list of aggregation stages to the `MongoCollection.aggregate() <https://mongodb.github.io/mongo-java-driver/4.2/apidocs/mongodb-driver-core/com/mongodb/client/MongoCollection.html#aggregate(com.mongodb.client.ClientSession,java.util.List,java.lang.Class)>`__
+method.
+
+The Java driver provides the `Aggregates <https://mongodb.github.io/mongo-java-driver/4.2/apidocs/mongodb-driver-core/com/mongodb/client/model/Aggregates.html>`__
```
jyemin, line 148: I feel like we should make clear somewhere that it's not required to use the builders.  Not sure whether this is the place though.  What do you think?

source/fundamentals/aggregation.txt
terakilobyte: Great point, thanks!

source/fundamentals/aggregation.txt
terakilobyte: I was going to frame this in the builders page, and frame it as "here's this manually constructed aggregation, here's the builder version".

terakilobyte:
@jyemin I've addressed your comment

jyemin (approved):
LGTM

PR #55, opened 2021-02-08
DOCSP-9555: indexes

terakilobyte:
## Pull Request Info

Adding coverage for indexes. A link to the builders page will be added as part of that PR.

### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-9555

https://docs-mongodbcom-staging.corp.mongodb.com/1bf9dff/java/docsworker-xlarge/DOCSP-9555/fundamentals/indexes/

### Self-Review Checklist

### DIFF
--- a/source/fundamentals/indexes.txt
+++ b/source/fundamentals/indexes.txt
@@ -16,7 +16,7 @@ Overview
 --------
 
 Indexes support the efficient execution of queries in MongoDB. Without indexes, MongoDB must scan *every* document in a
-collection (a *collection scan*) to find the documents that match each query. These collection scans are slow and can
+collection (a **collection scan**) to find the documents that match each query. These collection scans are slow and can
 negatively affect the performance of your application. If an appropriate index exists for a query, MongoDB can use the
 index to limit the number of documents it must inspect.
 
@@ -31,28 +31,26 @@ When you execute a query against MongoDB, your command can include various eleme
 - projection criteria to specify the fields MongoDB should return (optional)
 - sort criteria to specify the order documents will be returned from MongoDB (optional)
 
-When all the fields specified in the query, projection, and sort are indexed, MongoDB returns results
-directly from the index without scanning any documents in the collection or loading them into memory.
+When all the fields specified in the query, projection, and sort are indexed, MongoDB returns results directly from the
+index without scanning any documents in the collection or loading them into memory, also called a **covered query**.
 
 .. important:: Sort Order
 
-   Set your sort criteria to match or invert the order that you specified the index.
+   Sort criteria must match or invert the order of the index.
 
-   .. example::
+   Consider an index on the fields "name" in ascending order (A-Z), "age" in descending order (9-0):
 
-      Consider an index on the fields "name" in ascending order (A-Z), "age" in descending order (9-0):
+   .. code-block:: json
 
-      .. code-block:: json
+      name_1_age_-1
 
-         name_1_age_-1
+   MongoDB would use this index when you sort your data by either:
 
-      MongoDB would use this index when you sort your data by either:
+   - "name" ascending, "age" descending
+   - "name" descending, "age" ascending
 
-      - "name" ascending, "age" descending
-      - "name" descending, "age" ascending
-
-      Specifying a sort order of "name" and "age" ascending or "name" and "age" descending would require an in-memory
-      sort.
+   Specifying a sort order of "name" and "age" ascending or "name" and "age" descending would require an in-memory
+   sort.
 
 For additional information on how to ensure your index covers your query criteria and projection, see the MongoDB manual
 articles on :manual:`query coverage </core/query-optimization/#read-operations-covered-query>`.
@@ -111,19 +109,25 @@ The following example creates an index in ascending order on the ``title`` field
 
 .. common-content-end
 
+.. driver-content-begin
 .. literalinclude:: /includes/fundamentals/code-snippets/IndexPage.java
    :language: java
    :dedent:
    :start-after: begin single index
    :end-before: end single index
+.. driver-content-end
 
 The following is an example of a query that would be covered by the index created above:
 
+.. driver-content-begin
 .. literalinclude:: /includes/fundamentals/code-snippets/IndexPage.java
    :language: java
    :dedent:
    :start-after: begin covered single query
    :end-before: end covered single query
+.. driver-content-end
+
+.. common-content-begin
 
 See the MongoDB server manual section on :manual:`single field indexes </core/index-single>` for more information.
 
@@ -135,7 +139,9 @@ for *multiple* fields of a document. You must specify the direction (ascending o
 index.
 
 The following example creates a compound index on the ``type`` and ``rated`` fields:
+.. common-content-end
 
+.. driver-content-begin
 .. literalinclude:: /includes/fundamentals/code-snippets/IndexPage.java
    :language: java
    :dedent:
@@ -149,7 +155,9 @@ The following is an example of a query that would be covered by the index create
    :dedent:
    :start-after: begin covered compound query
    :end-before: end covered compound query
+.. driver-content-end
 
+.. common-content-begin
 
 See the MongoDB server manual section on :manual:`Compound indexes </core/index-compound>` for more information.
 
@@ -162,6 +170,8 @@ an array value. You can define a multikey index using the same syntax as a singl
 The following example creates a compound, multikey index on the ``rated``, ``genres`` (an array of Strings), and
 ``title`` fields:
 
+.. common-content-end
+.. driver-content-begin
 .. literalinclude:: /includes/fundamentals/code-snippets/IndexPage.java
    :language: java
    :dedent:
@@ -175,6 +185,9 @@ The following is an example of a query that would be covered by the index create
    :dedent:
    :start-after: begin covered multikey query
    :end-before: end covered multikey query
+.. driver-content-end
+
+.. common-content-begin
 
 Multikey indexes behave differently from non-multikey indexes in terms of query coverage, index bound computation, and
 sort behavior. For a full explanation of multikey indexes, including a discussion of their behavior and limitations,
@@ -190,6 +203,10 @@ language as an option when creating the index.
 .. warning::
 
    Collections are limited to one text index. Keep in mind that a text index can cover multiple fields.
+
+.. common-content-end
+
+.. driver-content-begin
 
 The following example creates a text index on the ``plot`` field:
 
@@ -207,6 +224,9 @@ omitted because text indexes do not contain sort order.
    :dedent:
    :start-after: begin text query
    :end-before: end text query
+
+.. driver-content-end
+.. common-content-begin
 
 For a full explanation of text search with MongoDB, refer to :manual:`Text Indexes </core/index-text>` in the MongoDB
 manual.
@@ -252,6 +272,9 @@ The following example creates a ``2dsphere`` index on the ``location.geo`` field
 
    Attemping to create a geospatial index on a field that is covered by a geospatial index will result in an error.
 
+.. common-content-end
+.. driver-content-begin
+
 .. literalinclude:: /includes/fundamentals/code-snippets/IndexPage.java
    :language: java
    :dedent:
@@ -266,6 +289,8 @@ The following is an example of a geospatial query using the "location.geo" index
    :start-after: begin geospatial query
    :end-before: end geospatial query
 
+.. driver-content-end
+.. common-content-begin
 
 MongoDB also supports ``2d`` indexes for calculating distances on a Euclidean plane and for working with the "legacy
 coordinate pairs" syntax used in MongoDB 2.2 and earlier. See the :manual:`Geospatial Queries page </geospatial-queries>`
@@ -279,6 +304,9 @@ on the ``_id`` field during the creation of a collection. To create a unique ind
 fields that you want to prevent duplication on and set the ``unique`` option to ``true``.
 
 The following example creates a unique, descending index on the ``theaterId`` field:
+
+.. common-content-end
+.. driver-content-begin
 
 .. literalinclude:: /includes/fundamentals/code-snippets/IndexPage.java
    :language: java
@@ -296,3 +324,4 @@ The following example creates a unique, descending index on the ``theaterId`` fi
    E11000 duplicate key error index
 
 Refer to the :manual:`Unique Indexes page </core/index-unique>` in the MongoDB server manual for more information.
+.. driver-content-end
--- a/source/fundamentals/indexes.txt
+++ b/source/fundamentals/indexes.txt
@@ -1,0 +1,253 @@
+=======
+Indexes
+=======
+
+.. default-domain:: mongodb
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+Overview
+--------
+
+Indexes are data structures that support the efficient execution of queries in MongoDB. They contain copies of parts of
+the data in documents to make queries more efficient.
+
+Without indexes, MongoDB must scan *every* document in a collection to find the documents that match each query. These
+collection scans are slow and can negatively affect the performance of your application. By using an index to limit the
+number of documents MongoDB scans, queries can be more efficient and return faster.
+
+
+Query Coverage and Performance
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+When you execute a query against MongoDB, your query can include three parts:
+
+- query criteria that specify field(s) and value(s) you are looking for
+- options that affect the query's execution (e.g. read concern)
+- projection criteria to specify the fields MongoDB should return (optional)
+
+When all the fields specified in the query criteria and projection of a query are indexed, MongoDB returns results
+directly from the index without scanning any documents in the collection or loading them into memory.
+
+For additional information on how to ensure your index covers your query criteria and projection, see the MongoDB manual
+articles on :manual:`query coverage </core/query-optimization/#read-operations-covered-query>`.
+
+Operational Considerations
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+To improve query performance, build indexes on fields that appear often in your application's queries and operations
+that return sorted results. Each index that you add consumes disk space and memory when active so you should track index
+memory and disk usage for capacity planning. In addition, when a write operation updates an indexed field, MongoDB also
+has to update the related index.
+
+For more information on designing your data model and choosing indexes appropriate for your application, see the MongoDB
+server documentation on :manual:`Indexing Strategies </applications/indexes>` and
+:manual:`Data Modeling and Indexes </core/data-model-operations/#data-model-indexes>`.
+
+Index Types
+-----------
+
+MongoDB supports a number of different index types to support querying your data. The following sections describe the
+most common index types and provide sample code for creating each index type.
+
+.. tip::
+
+   The MongoDB Java Driver provides the :java-core-api:`Indexes </com/mongodb/client/model/Indexes.html>` class that
+   includes static factory methods to create index specification documents for different MongoDB Index key types.
+
+Single Field and Compound Indexes
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Single Field Indexes
+++++++++++++++++++++
+
+**Single field indexes** are indexes that improve performance for queries that specify ascending or descending sort
+order on a single field of a document.
+
+The following example uses the ``createIndex()`` method to create an ascending order index on the ``title`` field in the
+``movies`` collection in the ``sample_mflix`` database.
+
+.. literalinclude:: /includes/fundamentals/code-snippets/IndexPage.java
+   :language: java
+   :dedent:
+   :start-after: begin single index
+   :end-before: end single index
+
+The following is an example of a query that would be covered by the index created above.
+
+.. literalinclude:: /includes/fundamentals/code-snippets/IndexPage.java
+   :language: java
+   :dedent:
+   :start-after: begin covered single query
+   :end-before: end covered single query
+
+See the MongoDB server manual section on :manual:`single field indexes </core/index-single>` for more information.
+
+Compound Indexes
+++++++++++++++++
+
+**Compound indexes** are indexes that improve performance for queries that specify ascending or descending sort order
+for *multiple* fields of a document. You must specify the direction (ascending or descending) for each field in the
+index.
+
+The following example uses the ``createIndex()`` method to create a compound index on the ``type`` and ``rated`` fields
+in the ``movies`` collection in the ``sample_mflix`` database.
+
+.. literalinclude:: /includes/fundamentals/code-snippets/IndexPage.java
+   :language: java
+   :dedent:
+   :start-after: begin compound index
+   :end-before: end compound index
+
+The following is an example of a query that would be covered by the index created above.
+
+.. literalinclude:: /includes/fundamentals/code-snippets/IndexPage.java
+   :language: java
+   :dedent:
+   :start-after: begin covered compound query
+   :end-before: end covered compound query
+
+
+See the MongoDB server manual section on :manual:`Compound indexes </core/index-compound>` for more information.
+
+Multikey Indexes (Indexes on Array Fields)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+**Multikey indexes** are indexes that improve performance for queries that specify ascending or descending indexes on
+fields that contain an array value. You can define a multikey index using the same syntax as a single field or compound
+index.
+
+The following example use the ``createIndex()`` method to create a compound, multikey index on the ``rated``,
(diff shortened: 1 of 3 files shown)


source/fundamentals/indexes.txt
```diff
@@ -0,0 +1,253 @@
...
+Single Field and Compound Indexes
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Single Field Indexes
+++++++++++++++++++++
+
+**Single field indexes** are indexes that improve performance for queries that specify ascending or descending sort
+order on a single field of a document.
+
+The following example uses the ``createIndex()`` method to create an ascending order index on the ``title`` field in the
```
nathan-contino-mongo, line 71: In the Realm docs, we'd usually link to the reference doc for `createIndex()` here. Would that be a good idea here?

source/fundamentals/indexes.txt
```diff
@@ -0,0 +1,253 @@
...
+``movies`` collection in the ``sample_mflix`` database.
```
nathan-contino-mongo, line 72: This code snippet doesn't necessarily refer to a "movies" collection of a "sample_mflix" database. Should we add some lines to this snippet instantiating those objects, or should we just make this line more generic with something like "...field in a collection"?

source/fundamentals/indexes.txt
```diff
@@ -0,0 +1,253 @@
...
+
+Compound Indexes
+++++++++++++++++
+
+**Compound indexes** are indexes that improve performance for queries that specify ascending or descending sort order
+for *multiple* fields of a document. You must specify the direction (ascending or descending) for each field in the
+index.
+
+The following example uses the ``createIndex()`` method to create a compound index on the ``type`` and ``rated`` fields
+in the ``movies`` collection in the ``sample_mflix`` database.
```
nathan-contino-mongo, line 98: Same as above -- should we add lines to the code snippet demonstrating instantiating the "movies" collection and "sample_mflix" database, or make this line more generic?

source/includes/fundamentals/code-snippets/IndexPage.java
```diff
@@ -0,0 +1,150 @@
...
+        page.multiKeyIndex();
+        page.textIndex();
+        page.geoSpatialIndex();
+    }
+
+    private void singleIndex() {
+        System.out.println("single index");
+        // begin single index
+        String resultCreateIndex = collection.createIndex(Indexes.ascending("title"));
+        System.out.println(String.format("Index created: %s", resultCreateIndex));
```
nathan-contino-mongo, line 45: Is there a style reason you'd use String.format here instead of printf, or just personal choice? I'd think that printf would be easier to understand for intermediate-level developers.

source/fundamentals/indexes.txt
nathan-contino-mongo: I noticed that you don't use a colon to segue these snippet descriptions into the snippets themselves. I think most MongoDB documentation uses colons for situations like these -- would it make sense to switch from period to colons?

source/fundamentals/indexes.txt
```diff
@@ -0,0 +1,253 @@
...
+   :end-before: end covered compound query
+
+
+See the MongoDB server manual section on :manual:`Compound indexes </core/index-compound>` for more information.
+
+Multikey Indexes (Indexes on Array Fields)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+**Multikey indexes** are indexes that improve performance for queries that specify ascending or descending indexes on
+fields that contain an array value. You can define a multikey index using the same syntax as a single field or compound
```
nathan-contino-mongo, line 121: Personal opinion here: I think this sentence gets a bit lost in the weeds of "queries that specify ascending or descending indexes" to a point where it's easy to miss the actual idea (multikey indexes are indexes on fields that contain an array value). Arguably you've mitigated this issue in the title of this subsection, but I wonder if we could trim this sentence a little to make it less wordy.

source/includes/fundamentals/code-snippets/IndexPage.java
```diff
@@ -0,0 +1,150 @@
...
+        // end covered multikey query
+        cursor.forEach(doc -> System.out.println(doc));
+    }
+
+    private void textIndex() {
+        System.out.println("text index");
+        // begin text index
+        // create a text index of the "fullplot" field in the "movies" collection
+        // if a text index already exists with a different configuration, this will
+        // error
```
nathan-contino-mongo, line 95: This comment brings up a pretty important footgun w.r.t. text indexes -- the fact that you can only have one per collection. The rest of the comment is rendered redundant by the description. Would it make sense to lift this comment out into a warning about how you can only create one text index per collection? Otherwise I think it's unclear why this "will error".

Also, this comment mentions the "fullplot" field but the code itself uses the "plot" field. Which is correct?

source/fundamentals/indexes.txt
```diff
@@ -0,0 +1,253 @@
...
+
+For a full explanation of text search with MongoDB, refer to :manual:`Text Indexes </core/index-text>` in the MongoDB
+manual.
+
+Geospatial Indexes
+~~~~~~~~~~~~~~~~~~
+
+MongoDB supports queries of geospatial coordinate data using **2dsphere indexes**. With a 2dsphere index, you can query
+the geospatial data for inclusion, intersection, and proximity. For more information on querying geospatial data, see
+:manual:`Geospatial Queries </geospatial-queries/>`.
```
nathan-contino-mongo, line 178: This and other sections include the "For a full explanation" sentence at the very end of the section that already links to this page. The Node driver documentation contains a page on geospatial queries: https://docs.mongodb.com/drivers/node/fundamentals/crud/read-operations/geo
Would it make more sense to link to the Java driver docs equivalent page here, since this link is rendered redundant by the "for a full explanation" link at the end of the section?

source/fundamentals/indexes.txt
```diff
@@ -0,0 +1,253 @@
...
+            "type" : "Point",
+            "coordinates" : [
+               -118.36559,
+               33.897167
+            ]
+         }
+      }
+   }
+
+The following example uses the ``createIndexes()`` method to create a ``2dsphere`` index on the ``location.geo`` field
```
nathan-contino-mongo, line 208: I noticed that index names are sometimes rendered in monospace text, sometimes plaintext. I think it makes sense for the "text" index above since that's the literal method name you use to create the index, but that's not the case here. Should we use "geo2dsphere" in monospace here instead? Ditto for some of the other index sections regarding monospacing vs. plaintext.

nathan-contino-mongo (approved):
LGTM with a few nits/suggestions.

source/includes/fundamentals/code-snippets/IndexPage.java
terakilobyte: I used this to avoid having to do `System.out.printf("Index created: %s%n", resultCreateIndex);` for the newline.

source/fundamentals/indexes.txt
```diff
@@ -59,6 +59,17 @@ most common index types and provide sample code for creating each index type.
    The MongoDB Java Driver provides the :java-core-api:`Indexes </com/mongodb/client/model/Indexes.html>` class that
    includes static factory methods to create index specification documents for different MongoDB Index key types.
 
+The following examples use the
+:java-sync-api:`createIndex </com/mongodb/client/MongoCollection.html#createIndex(org.bson.conversions.Bson,com.mongodb.client.model.IndexOptions)>`
```
nathan-contino-mongo, line 63: Noticed something here that we struggle with in our Usage Examples for the Realm SDKs -- normally I like the idea of linking once to a concept at first mention. But for pages where individual sections stand alone as examples, I think it's sometimes useful to link one *per section*, especially since otherwise users might have to hunt through the javadoc to find the ``createIndex()`` method.

source/fundamentals/indexes.txt
```diff
@@ -117,20 +126,19 @@ See the MongoDB server manual section on :manual:`Compound indexes </core/index-
 Multikey Indexes (Indexes on Array Fields)
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-**Multikey indexes** are indexes that improve performance for queries that specify ascending or descending indexes on
-fields that contain an array value. You can define a multikey index using the same syntax as a single field or compound
-index.
+**Multikey indexes** are indexes that improve performance for queries that specify an index on a field that contains an
```
nathan-contino-mongo, line 129: Huge improvement here -- this gets to the point with a lot fewer words. Would this be even clearer if we removed "that specify an index", since queries do not have to explicitly specify an index to use it? Or do you think that's cutting too many words?

source/fundamentals/indexes.txt
```diff
@@ -149,16 +157,19 @@ Text Indexes
 string or an array of string elements. MongoDB supports text search for various languages. You can specify the default
 language as an option when creating the index.
 
-The following example uses the ``createIndex()`` method to create a ``text`` index on the ``fullplot`` field in the
-``movies`` collection in the ``sample_mflix`` database.
+.. warning::
+
+   Collections are limited to one text index. Keep in mind that a text index can cover multiple fields.
```
nathan-contino-mongo, line 162: Good call on the second sentence here!

source/fundamentals/indexes.txt
```diff
@@ -205,8 +216,11 @@ database is a GeoJSON Point object that describes the coordinates of the theater
       }
    }
 
-The following example uses the ``createIndexes()`` method to create a ``2dsphere`` index on the ``location.geo`` field
-in the ``theaters`` collection in the ``sample_mflix`` database to enable geospatial searches.
+The following example creates a ``2dsphere`` index on the ``location.geo`` field:
+
+.. note::
+
+   Attemping to create a geospatial index on a field that is covered by a geospatial index will result in an error.
```
nathan-contino-mongo, line 223: Is this a common enough problem to warrant a note? To me, this seems intuitive enough to omit, but maybe I'm missing something? If it helps even a small number of users, I think it's worth keeping, though.

source/fundamentals/indexes.txt
```diff
@@ -234,17 +248,18 @@ Unique indexes ensure that the indexed fields do not store duplicate values. By
...
    :dedent:
    :start-after: begin unique index
    :end-before: end unique index
 
-If you attempt to perform a write operation that stores a duplicate value that violates the unique index, the MongoDB
-Java driver will raise a ``DuplicateKeyException``, and MongoDB will throw an error resembling the following:
+.. warning::
+
+   If you attempt to perform a write operation that stores a duplicate value that violates the unique index, the MongoDB
+   Java driver will raise a ``DuplicateKeyException``, and MongoDB will throw an error resembling the following:
```
nathan-contino-mongo, line 262: Really liking the "footguns in notes" pattern here that you've established for each section.

nathan-contino-mongo (approved):
LGTM (again!) with one or two suggestions.

source/fundamentals/indexes.txt
terakilobyte: Normally I'd want to link to the method more, but I think there are two mitigating things here:
1. The builders page, which included updating this page with links to it once it is complete, will serve as the primary page for "this is how you do indexes in Java".
2. In all of the IDEs I've tested, this method almost immediately completes and shows signature help.

I'm hesitant to add more items to the page that make it busier with these two things in mind. Thoughts?

source/fundamentals/indexes.txt
terakilobyte: I've reworded this slightly, good catch. The emphasis should be on the field queried, which I think it is now.

source/fundamentals/indexes.txt
nathan-contino-mongo: Seems fair to assume that the builder links and general norm of using an IDE with Java can fill this gap in understanding. LGTM

source/fundamentals/indexes.txt
terakilobyte: So, the index story in MongoDB isn't uniform. The other index types that I didn't wrap in a try/catch can be created over and over on the same field, MongoDB will just happily return the name of the existing index.

These other types throw an error. Since I've wrapped the operation in a try/catch I thought it best to add a note.

nathan-contino-mongo: That sounds like a norm-defying footgun to me. LGTM.

terakilobyte:
Ready for review @jyemin @rozza

source/fundamentals/indexes.txt
```diff
@@ -0,0 +1,268 @@
...
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+Overview
+--------
+
+Indexes are data structures that support the efficient execution of queries in MongoDB. They contain copies of parts of
```
jyemin, line 16: The use of the phrase `data structures` might confuse people, as typically it is used to refer to in-memory constructs like linked lists or array rather than persistent on-disk representations.  Perhaps instead steal some language from the intro on https://docs.mongodb.com/manual/indexes/?

source/fundamentals/indexes.txt
```diff
@@ -0,0 +1,268 @@
...
+number of documents MongoDB scans, queries can be more efficient and return faster.
+
+
+Query Coverage and Performance
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+When you execute a query against MongoDB, your query can include three parts:
+
+- query criteria that specify field(s) and value(s) you are looking for
+- options that affect the query's execution (e.g. read concern)
```
jyemin, line 30: This isn't really relevant to indexes.  Consider removing.

source/fundamentals/indexes.txt
```diff
@@ -0,0 +1,268 @@
...
+
+Without indexes, MongoDB must scan *every* document in a collection to find the documents that match each query. These
+collection scans are slow and can negatively affect the performance of your application. By using an index to limit the
+number of documents MongoDB scans, queries can be more efficient and return faster.
+
+
+Query Coverage and Performance
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+When you execute a query against MongoDB, your query can include three parts:
```
jyemin, line 27: Also, sort criteria, which can make use of indexes as well.

jyemin, line 27: I'm concerned that this description is getting ahead of what we say on https://docs.mongodb.com/manual/indexes/#index-use.  And it's incomplete as it doesn't cover, for example, geospatial indexes.

source/fundamentals/indexes.txt
```diff
@@ -0,0 +1,268 @@
...
+
+- query criteria that specify field(s) and value(s) you are looking for
+- options that affect the query's execution (e.g. read concern)
+- projection criteria to specify the fields MongoDB should return (optional)
+
+When all the fields specified in the query criteria and projection of a query are indexed, MongoDB returns results
```
jyemin, line 33: I'm not sure we should make this promise.  For example, what about sort criteria?  

I think if we're going to include statements like this in driver docs, we need someone from server query to review as well.

source/fundamentals/indexes.txt
```diff
@@ -0,0 +1,268 @@
...
+   :end-before: end declaration
+
+
+Single Field and Compound Indexes
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Single Field Indexes
+++++++++++++++++++++
+
+**Single field indexes** are indexes that improve performance for queries that specify ascending or descending sort
```
jyemin, line 79: This is the first mention of sort criteria.  And it's also not the only or even primary use of a single field index.  For example, the `_id` index.

jyemin (changes requested):
I'm concerned about this section.  I found several statements that are either inaccurate or incomplete, and I'm not convinced there aren't more subtleties that I missed entirely. There is a lot of information in here that is describing the behavior of the server, not the driver, and driver engineers are not the right ones to verify its accuracy.

source/fundamentals/indexes.txt
terakilobyte: I think it's valuable to frame indexes in how they are used, and this is just a small reminder to readers what a query operation includes. Thoughts?

source/fundamentals/indexes.txt
terakilobyte: Good point!

terakilobyte: We'll have the query team lgtm the "common" parts as delimited in the rst

terakilobyte:
@cswanson310 Could you give a review to the content in rst wrapped in "common-content-x"?

jameswahlin:
> @cswanson310 Could you give a review to the content in rst wrapped in "common-content-x"?

@terakilobyte I can do a review pass for the Query team in Charlie's stead

source/fundamentals/indexes.txt
jyemin: I still find it confusing.  Perhaps if you want to include it, separate it out into a section that makes it clear that it's not relevant to index use.

On a related note, this whole section seems geared towards the find command, though indexes are using for aggregation and updates as well.

source/fundamentals/indexes.txt
terakilobyte: We do include mention of the `_id_` index further down in the page, but I added another callout here. I also added a short tip to let readers know indexes help other operations.

source/fundamentals/indexes.txt
```diff
@@ -0,0 +1,340 @@
...
+
+When you execute a query against MongoDB, your command can include various elements:
+
+- query criteria that specify field(s) and value(s) you are looking for
+- options that affect the query's execution (e.g. read concern)
+- projection criteria to specify the fields MongoDB should return (optional)
+- sort criteria to specify the order documents will be returned from MongoDB (optional)
+
+When all the fields specified in the query, projection, and sort are indexed, MongoDB returns results directly from the
+index without scanning any documents in the collection or loading them into memory, also called a **covered query**.
```
jameswahlin, line 40: I would consider dropping "loading them into memory" as it is not clear what this means? If it is a refers to avoiding an in-memory sort then I think more context will be needed, as the index key order also has to match the sort specification order.

source/fundamentals/indexes.txt
jameswahlin: +1 for the sentiment that sort order is not the only use case for a single field index, filtering being the primary example.

source/fundamentals/indexes.txt
```diff
@@ -0,0 +1,340 @@
...
+.. driver-content-end
+
+.. common-content-begin
+
+See the MongoDB server manual section on :manual:`single field indexes </core/index-single>` for more information.
+
+Compound Indexes
+++++++++++++++++
+
+**Compound indexes** are indexes that improve performance for queries that specify ascending or descending sort order
```
jameswahlin, line 148: Ditto here on sort just being one use case that indexes exist for.

source/fundamentals/indexes.txt
```diff
@@ -0,0 +1,340 @@
...
+.. literalinclude:: /includes/fundamentals/code-snippets/IndexPage.java
+   :language: java
+   :dedent:
+   :start-after: begin text query
+   :end-before: end text query
+
+.. driver-content-end
+.. common-content-begin
+
+For a full explanation of text search with MongoDB, refer to :manual:`Text Indexes </core/index-text>` in the MongoDB
```
jameswahlin, line 242: Maybe worth mentioning Atlas Search here? Maybe controversial as it is Atlas only, but could be a better solution for users on Atlas than text indexes.

source/fundamentals/indexes.txt
```diff
@@ -0,0 +1,340 @@
...
+To improve query performance, build indexes on fields that appear often in your application's queries and operations
+that return sorted results. Each index that you add consumes disk space and memory when active so you should track index
+memory and disk usage for capacity planning. In addition, when a write operation updates an indexed field, MongoDB also
+has to update the related index.
+
+For more information on designing your data model and choosing indexes appropriate for your application, see the MongoDB
+server documentation on :manual:`Indexing Strategies </applications/indexes>` and
+:manual:`Data Modeling and Indexes </core/data-model-operations/#data-model-indexes>`.
+
+Index Types
```
jameswahlin, line 76: Looks like wildcard indexes are missing from this list?

jameswahlin (changes requested):

source/fundamentals/indexes.txt
terakilobyte: Thanks, I've removed the terminology. In regards to your point about sorting, do you think the sort order note that follows this paragraph is sufficient in explaining this?

source/fundamentals/indexes.txt
```diff
@@ -0,0 +1,340 @@
...
+--------
+
+Indexes support the efficient execution of queries in MongoDB. Without indexes, MongoDB must scan *every* document in a
+collection (a **collection scan**) to find the documents that match each query. These collection scans are slow and can
+negatively affect the performance of your application. If an appropriate index exists for a query, MongoDB can use the
+index to limit the number of documents it must inspect.
+
+.. tip::
+
+   Indexes are also used by update operations when finding the document(s) to update, and by
```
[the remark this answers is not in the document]
terakilobyte: good point

terakilobyte:
@jameswahlin Ready for another look. I've simplified some language and removed emphasis on sorting. I've also listed wildcard indexes in the "operational considerations" section. There's also a callout about Atlas search encouraging users to use it if they are on Atlas.

jameswahlin (approved):
LGTM

terakilobyte:
@jyemin Ready for another pass on the driver specific content.

source/includes/fundamentals/code-snippets/IndexPage.java
```diff
@@ -0,0 +1,147 @@
...
+        // begin geospatial query
+        // MongoDB Headquarters in NY, NY.
+        Point refPoint = new Point(new Position(-73.98456, 40.7612));
+        Bson filter = near("location.geo", refPoint, 1000.0, 0.0);
+        FindIterable<Document> cursor = collection.find(filter);
+        // end geospatial query
+        cursor.forEach(doc -> System.out.println(doc));
+    }
+
+    private void uniqueIndex() {
```
jyemin, line 134: This method is not called from main, so it's untested.

source/includes/fundamentals/code-snippets/IndexPage.java
```diff
@@ -0,0 +1,147 @@
...
+        System.out.println("unique index");
+        collection = database.getCollection("theaters");
+        // begin unique index
+        try {
+            IndexOptions indexOptions = new IndexOptions().unique(true);
+            String resultCreateIndex = collection.createIndex(Indexes.descending("theaderId"), indexOptions);
+            System.out.println(String.format("Index created: %s", resultCreateIndex));
+        } catch (DuplicateKeyException e) {
```
jyemin, line 142: Even if I create the index twice in a row, I don't get a DuplicateKeyException (or any exception for that matter).  Is MongoDB just treating the second attempt as a no-op?

jyemin (changes requested):

source/fundamentals/indexes.txt
```diff
@@ -0,0 +1,364 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+When you execute a query against MongoDB, your command can include various elements:
+
+- query criteria that specify field(s) and value(s) you are looking for
+- options that affect the query's execution (e.g. read concern)
+- projection criteria to specify the fields MongoDB should return (optional)
+- sort criteria to specify the order documents will be returned from MongoDB (optional)
+
+When all the fields specified in the query, projection, and sort are indexed, MongoDB returns results directly from the
```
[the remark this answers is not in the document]
terakilobyte: I'll update this, it's an easy clarification to make.

source/includes/fundamentals/code-snippets/IndexPage.java
terakilobyte: The DuplicateKeyException would be thrown if you tried to create an index with unique constraints on a field that [isn't unique](https://docs.mongodb.com/manual/core/index-unique/#restrictions).

```java
    private void uniqueIndex() {
        System.out.println("unique index");
        collection = database.getCollection("uniqueViolator");

        collection.insertMany(Arrays.asList(new Document("a", 1), new Document("a", 1)));
        // begin unique index
        // this will fail if any duplicate keys (values) exist
        try {
            IndexOptions indexOptions = new IndexOptions().unique(true);
            String resultCreateIndex = collection.createIndex(Indexes.descending("a"), indexOptions);
            System.out.println(String.format("Index created: %s", resultCreateIndex));
        } catch (DuplicateKeyException e) {
            System.out.println("duplicate key, couldn't create index");
        }
        // end unique index
    }
```

terakilobyte:
@jyemin I've pushed some clarifications based on your feedback.

source/includes/fundamentals/code-snippets/IndexPage.java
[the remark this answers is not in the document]
terakilobyte: I've also added an imports section to the setup

jyemin (approved):
LGTM

PR #58, opened 2021-03-02
DOCSP-9584: projections

terakilobyte:
## Pull Request Info

Projections builders.

### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-9584

https://docs-mongodbcom-staging.corp.mongodb.com/de17980/java/docsworker-xlarge/DOCSP-9584/fundamentals/builders/projections/

### Self-Review Checklist

### DIFF
--- a/source/fundamentals/builders/projections.txt
+++ b/source/fundamentals/builders/projections.txt
@@ -139,7 +139,7 @@ The output should resemble:
 Combining Projections
 ---------------------
 
-Use the ``fields`` method to combine multiple projections.
+Use the ``fields()`` method to combine multiple projections.
 
 The following example includes the ``year`` and ``type`` fields and excludes the
 ``_id`` field:
@@ -310,7 +310,7 @@ The output should resemble:
 Text Score
 ----------
 
-Use the ``metaTextScore`` method to specify a projection of
+Use the ``metaTextScore()`` method to specify a projection of
 :manual:`score of a text query </reference/operator/query/text/#return-the-text-search-score/>`
 
 The following example projects the text score as the value of the ``score`` field:
--- a/source/fundamentals/builders/projections.txt
+++ b/source/fundamentals/builders/projections.txt
@@ -18,9 +18,9 @@ Overview
 MongoDB supports **field projection**, specifying which fields to include and exclude when returning results from a
 query. Projection in MongoDB follows some basic rules:
 
-- The ``_id`` field is **always** included and must explicitly be excluded
+- The ``_id`` field is **always** included unless explicitly
 - Specifying a field for inclusion implicitly excludes all other fields **except** the ``_id`` field
-- Specifying a field for exclusion removes only that field in a query result
+- Specifying a field for exclusion removes **only** that field in a query result
 
 Find more information about projection mechanics :manual:`here </tutorial/project-fields-from-query-results/>`.
 
@@ -54,6 +54,8 @@ The examples also use the following setup in addition to the ``Projections`` sta
    :language: java
    :dedent:
 
+The collection is set up with the following documents:
+
 .. literalinclude:: /includes/fundamentals/code-snippets/builders/Projections.java
    :start-after: begin collection setup
    :end-before: end collection setup
@@ -73,7 +75,7 @@ The following example includes the ``year`` field and (implicitly) the ``_id`` f
    :language: java
    :dedent:
 
-The output should resemble:
+The following code shows the output from this projection:
 
 .. code-block:: json
    :copyable: false
@@ -90,7 +92,7 @@ The following example includes the ``year`` and ``type`` fields and (implicitly)
    :language: java
    :dedent:
 
-The output should resemble:
+The following code shows the output from this projection:
 
 .. code-block:: json
    :copyable: false
@@ -111,7 +113,7 @@ The following example excludes the ``temperatures`` field:
    :language: java
    :dedent:
 
-The output should resemble:
+The following code shows the output from this projection:
 
 .. code-block:: json
    :copyable: false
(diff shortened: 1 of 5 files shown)


source/fundamentals/builders/projections.txt
```diff
@@ -11,3 +11,169 @@ Projections Builders
...
 
 .. _projections-builders:
+
+Overview
+--------
+
+MongoDB supports **field projection**, specifying which fields to include and exclude when returning results from a
+query. Projection in MongoDB follows some basic rules:
+
+- the :guilabel:`_id` field is **always** included and must explicitly be excluded
```
MongoCaleb, line 21: Since these bullets aren't completing a sentence, I think they should be capitalized.

source/fundamentals/builders/projections.txt
```diff
@@ -11,3 +11,169 @@ Projections Builders
...
+- specifying a field for inclusion implicitly excludes all other fields **except** the :guilabel:`_id` field
+- specifying a field for exclusion removes only that field in a query result
+
+Find more information :manual:`here </tutorial/project-fields-from-query-results/>`.
```
MongoCaleb, line 25: For more information on which part? Could you say something like "For more information on selecting projection fields from query results, see ``?

source/fundamentals/builders/projections.txt
```diff
@@ -11,3 +11,169 @@ Projections Builders
...
+   ``excludeId`` is a convenience method and is shorthand for:
+
+   .. code-block:: java
+
+      exclude("_id")
+
+Project an Array Element Match
+------------------------------
+
+Use the ``elemMatch`` methods to specify an array projection.
```
MongoCaleb, line 97: This section is a bit confusing. ``elemMatch(String, Bson)`` returns the first element of an array that matches the Bson, but ``elemMatch(String)`` returns the first element that matches what? I think the tip below is intended to clarify this, and perhaps more seasoned java folks will understand, but I find this section needing a bit more clarity.

source/fundamentals/builders/projections.txt
```diff
@@ -11,3 +11,169 @@ Projections Builders
...
+---------------------
+
+Use the ``fields`` method to combine multiple projections.
+
+The following example includes the :guilabel:`quantity` and :guilabel:`totalAmount` fields and excludes the
+:guilabel:`_id` field:
+
+.. code-block:: java
+
+   fields(include("quantity", "totalAmount"), excludeId())
```
MongoCaleb, line 178: I was expecting, at the end of the page, a more complete code example showing how to put this in action. Does that exist elsewhere?

source/fundamentals/builders/projections.txt
terakilobyte: Thanks. Like I said, I wasn't happy with it. I think the new version is a lot more clear. Thoughts?

source/fundamentals/builders/projections.txt
terakilobyte: There are more complex examples in another PR for indexes (which shows filters, sorts, and projections). Trying to keep them lightweight for now, we can add denser examples should users request it.

ccho-mongodb (commented):
Just a few initial comments:

It could be helpful to explain how to include a projection built with a Builder in a query.

Let's use guilabel only for identifying sections of a UI rather than for field names, unless we're referring to a web/app view of documents like you might have in Compass or Realm UI. I think you've used it in cases in which we would just use the preformatted double-backtick syntax.

Let's use `()` after method names to identify them as such which we've been doing for Node.js and Java driver docs.

As Caleb mentioned already, it would be great to see output to see how each projection method alters the returned data. Maybe we can adopt a similar structure to what we used in Node.js for fundamentals articles in which we provide a sample data set and then show how we can alter the returned document with each method. See Alek's Sorts Builder PR for an example: https://github.com/mongodb/docs-java/pull/59/files

MongoCaleb (approved):
LGTM, though I still think more complete examples -- or a link to them --  would be very helpful.

terakilobyte:
Ready for review @ccho-mongodb

source/fundamentals/builders/projections.txt
```diff
@@ -11,3 +11,320 @@ Projections Builders
...
 
 .. _projections-builders:
+
+Overview
+--------
+
+MongoDB supports **field projection**, specifying which fields to include and exclude when returning results from a
+query. Projection in MongoDB follows some basic rules:
+
+- The ``_id`` field is **always** included and must explicitly be excluded
```
MongoCaleb, line 21: nit:
The ``_id`` field is **always** included unless explicitly be excluded

source/fundamentals/builders/projections.txt
```diff
@@ -11,3 +11,320 @@ Projections Builders
...
+
+The following example includes the ``year`` field and (implicitly) the ``_id`` field:
+
+.. literalinclude:: /includes/fundamentals/code-snippets/builders/Projections.java
+   :start-after: begin include single field
+   :end-before: end include single field
+   :language: java
+   :dedent:
+
+The output should resemble:
```
MongoCaleb, line 76: This may be 100% personal preference, but I always think the words preceding the colon should be a complete sentence, like: The following code shows the output from this projection:

¯\\_(ツ)_/¯

MongoCaleb: (As you already do preceding each main code snippet)

MongoCaleb (approved):
LGTM with two small suggestions

source/fundamentals/builders/projections.txt
```diff
@@ -11,3 +11,322 @@ Projections Builders
...
 
 .. _projections-builders:
+
+Overview
+--------
+
+MongoDB supports **field projection**, specifying which fields to include and exclude when returning results from a
+query. Projection in MongoDB follows some basic rules:
+
+- The ``_id`` field is **always** included unless explicitly
```
jyemin, line 21: `unless explicitly excluded`?

jyemin (approved):
LGTM!

PR #61, opened 2021-03-03
DOCSP-9583 builders filters page

kyuan-mongodb:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-9583

https://docs-mongodbcom-staging.corp.mongodb.com/cefb5dc/java/docsworker-xlarge/DOCSP-9583-BuildersFiltersPage/fundamentals/builders/filters/

### Self-Review Checklist

### DIFF
--- a/source/fundamentals/builders/filters.txt
+++ b/source/fundamentals/builders/filters.txt
@@ -41,7 +41,7 @@ The comparison operator methods include:
 - ``gte``: Matches values that are greater than or equal to a specified value.
 - ``lt``: Matches values that are less than a specified value.
 - ``lte``: Matches values that are less than or equal to a specified value.
-- ``ne``: Matches all values that are not equal to a specified value. 
+- ``ne``: Matches values that are not equal to a specified value. 
 - ``in``: Matches any of the values specified in an array.
 - ``nin``: Matches none of the values specified in an array.
 - ``empty``: Matches all the documents.
@@ -57,6 +57,7 @@ the value of the ``qty`` field equals ``20``:
 which will render as:
 
 .. code-block:: none
+   :copyable: false
    
    { 
     "qty" : 20 
@@ -84,6 +85,7 @@ because the predicate is empty:
 which will render as:
 
 .. code-block:: none
+   :copyable: false
    
    {}
 
@@ -93,14 +95,14 @@ Use ``logical`` operators to perform join or invert filters.
 
 The logical operator methods include:
 
-- ``and``: Joins filters with a logical AND and selects all documents that match the conditions of both filters.
-- ``or``: Joins filters with a logical OR and selects all documents that match the conditions of either filters.
+- ``and``: Joins filters with a logical AND and selects documents that match the conditions of both filters.
+- ``or``: Joins filters with a logical OR and selects documents that match the conditions of either filters.
 - ``not``: Inverts the effect of a query expression and selects documents that do not match the filter.
-- ``nor``: Joins filters with a logical NOR and selects all documents that fail to match both filters.
+- ``nor``: Joins filters with a logical NOR and selects documents that fail to match both filters.
 
-The following example creates a filter that selects all documents where
-the value of the ``qty`` field is greater than ``20`` and the value of
-the ``user`` field equals ``"jdoe"``:
+The following example creates a filter that selects documents where the
+value of the ``qty`` field is greater than ``20`` and the value of the
+``user`` field equals ``"jdoe"``:
 
 .. code-block:: java
 
@@ -109,14 +111,15 @@ the ``user`` field equals ``"jdoe"``:
 which will render as:
 
 .. code-block:: none
+   :copyable: false
    
    { 
     "qty" : { "$gt" : 20 }, 
     "user" : "jdoe" 
    }
 
-The following example creates a filter that selects all documents where
-the  ``price`` field value equals ``0.99`` or ``1.99``; and the ``sale``
+The following example creates a filter that selects documents where the
+``price`` field value equals ``0.99`` or ``1.99``; and the ``sale``
 field value is equal to ``true`` or the ``qty`` field value is less than
 ``20``:
 
@@ -128,6 +131,7 @@ field value is equal to ``true`` or the ``qty`` field value is less than
 which will render as:
 
 .. code-block:: none
+   :copyable: false
    
    { 
    "$and" : 
@@ -202,7 +206,7 @@ and diacritic-sensitive searches.
 
 .. code-block:: java
 
-   text("café", TextSearchOptions().diacriticSensitive(true))
+   text("coffee", TextSearchOptions().diacriticSensitive(true))
 
 Bitwise
 -------
@@ -210,8 +214,8 @@ Use ``bitwise`` operators to filter fields on a bit-level.
 
 The bitwise query operators include:
 
-- ``bitsAllSet``: Selects documents where the all the specified bits of a field are set (i.e. 1).
-- ``bitsAllClear``: Selects documents where the all the specified bits of a field are clear (i.e. 0).
+- ``bitsAllSet``: Selects documents where the specified bits of a field are set (i.e. 1).
+- ``bitsAllClear``: Selects documents where the specified bits of a field are clear (i.e. 0).
 - ``bitsAnySet``: Selects documents where at least one of the specified bits of a field are set (i.e. 1).
 - ``bitsAnyClear``: Selects documents where at least one of the specified bits of a field are clear (i.e. 0).
 
@@ -224,5 +228,35 @@ with bits set at positions of the corresponding bitmask ``50`` (i.e.
    bitsAllSet("bitField", 50)
 
 Geospatial
-~~~~~~~~~~
+----------
+Use ``geospatial`` operators to 
 
+The geospatial operator methods include:
+
+- ``geoWithin``: Selects documents containing a GeoJSON geometry field that falls within within a bounding GeoJSON geometry.
+- ``geoWithinBox``: Selects documents containing a grid coordinates field that exist within the specified box.
+- ``geoWithinPolygon``: Selects documents containing a grid coordinates field that exist within the specified polygon.
+- ``geoWithinCenter``: Selects documents containing a grid coordinates field that exist within the specified circle.
+- ``geoWithinCenterSphere``: Selects geometries containing a geospatial data field (GeoJSON or legacy coordinate pairs) that exist within the specified circle, using spherical geometry.
+- ``geoIntersects``: Selects geometries that intersect with a GeoJSON geometry. The 2dsphere index supports $geoIntersects.
+- ``near``: Returns geospatial objects in proximity to a point. Requires a geospatial index. The 2dsphere and 2d indexes support $near.
+- ``nearSphere``: Returns geospatial objects in proximity to a point on a sphere. Requires a geospatial index. The 2dsphere and 2d indexes support $nearSphere.
+
+The following example creates a filter that selects documents where the
+``geo`` field contains a GeoJSON Geometry object that falls within the
+given polygon:
+
+.. code-block:: java
+
+   val polygon: Polygon = Polygon(Seq(Position(0, 0), Position(4, 0),
+                                       Position(4, 4), Position(0, 4),
+                                       Position(0, 0)))
+   geoWithin("geo", polygon)
+
+The following example creates a filter that selects documents where the
+``geo`` field contains a GeoJSON Geometry object that intersects the
+given Point:
+
+.. code-block:: java
+
+   geoIntersects("geo", Point(Position(4, 0)))
--- a/source/fundamentals/builders/filters.txt
+++ b/source/fundamentals/builders/filters.txt
@@ -56,10 +56,10 @@ the value of the :guilabel:`qty` field equals :guilabel:`20`:
 
 which will render as:
 
-..code-block:: none
+.. code-block:: none
    { 
     "qty" : 20 
-    }
+   }
 
 .. note::
 
--- a/source/fundamentals/builders/filters.txt
+++ b/source/fundamentals/builders/filters.txt
@@ -60,15 +60,6 @@ type, which you can pass to any method that expects a query filter.
 
    The examples below assume this static import.
 
-The following sections show you how to use the ``Filters`` class to sort
-your queries using the following setup:
-
-.. literalinclude:: /includes/fundamentals/code-snippets/Filters.java
-   :language: java
-   :dedent:
-   :start-after: begin imports
-   :end-before: end imports 
-
 The Filter examples in this guide use the following sample collections:
 
 Collection: ``paint_purchases``
--- a/source/fundamentals/builders/filters.txt
+++ b/source/fundamentals/builders/filters.txt
@@ -156,7 +156,7 @@ both ``"ssl"`` and ``"security"``:
 
 Elements
 --------
-Use ``elements`` operators to select documents.
+Use ``elements`` operators to filter on a field.
 
 The elements operator methods include:
 
@@ -175,8 +175,34 @@ its value does not equal ``5`` or ``15``:
    and(exists("qty"), nin("qty", 5, 15))
 
 Evaluation
-~~~~~~~~~~
+----------
+Use ``evaluation`` to match documemts with an expression.
 
+The evaluation operator methods include:
+
+- ``mod``: Performs a modulo operation on the value of a field and selects documents with a specified result.
+- ``regex``: Selects documents where values match a specified regular expression.
+- ``text``: Selects documemts matching a full-text search expression.
+- ``where``: Matches documents that satisfy a JavaScript expression.
+
+The following example assumes a collection that has a text index in the
+field ``abstract``. It selects documents that have a ``abstract`` field
+containing the term ``coffee``:
+
+.. code-block:: java
+
+   text("coffee")
+
+Available with MongoDB 3.2, a version 3 text index allows case-sensitive
+and diacritic-sensitive searches.
+
+.. code-block:: java
+
+   text("coffee", TextSearchOptions().caseSensitive(true))
+
+.. code-block:: java
+
+   text("café", TextSearchOptions().diacriticSensitive(true))
 
 Bitwise
 ~~~~~~~
--- a/source/fundamentals/builders/filters.txt
+++ b/source/fundamentals/builders/filters.txt
@@ -27,7 +27,7 @@ Filtrs are used in the following ways:
 * :ref:`Geospatial <geospatial>`
 
 Filters are the operations MongoDB uses to limit your results to what
-you want to see. Some examples of data passed through filters are: 
+you want to see. Some examples of results from filters are: 
 
 - Items that cost $0 to $25
 - A hotel with ammenities you want
@@ -49,27 +49,46 @@ type, which you can pass to any method that expects a query filter.
    The examples below assume this static import.
 
 The following examples show you how to use the static factory methods
-provided by the ``Filteres`` class to sort your queries. The examples
-use a sample collection ``filter_example`` that contains the following
+provided by the ``Filters`` class to sort your queries. 
+
+The examples use these sample collections that contains the following
 documents: 
+
+``paint_purchaces``
 
 .. code-block:: json
 
-   {"_id": 1, "color": "red", "qty": 5}
-   {"_id": 2, "color": "purple", "qty": 10}
-   {"_id": 3, "color": "blue", "qty": 8}
-   {"_id": 4, "color": "white", "qty": 6}
-   {"_id": 5, "color": "yellow", "qty": 11}
-   {"_id": 6, "color": "pink", "qty": 5}
-   {"_id": 7, "color": "green", "qty": 8}
-   {"_id": 8, "color": "orange", "qty": 7}
-   
+   {"_id": 1, "color": "red", "qty": 5, "vendor": ["A"]}
+   {"_id": 2, "color": "purple", "qty": 10, "vendor": ["C", "D"]}
+   {"_id": 3, "color": "blue", "qty": 8, "vendor": ["B", "A"]}
+   {"_id": 4, "color": "white", "qty": 6, "vendor": ["D"]}
+   {"_id": 5, "color": "yellow", "qty": 11, "vendor": ["A", "B"]}
+   {"_id": 6, "color": "pink", "qty": 5, "vendor": ["C"]}
+   {"_id": 7, "color": "green", "qty": 8,"vendor": ["B", "C"]}
+   {"_id": 8, "color": "orange", "qty": 7, "vendor": ["A", "D"]}
+
+``binary_numbers``
+
+.. code-block:: json
+
+   { _id: 9, a: 54, binaryValue: "00110110" }
+   { _id: 10, a: 20, binaryValue: "00010100" }
+   { _id: 11, a: 68, binaryValue: "1000100" }
+   { _id: 12, a: 102, binaryValue: "01100110" }
+
+``geo_point``
+
+.. code-block:: json
+
+   { _id: 13, type: "Point", coordinates: [ 2,2 ] }
+   { _id: 14, type: "Point", coordinates: [ 5,6 ] }
+   { _id: 15, type: "Point", coordinates: [ 1,3 ] }
+   { _id: 16, type: "Point", coordinates: [ 4,7 ] }
+
 .. _comparison:
 
 Comparison 
 ----------
-A comparison filter returns documents matching your specified
-comparison. 
 
 The comparison operator methods include:
 
@@ -121,15 +140,15 @@ The following code snippet shows how to use a comparison filter:
 .. example::
 
    The following example creates a filter that selects all documents where
-   the value of the ``qty`` field equals ``5``:
+   the value of the ``qty`` field equals ``5`` in the paint_purchaces collection:
 
    .. code-block:: java
       
-      import static com.mongodb.client.model.Filters.
+      import static com.mongodb.client.model.Filters.eq
 
       // <MongoCollection setup code here>
 
-      Bson equalComparison = Filters.eq("qty", 5);
+      Bson equalComparison = eq("qty", 5);
       List<Document> results = new ArrayList<>();
       collection.find(equalComparison).into(results);
       for (Document result : results) {
@@ -150,23 +169,26 @@ The following code snippet shows how to use a comparison filter:
    .. code-block:: json
       :copyable: false
    
-      {"_id": 1, "color": "red", "qty": 5}
-      {"_id": 6, "color": "pink", "qty": 5}
+      {"_id": 1, "color": "red", "qty": 5, "vendor": ["A"]}
+      {"_id": 6, "color": "pink", "qty": 5, "vendor": ["C"]}
 
 .. example::
 
    The following example creates a filter that selects all documents
-   because the predicate is empty: 
+   because the predicate is empty in the paint_purchaces collection: 
 
    .. code-block:: java
       
-      import static com.mongodb.client.model.Filters.
+      import static com.mongodb.client.model.Filters.empty
 
       // <MongoCollection setup code here>
 
-      Bson emptyComparison = Filters.empty();
-      collection.find(emptyComparison);
-   
+      Bson emptyComparison = empty();
+      List<Document> results = new ArrayList<>();
+      collection.find(emptyComparison).into(results);
+      for (Document result : results) {
+         System.out.println(result.toJson());
+      }
    The Bson renders as:
 
    .. code-block:: none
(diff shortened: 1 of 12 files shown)


source/fundamentals/builders/filters.txt
```diff
@@ -11,3 +11,684 @@ Filters Builders
...
+Comparison 
+----------
+
+The comparison operator methods include:
+
+.. list-table::
+   :header-rows: 1
+   :stub-columns: 1
+
+   * - Comparison Operator
```
ccho-mongodb, line 98: Let's specify that these are methods. What they are equivalent to are called comparison operators like "$eq"

source/fundamentals/builders/filters.txt
```diff
@@ -11,3 +11,684 @@ Filters Builders
...
+     - any of the values specified in an array.
+
+   * - ``nin()``
+     - none of the values specified in an array.
+
+   * - ``empty()``
+     - all the documents.
+
+
+The following code snippet shows how to use a comparison filter:
```
ccho-mongodb, line 129: Let's stay consistent with terminology. Prior to this, you've called them "comparison operator methods" and never "comparison filter", so that switch can be confusing to readers.

source/fundamentals/builders/filters.txt
```diff
@@ -11,3 +11,684 @@ Filters Builders
...
+
+.. code-block:: java
+   
+   import static com.mongodb.client.model.Filters.*;
+
+   // <MongoCollection setup code here>
+
+   collection.find(<comparison filter>);
```
ccho-mongodb, line 137: Looking at this now, I'm not sure this code snippet makes much sense since you show it's usage in the example. This only shows how to use it for `find()`, but you can actually pass it to many other different types of CRUD operations.

source/fundamentals/builders/filters.txt
```diff
@@ -11,3 +11,684 @@ Filters Builders
...
+
+.. example::
+
+   The following example creates a filter that selects all documents where
+   the value of the ``qty`` field equals ``5`` in the paint_purchaces collection:
+
+   .. code-block:: java
+      
+      import static com.mongodb.client.model.Filters.eq;
```
ccho-mongodb, line 146: Earlier, you mentioned that "The examples below assume this [wildcard] static import." so I think this can be confusing. If you remove the sentence about assuming they use the static import, specifying the exact import makes sense. Alternatively, you could consider removing this line but then it can be more confusing to determine which methods come from the wildcard import.

source/fundamentals/builders/filters.txt
```diff
@@ -11,3 +11,684 @@ Filters Builders
...
+         System.out.println(result.toJson());
+      }
+   
+   The Bson renders as:
+
+   .. code-block:: none
+      :copyable: false
+   
+      {
+         "qty": 20
```
ccho-mongodb, line 163: Make sure your example code matches what you write it is equivalent to.

source/fundamentals/builders/filters.txt
```diff
@@ -11,3 +11,684 @@ Filters Builders
...
+   .. code-block:: json
+      :copyable: false
+   
+      {"_id": 1, "color": "red", "qty": 5, "vendor": ["A"]}
+      {"_id": 6, "color": "pink", "qty": 5, "vendor": ["C"]}
+
+.. example::
+
+   The following example creates a filter that selects all documents in
+   the paint_purchaces collection because the predicate is empty:
```
ccho-mongodb, line 177: I think this is missing some information.  You create a Bson filter, but there's more than that.

source/fundamentals/builders/filters.txt
```diff
@@ -11,3 +11,684 @@ Filters Builders
...
+
+      // <MongoCollection setup code here>
+
+      Bson emptyComparison = empty();
+      List<Document> results = new ArrayList<>();
+      collection.find(emptyComparison).into(results);
+      for (Document result : results) {
+         System.out.println(result.toJson());
+      }
+   The Bson renders as:
```
ccho-mongodb, line 191: Nitpick: although it's not required, let's add an extra linebreak after a code-block to make it easier for us to read the RST.

source/fundamentals/builders/filters.txt
```diff
@@ -11,3 +11,684 @@ Filters Builders
...
+      Bson equalComparison = eq("qty", 5);
+      List<Document> results = new ArrayList<>();
+      collection.find(equalComparison).into(results);
+      for (Document result : results) {
+         System.out.println(result.toJson());
+      }
+   
+   The Bson renders as:
+
+   .. code-block:: none
```
ccho-mongodb, line 159: Nitpick: the query operator is expressed in extended json, so we should be ok using the json code highlighting.

source/fundamentals/builders/filters.txt
```diff
@@ -11,3 +11,684 @@ Filters Builders
...
+      :copyable: false
+   
+      {
```
ccho-mongodb, line 162: Since what we're showing is a single, short example, could we make the equivalent operator a single line rather than spacing them out over multiple lines?

source/fundamentals/builders/filters.txt
```diff
@@ -11,3 +11,684 @@ Filters Builders
...
+
+   The output of the above code snippet should look something like this:
+
+   .. code-block:: json
+      :copyable: false
+   
+      {"_id": 1, "color": "red", "qty": 5, "vendor": ["A"]}
+      {"_id": 6, "color": "pink", "qty": 5, "vendor": ["C"]}
+
+.. example::
```
ccho-mongodb, line 174: Is there a specific reason why you chose "empty" or can we choose something else? Since no filter is equivalent to the empty operator, and it's likely that no filter is used more often, I would use this space to demonstrate something else that might be used more often.

source/fundamentals/builders/filters.txt
```diff
@@ -11,3 +11,684 @@ Filters Builders
...
+
+.. list-table::
+   :header-rows: 1
+   :stub-columns: 1
+
+   * - Logical Operator
+     - 
+      
+   * - ``and()``
+     - Joins filters with a logical AND and selects documents that match the conditions of both filters.
```
ccho-mongodb, line 227: Nitpick: we could use monospace for the logical operators to identify them as terms rather than grammatical conjunctions.

source/fundamentals/builders/filters.txt
```diff
@@ -11,3 +11,684 @@ Filters Builders
...
+
+   * - ``or()``
+     - Joins filters with a logical OR and selects documents that match the conditions of either filters.
+      
+   * - ``not()``
+     - Inverts the effect of a query expression and selects documents that do not match the filter.
```
ccho-mongodb, line 233: Inverts might not be the most clear term to use. I think you can convey the meaning of the sentence with just the second half: "Selects the documents that do not match the filter".

source/fundamentals/builders/filters.txt
```diff
@@ -11,3 +11,684 @@ Filters Builders
...
+
+   // <MongoCollection setup code here>
+
+   collection.find(<logical filter>);
+
+.. example::
+
+   The following example creates a filter that selects documents where
+   the value of the ``qty`` field is greater than ``8`` or the value
+   of the ``color`` field equals ``"pink"`` in the paint_purchaces collection:
```
ccho-mongodb, line 252: For values, let's try to stick with using quotes and no monospace. E.g. "8" and "pink".
Also, remember to keep monospace formatting on collection names.

source/fundamentals/builders/filters.txt
```diff
@@ -11,3 +11,684 @@ Filters Builders
...
+
+   * - ``regex()``
+     - Selects documents where values match a specified regular expression.
+      
+   * - ``text()``
+     - Selects documemts matching a full-text search expression.
+     
+   * - ``where()``
+     - Matches documents that satisfy a JavaScript expression.
+
```
ccho-mongodb, line 467: I noticed the first two mention "specified" -- are these different, or should they also use "specified" for consistency?

source/fundamentals/builders/filters.txt
```diff
@@ -11,3 +11,684 @@ Filters Builders
...
+.. list-table::
+   :header-rows: 1
+   :stub-columns: 1
+
+   * - Evaluation Operator
+     - 
+      
+   * - ``mod()``
+     - Performs a modulo operation on the value of a field and selects documents with a specified result.
```
ccho-mongodb, line 457: You use "matches" in the other descriptions but "selects" here, is there a difference, or should they be the same for consistency?

source/fundamentals/builders/filters.txt
```diff
@@ -11,3 +11,684 @@ Filters Builders
...
+An ``evaluation`` filter to match documemts with an expression.
+
+The evaluation operator methods include:
+
+.. list-table::
+   :header-rows: 1
+   :stub-columns: 1
+
+   * - Evaluation Operator
+     -
```
ccho-mongodb, line 454: We should describe what's in this column to clarify.

source/fundamentals/builders/filters.txt
```diff
@@ -11,3 +11,684 @@ Filters Builders
...
+
+   The output of the above code snippet should look something like this:
+   
+   .. code-block:: json
+      :copyable: false
+      
+      {"_id": 7, "color": "green", "qty": 8,"vendor": ["B", "C"]}
+
+   Available with MongoDB 3.2, a version 3 text index allows case-sensitive
+   and diacritic-sensitive searches.
```
ccho-mongodb, line 513: Let's omit any information about features before MongoDB server v3.6 since those are unsupported now. You could also describe what you're about to show and what a diacritic-sensitive search is.

source/fundamentals/builders/filters.txt
```diff
@@ -11,3 +11,684 @@ Filters Builders
...
+-------
+
+The bitwise query operators include:
+
+.. list-table::
+   :header-rows: 1
+   :stub-columns: 1
+
+   * - Logical Operator
+     - Selects documents where
```
ccho-mongodb, line 535: Do you have any objections to using "Matches" here which might be more consistent? I think we should try to use similar terminology when referring to the same thing when possible. If you changed this to "matches", the descriptions could read something like this:
"when the specified bits of a field are set (i.e. 1)."

source/fundamentals/builders/filters.txt
```diff
@@ -11,3 +11,684 @@ Filters Builders
...
+   with bits set at positions of the corresponding bitmask 50 (i.e.
+   00110010) in the binary_numbers collection: 
+
+   .. code-block:: java
+      
+      import static com.mongodb.client.model.Filters.bitsAllSet;
+
+      // <MongoCollection setup code here>
+
+      Bson bitsComparison = bitsAllSet(1, 5);
```
ccho-mongodb, line 571: I think there might be some confusion here in what parameters you provide this method. For example, how do we know this is for field "a"? And isn't a bitmask of 5 equivalent to 0x101 which doesn't match the results in the example? 

Could you check the API documentation and correct it? Also, it could be useful to mention something about the parameters in the descriptions in the table above.

source/fundamentals/builders/filters.txt
```diff
@@ -11,3 +11,684 @@ Filters Builders
...
+
+.. list-table::
+   :header-rows: 1
+   :stub-columns: 1
+
+   * - Comparison Operator
+     - Matches
+      
+   * - ``geoWithin()``
+     - Selects documents containing a GeoJSON geometry field that falls within within a bounding GeoJSON geometry.
```
ccho-mongodb, line 613: Since we use "matches" in the column header, this reads as "Matches selects documents..." Could you update this for correctness and consistency?

source/fundamentals/builders/filters.txt
```diff
@@ -11,3 +11,684 @@ Filters Builders
...
+      :copyable: false
+      
+      { _id: 9, a: 54, binaryValue: "00110110" }
+      { _id: 12, a: 102, binaryValue: "01100110" }
+
+.. _geospatial:
+
+Geospatial
+----------
+
```
ccho-mongodb, line 602: For each of these sections, I think we had discussed adding a brief description (one or two sentences) as well as links to the sections that don't exist yet. For the latter, could you pretend a link exists and substitute the link directive with "(TODO: link to CRUD geospatial section)"? Eventually, maybe we can move the tables you've created here to those sections and cross link between the two to reduce the duplicated content.

source/fundamentals/builders/filters.txt
```diff
@@ -11,3 +11,684 @@ Filters Builders
...
+   
+   import static com.mongodb.client.model.Filters.*;
+
+   // <MongoCollection setup code here>
+
+   collection.find(<geospatial filter>);
+
+.. example::
+
+   TThe following example creates a filter that selects documents where the
```
ccho-mongodb, line 648:
```suggestion
   The following example creates a filter that matches documents in which the
```

source/fundamentals/builders/filters.txt
```diff
@@ -11,3 +11,684 @@ Filters Builders
...
+         }
+      }
+
+   The output of the above code snippet should look something like this:
+   
+   .. code-block:: json
+      :copyable: false
+
+      { _id: 13, type: "Point", coordinates: [ 2,2 ] }
+      { _id: 15, type: "Point", coordinates: [ 1,3 ] }
```
ccho-mongodb, line 694: I don't know the answer offhand, but would these be formatted with no spaces between numbers in the array?

ccho-mongodb (changes requested):
Thanks for addressing many of the points we discussed during our pair session. I think in this revision, we should focus on refining some of the following:
- consistency in terms, formatting, etc.
- conciseness
- proper level of explanation
Feel free to reach out with any questions!

source/fundamentals/builders/filters.txt
kyuan-mongodb: The jira ticket requests for this method to be included

source/fundamentals/builders/filters.txt
kyuan-mongodb: should I add a column in each table for the parameters of each method?

source/fundamentals/builders/filters.txt
ccho-mongodb: Can we show an additional example, perhaps one of the filter builders that uses one or more parameters?

I don't mind keeping this example as well; the linked ticket wasn't specific as to whether to include an example or just the information you included in the table.

source/fundamentals/builders/filters.txt
kyuan-mongodb: I ended up taking them out since https://docs.mongodb.com/manual/reference/operator/query/ doesn't have descriptions for them either.  Would you prefer if I do add a brief description?

source/fundamentals/builders/filters.txt
ccho-mongodb: It might be useful at some point to link these to API documentation, but can you explain examples that might not be self-explanatory like this one? From this API documentation page, unless I'm missing something, I'm not sure this will compile since the first argument should be a String:
https://mongodb.github.io/mongo-java-driver/4.2/apidocs/mongodb-driver-core/com/mongodb/client/model/Filters.html#bitsAllSet(java.lang.String,long)

source/fundamentals/builders/filters.txt
ccho-mongodb: Sorry, not sure what you mean. Maybe it would be easier to understand if I could see your changes (make sure to push them to your branch). If you're referring to not being able to find descriptions for geospatial filters, I see descriptions on both the page you linked as well as https://docs.mongodb.com/manual/geospatial-queries/

source/fundamentals/builders/filters.txt
```diff
@@ -11,3 +11,560 @@ Filters Builders
...
+   :stub-columns: 1
+
+   * - Geospatial Method 
+     - Matches
+      
+   * - ``geoWithin()``
+     - documents containing a GeoJSON geometry field that falls within within a bounding GeoJSON geometry.
+
+   * - ``geoWithinBox()``
+     - documents containing a grid coordinates field that exist within the specified box.
```
ccho-mongodb, line 517: I'm not familiar with the "grid coordinates field" term -- can you point me to where they're referred to as such? Or did you mean just "coordinates", or a specific type like a Point?

Also, I wouldn't say the "field" exists within the specified box but rather the "value" exists. Comment applies to all instances.

source/fundamentals/builders/filters.txt
```diff
@@ -11,3 +11,560 @@ Filters Builders
...
+
+   * - ``near()``
+     - geospatial objects in proximity to a point. Requires a geospatial index. The 2dsphere and 2d indexes support $near.
+
+   * - ``nearSphere()``
+     - geospatial objects in proximity to a point on a sphere. Requires a geospatial index. The 2dsphere and 2d indexes support $nearSphere.
+
+
+The following example creates a filter that matches documents in which
+the ``point`` field contains a GeoJSON Geometry object that falls within
```
ccho-mongodb, line 539: Question: should "Geometry" be capitalized here (I'm not sure)? And is geometry redundant since you specify GeoJSON, or is there a geometry type of GeoJSON?

source/fundamentals/builders/filters.txt
```diff
@@ -11,3 +11,560 @@ Filters Builders
...
+
+.. list-table::
+   :header-rows: 1
+   :stub-columns: 1
+
+   * - Bitwise Method
+     - Matches
+      
+   * - ``bitsAllSet()``
+     - documents when the specified bits of a field are set (i.e. 1).
```
ccho-mongodb, line 458: Let's use double quotes around values:

```suggestion
     - documents when the specified bits of a field are set (i.e. "1").
```

source/fundamentals/builders/filters.txt
ccho-mongodb: Not sure if you noticed this comment, so bumping it just in case ^

ccho-mongodb (changes requested):
Thanks for putting in all the work to create the test class and address the comments. I have more comments, including items that I missed previously. Please reach out if you have any questions.

source/fundamentals/builders/filters.txt
kyuan-mongodb: Just checked in the docs manual and the way it's used there is "GeoJSON geometry"

source/fundamentals/builders/filters.txt
```diff
@@ -11,3 +11,571 @@ Filters Builders
...
+--------
+In this guide, we show you how to use **builders** to specify
+**filters** for your queries.
+
+Builders are classes provided by the MongoDB Java driver that allow you
+to use the builder pattern to interact with your MongoDB instance. The
+builders provided by the MongoDB Java driver can improve the readability
+of your code.  See the :doc:`Builders overview page
+</fundamentals/builders/>` for a list of Builders available in the Java
+driver.
```
ccho-mongodb, line 25: Let's choose to use either the capitalized or lowercase form of "builders" consistently. 

For the link to "Builders overview page" you might consider using title case, or whatever the text is in the title of that page.

source/fundamentals/builders/filters.txt
```diff
@@ -11,3 +11,571 @@ Filters Builders
...
+   The examples below assume this static import.
+
+The following sections show you how to use the ``Filters`` class to sort
+your queries using the following setup:
+
+.. literalinclude:: /includes/fundamentals/code-snippets/Filters.java
+   :language: java
+   :dedent:
+   :start-after: begin imports
+   :end-before: end imports
```
ccho-mongodb, line 70: I'm not sure what this section is supposed to show. A number of these imports aren't related to the filters. And we'd only want to show which imports we're using for a specific filter which you've covered by mentioning that they use the static import in the tip.

source/fundamentals/builders/filters.txt
```diff
@@ -11,3 +11,571 @@ Filters Builders
...
+   { _id: 14, "coordinates": { type: "Point", coordinates: [5.0, 6.0] } }
+   { _id: 15, "coordinates": { type: "Point", coordinates: [1.0, 3.0] } }
+   { _id: 16, "coordinates": { type: "Point", coordinates: [4.0, 7.0] } }
+
+.. _comparison:
+
+Comparison 
+----------
+
+The comparison operator compares the value in a document to a specified value.
```
ccho-mongodb, line 110: I think there are two different ways to present each of the operator types that might be better than this. In the current form, it sounds like there is a single "comparison operator", but it is actually a category of several operators which you show in the table below.
One way to resolve this is to refer to the operators in plural form. E.g. "The comparison operators compare..."
Or to identify it as a category. E.g. "The comparison filters include all operators that compare ..."

source/fundamentals/builders/filters.txt
```diff
@@ -11,3 +11,571 @@ Filters Builders
...
+   :end-before: end emptyComparison
+
+The JSON output renders as the following:
+
+.. code-block:: json
+   :copyable: false
+   
+   { }
+
+The output of the above code snippet should look something like this:
```
ccho-mongodb, line 215: I think the user would get the idea if you mentioned here "The output of the code snippet above consists of all the documents in the collection." and then showed just the first three followed by an ellipsis. That way, we don't need to show all 8 lines to try to explain it.

source/fundamentals/builders/filters.txt
```diff
@@ -11,3 +11,571 @@ Filters Builders
...
+   {"_id": 5, "color": "yellow", "qty": 11, "vendor": ["A", "B"]}
+   {"_id": 6, "color": "pink", "qty": 5, "vendor": ["C"]}
+   {"_id": 7, "color": "green", "qty": 8,"vendor": ["B", "C"]}
+   {"_id": 8, "color": "orange", "qty": 7, "vendor": ["A", "D"]}
+
+.. _logical:
+
+Logical
+-------
+The logical operator looks at the conditions of the specified filters.
```
ccho-mongodb, line 233: Can you elaborate on this? I'm not sure if I could understand how they function from the description: "looks at the conditions".

source/fundamentals/builders/filters.txt
```diff
@@ -11,3 +11,571 @@ Filters Builders
...
+.. code-block:: json
+   :copyable: false
+   
+   { 
+      "$or" :
+      [
+         "qty" : { "$gt" : 8 }, 
+         "color" : "pink"
+      ]       
+   }
```
ccho-mongodb, line 277: Nitpick: Unless there's a reason not to, can you make this more compact? Maybe the curly braces could be on the same line and the square braces could be the trailing character on the previous lines.

source/fundamentals/builders/filters.txt
```diff
@@ -11,3 +11,571 @@ Filters Builders
...
+   * - :java-core-api:`bitsAllClear() <com/mongodb/client/model/Filters.html#bitsAllClear(java.lang.String,long)>`
+     - documents when the specified bits of a field are clear (i.e. "0").
+      
+   * - :java-core-api:`bitsAnySet() <com/mongodb/client/model/Filters.html#bitsAnySet(java.lang.String,long)>`
+     - documents when at least one of the specified bits of a field are set (i.e. "1").
+     
+   * - :java-core-api:`bitsAnyClear() <com/mongodb/client/model/Filters.html#bitsAnyClear(java.lang.String,long)>`
+     - documents when at least one of the specified bits of a field are clear (i.e. "0").
+
+The following example matches documents that have a ``bitField ``field
```
ccho-mongodb, line 483: The formatting breaks here because of the lack of a space after the closing "``"

source/fundamentals/builders/filters.txt
```diff
@@ -11,3 +11,571 @@ Filters Builders
...
+with bits set at positions of the corresponding bitmask "34" (i.e.
+00100010) in the ``binary_numbers`` collection:
```
ccho-mongodb, line 485: Should this value be in quotes too?

source/fundamentals/builders/filters.txt
```diff
@@ -11,3 +11,571 @@ Filters Builders
...
+      
+   { _id: 9, a: 54, binaryValue: "00110110" }
+   { _id: 12, a: 102, binaryValue: "01100110" }
+
+.. _geospatial:
+
+Geospatial
+----------
+
+A geospatial operator evaluates a specified coordinate to a shape or location.
```
ccho-mongodb, line 513: Not sure I understand "evaluates a specified coordinate to a shape or location". Maybe you meant something like "evaluates a specified coordinate and its relation to a shape or location"?

source/fundamentals/builders/filters.txt
```diff
@@ -11,3 +11,571 @@ Filters Builders
...
+         $geoWithin: {
+            $geometry: {
+               type: "Polygon",
+               coordinates: [ [[ 0, 0], [4 , 0], [4, 4], [0, 4], [0 , 0] ] ]
+            }
+         }
+      }
+   }
+
+The output of the above code snippet should look something like this:
```
ccho-mongodb, line 575: Nitpick: I was just thinking that maybe a more clear way to identify what this is could be:
"The following shows the output of the query using the operator specified above:"

ccho-mongodb (changes requested):
Looks good, just a few small items to fix first!

source/fundamentals/builders/filters.txt
```diff
@@ -11,3 +11,549 @@ Filters Builders
...
+Overview 
+--------
+In this guide, we show you how to use **builders** to specify
+**filters** for your queries.
+
+Builders are classes provided by the MongoDB Java driver that allow you
+to use the builder pattern to interact with your MongoDB instance. The
+builders provided by the MongoDB Java driver can improve the readability
+of your code.  See the :doc:`Builders
+</fundamentals/builders/>` page for a list of Builders available in the
```
ccho-mongodb, line 24: Nitpick: no change required, but I like to keep links (whether external links or directive-based ones) on the same line even if it goes beyond the 80 char width. In this case, I don't think it would even go beyond that width if you kept the Builders link on line 23.

source/fundamentals/builders/filters.txt
```diff
@@ -11,3 +11,549 @@ Filters Builders
...
+
+   The examples below assume this static import.
+
+The Filter examples in this guide use the following sample collections:
+
+Collection: ``paint_purchases``
+
+.. code-block:: json
+
+   {"_id": 1, "color": "red", "qty": 5, "vendor": ["A"]}
```
ccho-mongodb, line 69: Nitpick: I think I mentioned this in a comment in an earlier review of this PR about consistency of spacing in JSON. Sometimes you have spaces padding the "{" and the first field and spaces padding the "}" field. We should stay consistent.

In case it's helpful, here's a doc that provides some info on how our team decided to do javascript/json formatting:
https://docs.google.com/spreadsheets/d/1vxDasa0S2yrpXfTJkH8VooYBIasiHdOIotQOx_b8xfc/edit#gid=0
I think there's another source for this info, but can't locate it at the moment.

Applies to all the json instances.

source/fundamentals/builders/filters.txt
```diff
@@ -11,3 +11,549 @@ Filters Builders
...
+
+.. code-block:: json
+   :copyable: false
+   
+   {"_id": 1, "color": "red", "qty": 5, "vendor": ["A"]}
+   {"_id": 6, "color": "pink", "qty": 5, "vendor": ["C"]}
+
+The following example creates a filter that matches all documents where
+the value of the ``qty`` field is greater than or equal to "8" in the
+``paint_purchases`` collection:
```
ccho-mongodb, line 166: Nitpick: to keep the result set more compact, we could use 10 instead of 8. Then it would only return two results but still demonstrate values equal and greater, and save a couple lines.

source/fundamentals/builders/filters.txt
```diff
@@ -11,3 +11,549 @@ Filters Builders
...
+   
+The JSON output renders as the following:
+
+.. code-block:: json
+   :copyable: false
+   
+   { "vendor" : { 
+         "$all" : { 
+               ["A", "D"] 
+   } } }
```
ccho-mongodb, line 318: Nitpick: I like the placement of the three closing curly braces on one line, but I think this should be indented to match the first curly brace it closes which would be in the same column as "$all".

source/fundamentals/builders/filters.txt
```diff
@@ -11,3 +11,549 @@ Filters Builders
...
+   :copyable: false
+      
+   {"_id": 8, "color": "orange", "qty": 7, "vendor": ["A", "D"]}
+
+.. _elements:
+
+Elements
+--------
+
+The elements operators evaluate a specified field.
```
ccho-mongodb, line 332: Nitpick: I wonder if this description could be more specific. I think "evaluate a specified field" could refer to almost any operator.

source/fundamentals/builders/filters.txt
```diff
@@ -11,3 +11,549 @@ Filters Builders
...
+   {"_id": 2, "color": "purple", "qty": 10, "vendor": ["C", "D"]}
+   {"_id": 6, "color": "pink", "qty": 5, "vendor": ["C"]}
+
+
+.. _bitwise:
+
+Bitwise
+-------
+
+The bitwise operators convert a number into it's binary value to
```
ccho-mongodb, line 436:
```suggestion
The bitwise operators convert a number into its binary value to
```

source/fundamentals/builders/filters.txt
```diff
@@ -11,3 +11,549 @@ Filters Builders
...
+     - documents containing a coordinates value that exist within the specified polygon.
+
+   * - :java-core-api:`geoWithinCenter() <com/mongodb/client/model/Filters.html#geoWithinCenter(java.lang.String,double,double,double)>`
+     - documents containing a coordinates value that exist within the specified circle.
+
+   * - :java-core-api:`geoWithinCenterSphere() <com/mongodb/client/model/Filters.html#geoWithinCenterSphere(java.lang.String,double,double,double)>`
+     - geometries containing a geospatial data value (GeoJSON or legacy coordinate pairs) that exist within the specified circle, using spherical geometry.
+
+   * - :java-core-api:`geoIntersects() <com/mongodb/client/model/Filters.html#geoIntersects(java.lang.String,org.bson.conversions.Bson)>`
+     - geometries that intersect with a GeoJSON geometry. The 2dsphere index supports $geoIntersects.
```
ccho-mongodb, line 518: Let's use monospace for operators like "$geoIntersects".

ccho-mongodb (approved):
Looks good! I left some comments and nitpicks that would be great to address before getting an external review.

source/fundamentals/builders/filters.txt
```diff
@@ -11,3 +11,542 @@ Filters Builders
...
+   
+The JSON output renders as the following:
+
+.. code-block:: json
+   :copyable: false
+   
+   { "vendor" : { 
+      "$all" : { 
+         ["A", "D"] 
+         } } }
```
ccho-mongodb, line 315: Nitpick: this one is probably short enough to fit on one line. You can also close all the curly braces in a row without a space, e.g. "}}}"

source/fundamentals/builders/filters.txt
```diff
@@ -11,3 +11,542 @@ Filters Builders
...
+The JSON output renders as the following:
+
+.. code-block:: json
+      :copyable: false
+   
+   { "point": {
+         $geoWithin: {
+            $geometry: {
+               type: "Polygon",
+               coordinates: [ [[ 0, 0], [4 , 0], [4, 4], [0, 4], [0 , 0] ] ]
```
ccho-mongodb, line 543:
```suggestion
               coordinates: [ [ [ 0, 0], [4 , 0], [4, 4], [0, 4], [0 , 0] ] ]
```

source/fundamentals/builders/filters.txt
```diff
@@ -11,3 +11,534 @@ Filters Builders
...
+The following example creates a filter that matches all documents where
+the value of the ``qty`` field equals "5" in the ``paint_purchases`` collection:
+
+.. literalinclude:: /includes/fundamentals/code-snippets/Filters.java
+   :language: java
+   :dedent:
+   :start-after: begin equalComparison
+   :end-before: end equalComparison
+   
+The JSON output renders as the following:
```
jyemin, line 147: I found this a bit confusing, since the code above doesn't actually render the filter as JSON.  Do you think we should show code that renders the filter?

source/fundamentals/builders/filters.txt
```diff
@@ -11,3 +11,534 @@ Filters Builders
...
+   :dedent:
+   :start-after: begin existsComparison
+   :end-before: end existsComparison
+
+The JSON output renders as the following:
+
+.. code-block:: json
+   :copyable: false
+   
+   { "qty" : [ {
```
jyemin, line 354: In the 4.0 driver it currently renders as:

    {"$and": [{"qty": {"$exists": true}}, {"qty": {"$nin": [5, 8]}}]}

You can see with this code:

        var filter = and(exists("qty"), nin("qty", 5, 8));
        var doc = filter.toBsonDocument(BsonDocument.class, MongoClientSettings.getDefaultCodecRegistry());
        System.out.println(doc.toJson());

But it's subject to change given bugs like https://jira.mongodb.org/browse/JAVA-4006.  So we may need to mention that.

source/fundamentals/builders/filters.txt
```diff
@@ -11,3 +11,534 @@ Filters Builders
...
+   :dedent:
+   :start-after: begin bitsComparison
+   :end-before: end bitsComparison
+
+The JSON output renders as the following:
+
+.. code-block:: json
+      :copyable: false
+   
+   {  "a": { "$bitsAllSet": [1,5] }}
```
jyemin, line 465: bitsAllSet("a", 34)

renders as

    {"a": {"$bitsAllSet": 34}}

source/fundamentals/builders/filters.txt
```diff
@@ -11,3 +11,534 @@ Filters Builders
...
+The geospatial operator methods include:
+
+.. list-table::
+   :header-rows: 1
+   :stub-columns: 1
+
+   * - Geospatial Method 
+     - Matches
+      
+   * - :java-core-api:`geoWithin() <com/mongodb/client/model/Filters.html#geoWithin(java.lang.String,org.bson.conversions.Bson)>`
```
jyemin, line 492: Even fully expanded, the first column is wrapping to two lines, which looks funny for method names.  Can that be fixed?

source/fundamentals/builders/filters.txt
```diff
@@ -11,3 +11,534 @@ Filters Builders
...
+
+   * - :java-core-api:`near() <com/mongodb/client/model/Filters.html#near(java.lang.String,org.bson.conversions.Bson,java.lang.Double,java.lang.Double)>`
+     - geospatial objects in proximity to a point. Requires a geospatial index. The 2dsphere and 2d indexes support ``$near``.
+
+   * - :java-core-api:`nearSphere() <com/mongodb/client/model/Filters.html#nearSphere(java.lang.String,org.bson.conversions.Bson,java.lang.Double,java.lang.Double)>`
+     - geospatial objects in proximity to a point on a sphere. Requires a geospatial index. The 2dsphere and 2d indexes support ``$nearSphere``.
+
+The following example creates a filter that matches documents in which
+the ``point`` field contains a GeoJSON geometry that falls within
+the given polygon in the ``geo_points`` collection:
```
jyemin, line 518: Could we link to the Javadoc for Polygon somewhere?

source/fundamentals/builders/filters.txt
```diff
@@ -11,3 +11,534 @@ Filters Builders
    :class: singlecol
 
 .. _filters-builders:
+
+Overview 
+--------
+In this guide, we show you how to use **builders** to specify
+**filters** for your queries.
```
jyemin, line 18: I think we should point out some of the places where filters are actually used, e.g.

* find
* match stage of aggregate
* findAndDelete/findAndUpdate/findAndReplace
* countDocuments
* distinct
* deleteOne/deleteMany
* updateOne/updateMany
* replaceOne

jyemin (changes requested):
Looks really nice.  Just a few comments/clarifications to address.

source/fundamentals/builders/filters.txt
kyuan-mongodb: I thought it would help clarify to the reader what the filter is doing. This was brought up in another PR for me too and I was told we can assume the user has a general understanding and won't need to see what it renders as.  I'll take these out in all sections.

jyemin: No, I like it and am not saying we should remove.  I was just wondering if we should show a sample of how one would show the rendering oneself.

jyemin: But if you all decide to remove it I'm ok with that too.

jyemin (approved):

PR #63, opened 2021-03-05
DOCSP-9587 builders updates

kyuan-mongodb:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-9587

https://docs-mongodbcom-staging.corp.mongodb.com/652a724/java/docsworker-xlarge/DOCSP-9587-BuildersUpdates/fundamentals/builders/updates/

### Self-Review Checklist

### DIFF
--- a/source/fundamentals/builders/updates.txt
+++ b/source/fundamentals/builders/updates.txt
@@ -69,7 +69,7 @@ The following example sets the value of the ``qty`` field to "11":
    :start-after: begin setUpdate
    :end-before: end setUpdate
 
-The example above updates the document to the following state:
+The example above updates the original document to the following state:
 
 .. code-block:: json
    :copyable: false
@@ -95,7 +95,7 @@ The following example deletes the ``qty`` field:
    :start-after: begin unsetUpdate
    :end-before: end unsetUpdate
 
-The example above updates the document to the following state:
+The example above updates the original document to the following state:
    
 .. code-block:: json
    :copyable: false
@@ -122,7 +122,7 @@ if an upsert resulted in the insert of a document:
    :start-after: begin setOnInsertUpdate
    :end-before: end setOnInsertUpdate
 
-The example above updates the document to the following state:
+The example above updates the original document to the following state:
    
 .. code-block:: json
    :copyable: false
@@ -134,6 +134,9 @@ The example above updates the document to the following state:
       "vendor": [ "A", "D", "M" ], 
       "lastModified" = Fri Mar 05 00:00:00 EST 2021
    }
+
+.. note::
+   If the document is not inserted, no change will occur.
 
 Increment
 ~~~~~~~~~
@@ -149,7 +152,7 @@ The following example increments the value of the ``qty`` field by
    :start-after: begin incUpdate
    :end-before: end incUpdate
 
-The example above updates the document to the following state:
+The example above updates the original document to the following state:
    
 .. code-block:: json
    :copyable: false
@@ -176,7 +179,7 @@ The following example multiplies the value of the ``qty`` field by
    :start-after: begin mulUpdate
    :end-before: end mulUpdate
 
-The example above updates the document to the following state:
+The example above updates the original document to the following state:
    
 .. code-block:: json
    :copyable: false
--- a/source/fundamentals/builders/updates.txt
+++ b/source/fundamentals/builders/updates.txt
@@ -219,8 +219,8 @@ The example above updates the document to the following state:
       
    {
       "_id": 1,
-      "color": "red", "
-      vendor": [ "A", "D", "M" ],
+      "color": "red", 
+      "vendor": [ "A", "D", "M" ],
       "lastModified": { "$date": "2021-03-05T05:00:00Z" }, 
       "quantity": 5
    }
@@ -245,7 +245,7 @@ The example above updates the document to the following state:
    {
       "_id": 1, 
       "color": "red", 
-      qty": 2, 
+      "qty": 2, 
       "vendor": [ "A", "D", "M" ], 
       "lastModified": { "$date": "2021-03-05T05:00:00Z" }
    }
@@ -303,7 +303,7 @@ The example above updates the document to the following state:
       "color": "red", 
       "qty": 5, 
       "vendor": [ "A", "D", "M" ],
-      "lastModified": {"$date": "2021-03-22T21:01:20.027Z"
+      "lastModified": { "$date": "2021-03-22T21:01:20.027Z" }
    }
 
 Current Timestamp
--- a/source/fundamentals/builders/updates.txt
+++ b/source/fundamentals/builders/updates.txt
@@ -15,7 +15,17 @@ Updates Builders
 Overview 
 --------
 In this guide, we show you how to use **builders** to specify
-**updates** for your queries:
+**updates** for your queries.
+
+Builders are classes provided by the MongoDB Java driver that allow you
+to use the builder pattern to interact with your MongoDB instance. The
+builders provided by the MongoDB Java driver can improve the readability
+of your code.  See the :doc:`Builders overview page
+</fundamentals/builders/>` for a list of Builders available in the Java
+driver. 
+
+This guide shows you how to use builders with examples of the following
+types of updates: 
 
 - :ref:`Field Updates <field_updates>`
 - :ref:`Array Updates <array_updates>`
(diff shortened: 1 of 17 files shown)


terakilobyte (changes requested):
This looks really good. I just have a few questions. One is in a review comment, and the other is:

When I read through the examples and how the document changes, the wording is
`The example above updates the document to the following state:`. Could we clarify this by adding something like `... updates the original document...` or making them build upon on another? That way the user doesn't have to keep the original document state in mind, or scroll up if they are incredulous.

source/fundamentals/builders/updates.txt
```diff
@@ -11,3 +11,501 @@ Updates Builders
...
+      "color": "red", 
+      "vendor": [ "A", "D", "M" ], 
+      "lastModified" = Fri Mar 05 00:00:00 EST 2021
+   }
+
+Set On Insert
+~~~~~~~~~~~~~
+Use the :java-core-api:`setOnInsert() <com/mongodb/client/model/Updates.html#setOnInsert(java.lang.String,TItem)>`
+method to assign the value of a field in an update operation on an
+insert of a document.
```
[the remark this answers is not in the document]
kyuan-mongodb: I actually tried it out in my playground, and if an update doesn't occur, nothing happens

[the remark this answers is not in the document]
kyuan-mongodb: I added a note about this on the page.  I also want to keep the examples to the original document in the case a user on the page is looking for a particular example.

terakilobyte (approved):
LGTM

source/fundamentals/builders/updates.txt
```diff
@@ -11,3 +11,505 @@ Updates Builders
...
+      "vendor": [ "A", "D", "M" ],
+      "lastModified" = Fri Mar 05 00:00:00 EST 2021
+   }
+
+Current Date
+~~~~~~~~~~~~
+Use the :java-core-api:`currentDate() <com/mongodb/client/model/Updates.html#currentDate(java.lang.String)>`
+method to assign the value of a field in an update operation to the
+current date as a :manual:`date </reference/bson-types/#document-bson-type-date>`
+or :manual:`timestamp <reference/bson-types/#document-bson-type-timestamp>`.
```
jyemin, line 280: To update it to a timestamp, you have to use `currentTimestamp` I think

source/fundamentals/builders/updates.txt
```diff
@@ -11,3 +11,505 @@ Updates Builders
...
+      "color": "red", 
+      "qty": 5, 
+      "vendor": [ "D" ],
+      "lastModified" = Fri Mar 05 00:00:00 EST 2021
+   }
+
+Pull
+~~~~
+Use the :java-core-api:`pull() <com/mongodb/client/model/Updates.html#pull(java.lang.String,TItem)>`
+method to remove all instances of a value from an existing array in
```
jyemin, line 436: Does it remove all or just the first?  As is, the description of this method is the same as the description for pullAll. (I think the Javadoc has the same confusion, but it would be good to clarify here and perhaps also fix the Javadoc)

jyemin (changes requested):

source/fundamentals/builders/updates.txt
```diff
@@ -11,3 +11,505 @@ Updates Builders
...
+The examples in this guide use the following document: 
+
+.. code-block:: json
+
+   {
+       "_id": 1, 
+       "color": "red", 
+       "qty": 5, 
+       "vendor": [ "A", "D", "M" ], 
+       "lastModified" = Fri Mar 05 00:00:00 EST 2021
```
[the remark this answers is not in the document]
kyuan-mongodb: That's what printed out when I performed a System.out.print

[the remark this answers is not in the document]
jyemin: Right.  But that uses the `toString()` method, which doesn't produce JSON.  If you use the `toJson()` method instead, you'll get something like: `{"$date": "2021-03-18T19:18:17.265Z"}` for any date value.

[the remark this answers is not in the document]
kyuan-mongodb: Correct, I just checked and removed my comment since I saw what you meant.

source/fundamentals/builders/updates.txt
kyuan-mongodb: https://docs.mongodb.com/manual/reference/operator/update/currentDate/

jyemin: Right, but the currentDate builder doesn't work that way: it only generates dates.  You have to use currentTimestamp to generate a timestamp.  Maybe not worth mentioning that though since you rarely need a timestamp.

source/fundamentals/builders/updates.txt
kyuan-mongodb: So the difference is that pull only removes one value, but pullAll can do multiple. I didn't properly convey it in this PR, so I'll be updating it

jyemin (commented):
Waiting for request for second review

kyuan-mongodb:
@jyemin This is ready for a re-review

source/fundamentals/builders/updates.txt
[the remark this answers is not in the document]
jyemin: Looks like you replaced it with the shell syntax.  I think it would be better to use extended JSON syntax instead.

source/fundamentals/builders/updates.txt
jyemin: What did you end up doing with this one?

source/fundamentals/builders/updates.txt
jyemin: What did you end up doing with this one?

source/includes/fundamentals/code-snippets/builders/Updates.java
```diff
@@ -0,0 +1,264 @@
...
+        database = mongoClient.getDatabase("builders");
+        collection = database.getCollection("updates");
+        // end declaration
+    }
+
+    public static void main(String[] args) {
+        Updates updates = new Updates();
+        updates.resetCollection(updates);
+
+        System.out.println("setUpdate:");
```
jyemin, line 43: Did you intend to leave the println statements in?

kyuan-mongodb: Yes, I forgot to include them before and it took me a while to figure out the order of the updates being performed.

source/fundamentals/builders/updates.txt
kyuan-mongodb: I fixed the wording for pull to be "of a value" and pullAll to be "of values" since that's how they differ.

source/fundamentals/builders/updates.txt
kyuan-mongodb: I left it as it is since this is what https://mongodb.github.io/mongo-java-driver/4.2/builders/updates/ has for currentDate.  Would you like it to be changed to currentTimestamp?

jyemin: Yeah, I think current docs are wrong.  currentDate helper gives you a date, currentTimestamp gives you a timestamp.

kyuan-mongodb: I just tested both methods.
currentTimestamp returns: lastModified=Timestamp{value=6942526372958961671, seconds=1616432884, inc=7}
currentDate returns: lastModified=Mon Mar 22 13:08:04 EDT 2021

I think currentDate performs as expected.  So to clarify, you want currentTimestamp to replace the currentDate section?

jyemin: How about two sections, one for each helper method?

jyemin (changes requested):

source/fundamentals/builders/updates.txt
```diff
@@ -138,7 +138,7 @@ The example above updates the original document to the following state:
       "color": "red", 
       "qty": 5, 
       "vendor": [ "A", "D", "M" ], 
-      "lastModified" = "2021-03-05T00:00:00.000Z"
+      "lastModified" = "lastModified": { "$date": "2021-03-05T05:00:00Z" }
```
jyemin, line 141: Nit: doubled-up `lastModified`?

jyemin (approved):
One apparent typo, but no need to re-review.

LGTM!

PR #64, opened 2021-03-12
DOCSP-9588 builders indexes

kyuan-mongodb:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-9588

https://docs-mongodbcom-staging.corp.mongodb.com/abd2db5/java/docsworker-xlarge/DOCSP-9588-BuildersIndexes/fundamentals/builders/indexes/

### Self-Review Checklist

### DIFF
--- a/source/fundamentals/builders/indexes.txt
+++ b/source/fundamentals/builders/indexes.txt
@@ -26,16 +26,15 @@ convenient methods for different types of indexes:
 - :ref:`Hashed Indexes <index-hashed-indexes>`
 - :ref:`Geospatial Indexes <geospatial-indexes>`
 
-In a textbook, the index provides the user a way to view pages a
-keyword they're looking for is. Likewise, an index in MongoDB groups
-data based on the field you provide and by an index builder mentioned
-above. See the :doc:`Indexes </fundamentals/indexes>` page to see
-examples of queries on indexes. 
+Indexes store small portions of the collection’s data set. The index
+stores the value of a specific field or set of fields, ordered by the
+value of the field. See the :doc:`Indexes </fundamentals/indexes>` page
+to see examples of queries on indexes. 
 
 The :java-core-api:`Indexes <com/mongodb/client/model/Indexes.html>`
-class provides static factory methods for all the MongoDB index key
-types. Each method returns a :ref:`Bson <bson>` instance, which you can
-pass to :java-sync-api:`createIndex() <com/mongodb/client/MongoCollection.html#createIndex(com.mongodb.client.ClientSession,org.bson.conversions.Bson)>`.
+class provides static factory methods for all the MongoDB index types.
+Each method returns a :ref:`Bson <bson>` instance, which you can pass to
+:java-sync-api:`createIndex() <com/mongodb/client/MongoCollection.html#createIndex(org.bson.conversions.Bson)>`.
 
 .. tip::
 
@@ -54,14 +53,21 @@ pass to :java-sync-api:`createIndex() <com/mongodb/client/MongoCollection.html#c
 Ascending
 ---------
 
-Pass the field name you want to index to the 
-:java-core-api:`ascending() <com/mongodb/client/model/Indexes.html#ascending(java.lang.String...)>`
-method to create an ascending index. An ascending index sorts the
-index key from smallest to largest.
+Create a BSON instance by passing the field name or names you want to
+index to the :java-core-api:`ascending() <com/mongodb/client/model/Indexes.html#ascending(java.lang.String...)>`
+method. Then, pass the BSON instance into the createIndex() method to
+create an ascending index. 
 
-The following example specifies an ascending index key for the ``name`` field:
+An ascending index sorts the index key from smallest to largest.
 
-.. literalinclude:: /includes/fundamentals/code-snippets/Indexes.java
+.. note:: 
+
+If an ascending or a descending index is on a single field, the sort
+operation on the field can be in either direction. 
+
+The following example specifies an ascending index on the ``name`` field:
+
+.. literalinclude:: /includes/fundamentals/code-snippets/builders/Indexes.java
    :language: java
    :dedent:
    :start-after: begin ascendingIndex
@@ -72,15 +78,16 @@ The following example specifies an ascending index key for the ``name`` field:
 Descending
 ----------
 
-Pass the field name you want to index to the 
-:java-core-api:`descending() <com/mongodb/client/model/Indexes.html#descending(java.lang.String...)>`
-method to create a descending index. A descending index sorts the
-index key from largest to smallest.
+Create a BSON instance by passing the field name or names you want to
+index to the :java-core-api:`descending() <com/mongodb/client/model/Indexes.html#descending(java.lang.String...)>`
+method. Then, pass the BSON instance into the createIndex() method to
+create a descending index. 
 
-The following example specifies a compound index key composed of
-descending index keys for the ``capacity`` field: 
+A descending index sorts the index key from largest to smallest.
 
-.. literalinclude:: /includes/fundamentals/code-snippets/Indexes.java
+The following example specifies a descending index on the ``capacity`` field: 
+
+.. literalinclude:: /includes/fundamentals/code-snippets/builders/Indexes.java
    :language: java
    :dedent:
    :start-after: begin descendingIndex
@@ -91,15 +98,16 @@ descending index keys for the ``capacity`` field:
 Compound Indexes
 ----------------
 
-Pass the field name you want to index to the 
-:java-core-api:`compoundIndex() <com/mongodb/client/model/Indexes.html#compoundIndex(org.bson.conversions.Bson...)>`
-method to create a compound index. 
+Create a BSON instance by passing the field names you want to index to
+the :java-core-api:`compoundIndex() <com/mongodb/client/model/Indexes.html#compoundIndex(org.bson.conversions.Bson...)>`
+method. Then, pass the BSON instance into the createIndex() method to
+create a compound index. 
 
-The following example specifies a compound index key composed of
-descending index keys for the ``capacity`` and ``year`` field, followed
-by an ascending index key for the ``name`` field: 
+The following example specifies a compound index composed of
+descending index on the ``capacity`` and ``year`` field, followed
+by an ascending index on the ``name`` field: 
 
-.. literalinclude:: /includes/fundamentals/code-snippets/Indexes.java
+.. literalinclude:: /includes/fundamentals/code-snippets/builders/Indexes.java
    :language: java
    :dedent:
    :start-after: begin compoundIndexExample
@@ -110,14 +118,16 @@ by an ascending index key for the ``name`` field:
 Text Indexes
 ------------
 
-Pass the field name you want to index to the 
-:java-core-api:`text() <com/mongodb/client/model/Indexes.html#text(java.lang.String)>`
-method to create a text index key. A text index groups the index key
-by the word you specify.
+Create a BSON instance by passing the field name you want to index to
+the :java-core-api:`text() <com/mongodb/client/model/Indexes.html#text(java.lang.String)>`
+method. Then, pass the BSON instance into the createIndex() method to
+create a text index key. 
 
-The following example specifies a text index key for "theaters":
+A text index groups the index key by the word you specify.
 
-.. literalinclude:: /includes/fundamentals/code-snippets/Indexes.java
+The following example specifies a text index key on the "theaters" field:
+
+.. literalinclude:: /includes/fundamentals/code-snippets/builders/Indexes.java
    :language: java
    :dedent:
    :start-after: begin textIndex
@@ -128,15 +138,17 @@ The following example specifies a text index key for "theaters":
 Hashed Indexes
 --------------
 
-Pass the field name you want to index to the 
-:java-core-api:`hashed() <com/mongodb/client/model/Indexes.html#hashed(java.lang.String)>`
-method to create a hashed index key. A hashed index groups the index
-key by its hash value.
+Create a BSON instance by passing the field name you want to index to
+the :java-core-api:`hashed() <com/mongodb/client/model/Indexes.html#hashed(java.lang.String)>`
+method. Then, pass the BSON instance into the createIndex() method to
+create a hashed index key. 
 
-The following example specifies a hashed index key for the ``capacity``
+A hashed index groups the index key by its hash value.
+
+The following example specifies a hashed index on the ``capacity``
 field:
 
-.. literalinclude:: /includes/fundamentals/code-snippets/Indexes.java
+.. literalinclude:: /includes/fundamentals/code-snippets/builders/Indexes.java
    :language: java
    :dedent:
    :start-after: begin hashedIndex
@@ -147,14 +159,16 @@ field:
 Geospatial Indexes
 ------------------
 
-Pass the field name you want to index to the 
-:java-core-api:`geo2dsphere() <com/mongodb/client/model/Indexes.html#geo2dsphere(java.lang.String...)>`
-method to create a geo2dsphere index key. A geo2dsphere index
-groups the index key by their coordinates.
+Create a BSON instance by passing the field name or names you want to
+index to the :java-core-api:`geo2dsphere() <com/mongodb/client/model/Indexes.html#geo2dsphere(java.lang.String...)>`
+method . Then, pass the BSON instance into the createIndex() method to
+create a geo2dsphere index key. 
+
+A geo2dsphere index groups the index key by their coordinates.
 
 The following example specifies a geo2dsphere index on the ``location`` field:
 
-.. literalinclude:: /includes/fundamentals/code-snippets/Indexes.java
+.. literalinclude:: /includes/fundamentals/code-snippets/builders/Indexes.java
    :language: java
    :dedent:
    :start-after: begin geo2dsphereIndex
--- a/source/fundamentals/builders/indexes.txt
+++ b/source/fundamentals/builders/indexes.txt
@@ -19,8 +19,8 @@ This guide shows you how to specify **indexes** using the Java Driver
 :doc:`builders </fundamentals/builders/>`. Index builders provide
 helper methods for the following types of indexes:
 
-- :ref:`Ascending <index-ascending>`
-- :ref:`Descending <index-descending>`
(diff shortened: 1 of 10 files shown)


ccho-mongodb (changes requested):
Left some comments. Here's a summary of my change requests:

- make sure we're adhering to MongoDB terminology when possible
- identify audience and try to include information that may help them understand the information on the page (not an easy task, but should get easier over time)
- check for any extraneous or missing content when proofreading
- always use a spell-checker before requesting a review; it's important to polish our writing with something that we can automate

source/fundamentals/builders/indexes.txt
```diff
@@ -11,3 +11,156 @@ Indexes Builders
...
+   { "_id": 4, "name": "Classy Views", "logo": "red", "capacity": 847, "year": 1998, "location": { "type": "Point", "coordinates": [-119.7412, 39.579536] } };
+   { "_id": 5, "name": "Fortune Theaters", "logo": "red", "capacity": 1368, "year": 2007, "location": { "type": "Point", "coordinates": [-100.81213, 46.829876] } };
+   { "_id": 6, "name": "8K Screening", "logo": "blue", "capacity": 619, "year": 2019, "location": { "type": "Point", "coordinates": [-118.11414, -118.114143] } };
+   { "_id": 7, "name": "Date Night", "logo": "red", "capacity": 715, "year": 2012, "location": { "type": "Point", "coordinates": [-100.50107, 31.435648] } };
+   { "_id": 8, "name": "City's Best Theaters", "logo": "red", "capacity": 888, "year": 2016, "location": { "type": "Point", "coordinates": [-78.382912, 40.490524] } };
+   { "_id": 9, "name": "Pets Movies", "logo": "blue", "capacity": 1092, "year": 2010, "location": { "type": "Point", "coordinates": [-117.674814, 33.590599] } };
+   { "_id": 10, "name": "Fancy Theaters", "logo": "red", "capacity": 673, "year": 2006, "location": { "type": "Point", "coordinates": [-72.583824, 41.998211] } };
```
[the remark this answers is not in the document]
kyuan-mongodb: It's not, but should I include one saying that is how all documents look like in the collection to show the format each index is creating?

source/fundamentals/builders/indexes.txt
```diff
@@ -11,3 +11,156 @@ Indexes Builders
...
+- :ref:`Geospatial Indexes <geospatial-indexes>`
+
+Indexes support the efficient execution of queries in MongoDB. Instead
+of checking every document, indexes help limit the search to check
+the groups of data where your query should be. See the
+:doc:`Indexes </fundamentals/indexes>` page to see examples of queries
+on indexes.  
+
+The :java-core-api:`Indexes <com/mongodb/client/model/Indexes.html>`
+class provides static factory methods for all the MongoDB Index key
```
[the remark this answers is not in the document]
kyuan-mongodb: https://mongodb.github.io/mongo-java-driver/4.2/builders/indexes/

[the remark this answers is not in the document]
ccho-mongodb: The MongoDB manual conflicts with that, see https://docs.mongodb.com/manual/indexes/

Unless it's something specific to the Java implementation, we should try to stay consistent with the terminology standards used in the MongoDB manual. If uncertain, you can reach out to the Java driver engineers on Slack to determine whether it is specific to the Java implementation. If it is, we should make a point to explain it if it's not documented anywhere else.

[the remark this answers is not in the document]
ccho-mongodb: Also, what do you think about this comment about "index key"? I saw you resolved it, but not sure it was addressed:
https://github.com/mongodb/docs-java/pull/64#discussion_r594717698

source/fundamentals/builders/indexes.txt
```diff
@@ -11,3 +11,150 @@ Indexes Builders
...
+
+Like an index in a textbook, an index in MongoDB groups data based
+on the field you provide and by an index builder mentioned above. See
+the :doc:`Indexes </fundamentals/indexes>` page to see examples of
+queries on indexes. 
+
+The :java-core-api:`Indexes <com/mongodb/client/model/Indexes.html>`
+class provides static factory methods for all the MongoDB Index key
+types. Each method returns a :ref:`Bson <bson>` instance, which you can
+pass to `createIndex() <https://mongodb.github.io/mongo-java-driver/4.2/apidocs/mongodb-driver-sync/com/mongodb/client/MongoCollection.html#createIndex(com.mongodb.client.ClientSession,org.bson.conversions.Bson)>`_.
```
ccho-mongodb, line 37: Let's use the java-docs directive here so that it's theoretically easier to update in the future.

source/fundamentals/builders/indexes.txt
```diff
@@ -11,3 +11,150 @@ Indexes Builders
...
+convenient methods for different types of indexes:
+
+- :ref:`Ascending <index-ascending>`
+- :ref:`Descending <index-descending>`
+- :ref:`Compound Indexes <index-compound-indexes>`
+- :ref:`Text Indexes <index-text-indexes>`
+- :ref:`Hashed Indexes <index-hashed-indexes>`
+- :ref:`Geospatial Indexes <geospatial-indexes>`
+
+Like an index in a textbook, an index in MongoDB groups data based
```
ccho-mongodb, line 29: I like that you're using using an analogy, but I think this one might need more explanation. I should be able to read the description of how MongoDB groups data in the context of a textbook which would read:

In an index in a textbook, data is grouped based on the field you provide and an index builder.

Not sure what that corresponds to in a textbook. Maybe you can just limit the sentence to what parts of it are exactly the same and then describe the details specific to MongoDB in a following sentence.

source/fundamentals/builders/indexes.txt
```diff
@@ -11,3 +11,150 @@ Indexes Builders
...
+   The examples below assume this static import.
+
+.. _index-ascending:
+
+Ascending
+---------
+
+Pass the field name you want to index to the 
+:java-core-api:`ascending() <com/mongodb/client/model/Indexes.html#ascending(java.lang.String...)>`
+method to create an ascending index. An ascending index will sort the
```
ccho-mongodb, line 58: Let's avoid using the future tense (identified here by "will") when possible as the style guide asks to use present tense https://docs.mongodb.com/meta/style-guide/writing/use-present-tense/:

```suggestion
method to create an ascending index. An ascending index sorts the
```

Applies to all instances.

source/fundamentals/builders/indexes.txt
```diff
@@ -11,3 +11,150 @@ Indexes Builders
...
+
+Text Indexes
+------------
+
+Pass the field name you want to index to the 
+:java-core-api:`text() <com/mongodb/client/model/Indexes.html#text(java.lang.String)>`
+method to create a text index key. A text index will group the index key
+by the word you specify.
+
+The following example specifies a text index key for "Theaters":
```
ccho-mongodb, line 117: Let's use lowercase for field names to follow javascript naming conventions. There's technically no restriction, but we use the javascript/JSON camelCase convention throughout our documentation.

ccho-mongodb (changes requested):
Looks mostly good, left a few comments to address.

source/fundamentals/builders/indexes.txt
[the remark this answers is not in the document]
kyuan-mongodb: I agreed with "index key" and fixed all occurrences of it. 
But are you saying "Index key" should be "indexes key"?

ccho-mongodb (changes requested):
Left some comments.
I think we need to make it clear to the reader that the builder methods are not creating indexes in every section because the current wording makes it seem like they are.

source/fundamentals/builders/indexes.txt
```diff
@@ -11,3 +11,151 @@ Indexes Builders
...
+   The examples below assume this static import.
+
+.. _index-ascending:
+
+Ascending
+---------
+
+Pass the field name you want to index to the 
+:java-core-api:`ascending() <com/mongodb/client/model/Indexes.html#ascending(java.lang.String...)>`
+method to create an ascending index. An ascending index sorts the
```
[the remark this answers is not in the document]
kyuan-mongodb: It's actually mentioned at line 38 that it gets passed to createIndex(). Should it be repeated throughout each type of index as a reminder?

source/fundamentals/builders/indexes.txt
```diff
@@ -11,3 +11,151 @@ Indexes Builders
...
+
+   .. code-block:: java
+
+      import static com.mongodb.client.model.Indexes.*;
+
+   The examples below assume this static import.
+
+.. _index-ascending:
+
+Ascending
```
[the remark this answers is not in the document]
kyuan-mongodb: You mean the order of ascending and descending don't matter in a compound index?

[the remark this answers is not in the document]
ccho-mongodb: No, maybe I need to be more clear in my comment-writing. You can read it as follows:

Somewhere, we should mention that if we only specify a single index key -- not a compound one -- it does not matter whether we create an ascending or descending index because MongoDB can scan both directions.

source/fundamentals/builders/indexes.txt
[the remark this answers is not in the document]
ccho-mongodb: I'm not making the claim that no user can draw the connection if they've read the text around line 38, but it does require additional information to be interpreted correctly.

Your sentence  "[you can] pass the field name you want to index to the ascending() method to create an ascending index" reads as if that is the only step to create the index.

Take this sentence as an example of a sentence that also requires additional information to make sense:
"Save your file in VSCode to publish it on docs.mongodb.com."

And if I included a statement earlier that said "you need to commit saved files to a repository that has an autobuilder webhook set up to publish your documentation", you've pointed the user to a potential outcome, but it doesn't mean that saving your file achieves the outcome of publishing it on docs.mongodb.com.

The Builders are only creating Bson instances, so if you want to claim that they create the index, you need to qualify it by stepping through what's required to reach the outcome of creating it. Let me know if that makes sense.

source/fundamentals/builders/indexes.txt
```diff
@@ -11,3 +11,165 @@ Indexes Builders
...
+---------
+
+Create a BSON instance by passing the field name or names you want to
+index to the :java-core-api:`ascending() <com/mongodb/client/model/Indexes.html#ascending(java.lang.String...)>`
+method. Then, pass the BSON instance into the createIndex() method to
+create an ascending index.
```
ccho-mongodb, line 59: I think all the information is here, but let's reorder it so that we present what we're trying to accomplish first (show the user how to create an ascending index) instead of last. Also, make sure you distinguish between the ``Bson`` type and BSON format (here, we are talking about the former).

```suggestion
In order to create an ascending index, first call the ``ascending()`` builder method to create a ``Bson`` instance that represents the index document., passing the name or names of the fields you want to index. Then, call 
the ``createIndex()`` method on the collection, passing the``Bson`` instance that contains the index document.
```

Same recommendation for all instances.

source/fundamentals/builders/indexes.txt
```diff
@@ -11,3 +11,165 @@ Indexes Builders
...
+
+Ascending
+---------
+
+Create a BSON instance by passing the field name or names you want to
+index to the :java-core-api:`ascending() <com/mongodb/client/model/Indexes.html#ascending(java.lang.String...)>`
+method. Then, pass the BSON instance into the createIndex() method to
+create an ascending index. 
+
+An ascending index sorts the index key from smallest to largest.
```
ccho-mongodb, line 61: Should this be at the top of the section so the reader understands what they're looking at before reading how to implement it? 

Same question for all instances.

source/fundamentals/builders/indexes.txt
```diff
@@ -11,3 +11,165 @@ Indexes Builders
...
+
+Geospatial Indexes
+------------------
+
+Create a BSON instance by passing the field name or names you want to
+index to the :java-core-api:`geo2dsphere() <com/mongodb/client/model/Indexes.html#geo2dsphere(java.lang.String...)>`
+method. Then, pass the BSON instance into the createIndex() method to
+create a geo2dsphere index key. 
+
+A geo2dsphere index groups the index key by their coordinates.
```
ccho-mongodb, line 167: This might be more accurate and succinct since the index groups documents, not index keys:

```suggestion
A geo2dsphere index groups documents by the coordinates in the indexed field.
```

Similar comment applies to all instances in which you mention "groups the index key".

source/fundamentals/builders/indexes.txt
ccho-mongodb: The index key is the name of the field, e.g.
```
{ item: 1, quantity: -1 }
```
In the above document that specifies an index, "item" and "quantity" are the index keys. The index sorts documents according to the values they hold in those fields. The index is a btree data structure that's optimized for locating and ordering results based on the type of index. MongoDB traverses the btree (index) to locate documents or provide a sort order. Documents in the collection do not contain index keys, just fields and values. 

```suggestion
An ascending index enables you to sort query results by the value of the indexed field(s) from smallest to largest.
```

Similar comment applies to all instances.

ccho-mongodb (changes requested):
I think all the information is there, but could use a little clean up and clarification.

source/fundamentals/builders/indexes.txt
```diff
@@ -11,3 +11,181 @@ Indexes Builders
...
+- :ref:`Descending <index-descending>`
+- :ref:`Compound Indexes <index-compound-indexes>`
+- :ref:`Text Indexes <index-text-indexes>`
+- :ref:`Hashed Indexes <index-hashed-indexes>`
+- :ref:`Geospatial Indexes <index-geospatial-indexes>`
+
+Indexes store a subset of the collection’s data set. The index stores
+the value of a specific field or set of fields, ordered by the value of
+the field. See our guide on :doc:`Indexes </fundamentals/indexes>` to
+see examples of queries covered by indexes.
```
ccho-mongodb, line 32: Nitpick: sorry, missed this in my suggestion from the previous review. No change necessary, but it might be better to ask them to "see" only once.
```suggestion
the field. See our guide on :doc:`Indexes </fundamentals/indexes>` for
examples of queries covered by indexes. 
```

source/fundamentals/builders/indexes.txt
```diff
@@ -11,3 +11,181 @@ Indexes Builders
...
+- :ref:`Geospatial Indexes <index-geospatial-indexes>`
+
+Indexes store a subset of the collection’s data set. The index stores
+the value of a specific field or set of fields, ordered by the value of
+the field. See our guide on :doc:`Indexes </fundamentals/indexes>` to
+see examples of queries covered by indexes. 
+
+The :java-core-api:`Indexes <com/mongodb/client/model/Indexes.html>`
+class provides static factory methods for all the MongoDB index types.
+Each method returns a :java-docs:`Bson <apidocs/bson/org/bson/conversions/Bson.html?is-external=true>`
```
ccho-mongodb, line 36: Nitpick: `?is-external=true` isn't necessary, and I'm not sure what it's used for.

```suggestion
Each method returns a :java-docs:`Bson <apidocs/bson/org/bson/conversions/Bson.html?is-external=true>` 
```

source/fundamentals/builders/indexes.txt
```diff
@@ -11,3 +11,181 @@ Indexes Builders
...
+:java-core-api:`ascending() <com/mongodb/client/model/Indexes.html#ascending(java.lang.String...)>`
+builder method to create a ``Bson`` instance that represents the index
+document, passing the name or names of the fields you want to index.
+Then, call the ``createIndex()`` method on the collection, passing the ``Bson``
+instance that contains the index document.
+
+.. note:: 
+
+   If an ascending or a descending index is on a single field, the sort
+   operation on the field can be in either direction.
```
ccho-mongodb, line 70: Nitpick: I think this might be a little more clear in describing the sort behavior.

```suggestion
   If you have an ascending or a descending index on a single field, MongoDB
   can sort using the index in either direction. 
```

source/fundamentals/builders/indexes.txt
```diff
@@ -11,3 +11,181 @@ Indexes Builders
...
+
+Overview 
+--------
+
+This guide shows you how to specify **indexes** using the Java Driver
+:doc:`builders </fundamentals/builders/>`. Index builders provide
+helper methods for the following types of indexes:
+
+- :ref:`Ascending <index-ascending>`
+- :ref:`Descending <index-descending>`
```
ccho-mongodb, line 23: Nitpick:
Just thought it was strange that the other bulleted items include "Indexes" and these don't. Should they be similar/different?

ccho-mongodb (approved):
LGTM! Just a few nitpicks.

source/fundamentals/builders/indexes.txt
```diff
@@ -11,3 +11,181 @@ Indexes Builders
...
+
+.. _index-hashed-indexes:
+
+Hashed Indexes
+--------------
+
+A hashed index groups documents by the hash value in the indexed field.
+
+In order to create a hashed index, first call the
+:java-core-api:`hashed() <com/mongodb/client/model/Indexes.html#hashed(java.lang.String)>`
```
jyemin, line 155: Whoops, looks like the link to the MongoDB docs contained in the Java API docs is broken (https://docs.mongodb.com/manual/core/hashed/).

Any way we can get that link to redirect to https://docs.mongodb.com/manual/core/index-hashed/index.html?

jyemin (approved):
One comment but it's tangentially related.  LGTM!

source/fundamentals/builders/indexes.txt
kyuan-mongodb: The link I have there is actually: https://mongodb.github.io/mongo-java-driver/4.2/apidocs/mongodb-driver-core/com/mongodb/client/model/Indexes.html#hashed(java.lang.String) 

But if you'd like, I can switch it to the link you requested

jyemin: Sorry, I wasn't clear.  It's the link from the driver API docs to the MongoDB docs that is broken.  I was hoping that we can make that link work again as we can't easily update the Javadoc for already-released versions of the driver.

kyuan-mongodb: I created a ticket to get a redirect created: https://jira.mongodb.org/browse/DOP-2067?filter=-2

PR #65, opened 2021-03-19
DOCSP-12161 fundamentals codecs

ccho-mongodb:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-12161

https://docs-mongodbcom-staging.corp.mongodb.com/7481d86/java/docsworker-xlarge/DOCSP-12161-fundamentals-codecs/fundamentals/data-formats/codecs/

### Self-Review Checklist


source/fundamentals/data-formats/codecs.txt
```diff
@@ -11,38 +11,337 @@ Codecs
...
+
+If you are customizing encoding and decoding logic for Plain old Java objects
+(POJOs), read our guide on :doc:`POJO Customization </fundamentals/data-formats/pojo-customization>`.
+
+.. _codecs-codec:
+
+Codec
+-----
+
+The ``Codec`` interface contains abstract methods for serializing and
```
biniona-mongodb, line 42:
```suggestion
The :java-docs:`Codec <apidocs/bson/org/bson/codecs/Codec.html>` interface contains abstract methods for serializing and
```

biniona-mongodb: Should this link to the API docs for Codec interface?

source/fundamentals/data-formats/codecs.txt
```diff
@@ -11,38 +11,337 @@ Codecs
...
+     - An instance of a class that implements ``BsonReader``, an interface type
+       that exposes methods for reading a BSON document. For example, the
+       ``BsonBinaryReader`` implementation reads from a binary stream of data.
+
+   * - ``decoderContext``
+     - Contains information about the BSON data that it decodes to a Java
+       object.
+
+The ``getEncoderClass()`` method returns a class instance of the Java class
+since Java cannot infer the type due to due to type erasure.
```
biniona-mongodb, line 94:
```suggestion
since Java cannot infer the type due to type erasure.
```

source/fundamentals/data-formats/codecs.txt
```diff
@@ -11,38 +11,337 @@ Codecs
...
-- Description
-- Example of get method implementation
+A ``CodecProvider`` is an interface that contains abstract methods that create
+``Codec`` instances and assign them to a ``CodecRegistry`` instance. Similar
+to the ``CodecRegistry``, the BSON library uses the Codecs retrieved by
+the ``get()`` method to convert between Java and BSON data types.
+
+However, in cases in which you add a class that contains fields that require
+corresponding ``Codec`` instances, you need to ensure you construct the Codecs
+for the class' fields before the ``Codec`` for the class. You can use the
```
biniona-mongodb, line 186:
```suggestion
for the class' fields before the ``Codec`` instance for the class. You can use the
```

biniona-mongodb: Should this specify `Codec` instance?

source/fundamentals/data-formats/codecs.txt
```diff
@@ -11,38 +11,337 @@ Codecs
...
+``CodecProvider`` to add the Codecs to the registry before the one for which
+they are required.
+
+The following code example shows how you can implement ``CodecProvider`` to
+instantiate the Codecs on which the ``Monolight`` depends including the
```
biniona-mongodb, line 191: Should `Monolight` and `PowerStatusCodec` link to the section of the document where they are specified.

source/fundamentals/data-formats/codecs.txt
```diff
@@ -11,38 +11,337 @@ Codecs
...
+.. code-block:: none
+   :copyable: false
+
+   Java type: java.util.Set
+
+For a complete list of the default mappings, refer to the
+:java-docs:`BsonTypeClassMap API documentation <apidocs/bson/org/bson/codecs/BsonTypeClassMap.html>`.
+
+For an example of how the ``Document`` class uses ``BsonTypeClassMap``, see
+the driver source code for the following classes:
```
biniona-mongodb, line 266: Should these source code links link to a line number? It is difficult to tell where in the source code you are referring to.

source/fundamentals/data-formats/codecs.txt
```diff
@@ -11,38 +11,337 @@ Codecs
...
+     - Contains information about the BSON data that it decodes to a Java
+       object.
+
+The ``getEncoderClass()`` method returns a class instance of the Java class
+since Java cannot infer the type due to due to type erasure.
+
+.. _codecs-powerstatus-codec:
+
+See the following ``PowerStatusCodec`` class that implements ``Codec`` in
+order to convert the Java ``String`` values "on" and "off" to
```
biniona-mongodb, line 99: Are we switching to the convention  where we use monospace and language syntax for values, so`"on"` and `"off"`?

biniona-mongodb (commented):
Looks good! Just submitting a review to practice reviewing. Interested to see what other's find. Please let me know if you disagree with any of my comments or with anything about my approach.

source/fundamentals/data-formats/codecs.txt
```diff
@@ -11,38 +11,337 @@ Codecs
...
+-----
+
+The ``Codec`` interface contains abstract methods for serializing and
+deserializing Java objects to BSON data. You can define your conversion logic
+between BSON and your Java object in your implementation of this interface.
+
+To implement the ``Codec`` interface, override the ``encode()``, ``decode()``,
+and ``getEncoderClass()`` abstract methods.
+
+The ``encode()`` method requires the following parameters:
```
kyuan-mongodb, line 49: You mention what the methods decode() and getEncoderClass() return.  For consistency, maybe include it for encode() too

source/fundamentals/data-formats/codecs.txt
```diff
@@ -11,38 +11,337 @@ Codecs
...
+
+.. list-table::
+   :header-rows: 1
+   :widths: 10 20
+
+   * - Parameter Type
+     - Description
+
+   * - ``writer``
+     - An instance of a class that implements ``BsonWriter``, an interface type
```
kyuan-mongodb, line 59:
```suggestion
     - An instance of a class that implements ``BsonWriter``, and an interface type
```

kyuan-mongodb: I'm not sure if you meant it the way you wrote it, but it sounded incomplete when I read this.  Applies to all instances.

source/fundamentals/data-formats/codecs.txt
```diff
@@ -11,38 +11,337 @@ Codecs
...
+You can retrieve the Codecs from the ``CodecRegistry`` instance from the
+prior example using the following code:
+
+.. code-block:: java
+   :copyable: true
+
+   Codec<String> powerStatus = codecRegistry.get(String.class);
+   Codec<Integer> integerCodec = codecRegistry.get(Integer.class);
+
+If you attempt to retrieve a ``Codec`` instance for a class that is not
```
kyuan-mongodb, line 164: I think this is should be in a note block to make it stand out more to the reader.

kyuan-mongodb (commented):
Looks good overall, but I think there are a few things that can improve the page.

source/fundamentals/data-formats/codecs.txt
```diff
@@ -11,38 +11,337 @@ Codecs
...
+  that converts certain Java strings to BSON booleans.
+
+You can retrieve the Codecs from the ``CodecRegistry`` instance from the
+prior example using the following code:
+
+.. code-block:: java
+   :copyable: true
+
+   Codec<String> powerStatus = codecRegistry.get(String.class);
+   Codec<Integer> integerCodec = codecRegistry.get(Integer.class);
```
terakilobyte, line 162: Since this can throw, is it worth wrapping in a try/catch?

source/fundamentals/data-formats/codecs.txt
terakilobyte: Perhaps a `.. warning`

source/fundamentals/data-formats/codecs.txt
```diff
@@ -11,38 +11,337 @@ Codecs
...
 BsonTypeClassMap
 ~~~~~~~~~~~~~~~~
 
-- Description
-- Example of how to use to specify Bson mapping
+The ``BsonTypeClassMap`` class contains a recommended mapping between BSON
+and Java types. You can use this class to help you manage which Java types
+to decode your BSON types to in your custom ``Codec`` or ``CodecProvider``.
+You can add or modify the mapping by passing a ``Map`` containing new or
+replacement or entries. The ``Document`` class uses the recommended mapping.
```
terakilobyte, line 220:
```suggestion
replacement entries. The ``Document`` class uses the recommended mapping.
```

source/fundamentals/data-formats/codecs.txt
```diff
@@ -11,38 +11,337 @@ Codecs
...
+   System.out.println("Java type: " + clazz.getName());
+
+This code outputs the following:
+
+.. code-block:: none
+   :copyable: false
+
+   Java type: java.util.List
+
+You can modify these mappings in your instance by specifying them in the
```
terakilobyte, line 240:
```suggestion
You can modify these mappings in your instance by specifying replacements in the
```
nitpick:  'them' was unclear to me here until reading further below

source/fundamentals/data-formats/codecs.txt
```diff
@@ -11,38 +11,337 @@ Codecs
...
+
+This class contains the following fields, each of which we need to assign a
+``Codec``:
+
+- ``colorTemperature`` describes the color of the light and contains an
+  ``Integer`` value for which we use the ``IntegerCodec`` included in the
+  BSON library.
+- ``powerStatus`` describes whether the light is "on" or "off" for which we use
+  the :ref:`PowerStatusCodec <codecs-powerstatus-codec>` that converts specific
+  string values to BSON booleans.
```
terakilobyte, line 298: should these be ordered here the same as they are declared in the code, or does it matter?

terakilobyte (changes requested):
Overall very clear and easy to follow! A few nits and questions.

source/fundamentals/data-formats/codecs.txt
ccho-mongodb: We should avoid links to API docs or other links external to the drivers/java site in the opening paragraph of a section that describes a concept. This decision is based on user and MongoDB writer feedback to keep the reader focused and instead link it later (see line 116).

source/fundamentals/data-formats/codecs.txt
ccho-mongodb: I initially thought about that, but decided against it since it returns nothing. Since it sounds like the additional clarification can help, I'll add to the description so that it's more apparent that it doesn't return anything and why that is.

source/fundamentals/data-formats/codecs.txt
ccho-mongodb: In this case, the comma is marking the start of a nonrestrictive clause to define the term "BsonWriter". The "BsonWriter" is the  "interface type that exposes methods for writingb a BSON document". If this was a list of two items, "BsonWriter" and "something else", the comma should not be included.

source/fundamentals/data-formats/codecs.txt
ccho-mongodb: Not that I know of, but maybe you have a different convention in mind? 

I've mentioned what convention I've been following for all the docs repos that our team currently works on, but can explain again. 

I'd like to stick to that for at least the current documentation repos, but if there's a pressing need to change it, we should discuss as a team and apply the convention across the docs.

source/fundamentals/data-formats/codecs.txt
ccho-mongodb: I think it could either:
- show the try/catch and omit the description underneath because that could be made apparent through the error message
- keep it short to just demonstrate the syntax, and then explain a potential runtime exception as it is currently structured

Is there one way that's better than the other? I don't think we've done much with handling specific exceptions in our prior documentation.

terakilobyte: When I did the indexes page, I wrapped some of the index creation methods in try/catch since they can throw. I'm unsure what the best approach is, perhaps it is to wrap in try/catch and catch the specific exception. If there is more than one, perhaps we add a note in that case?

source/fundamentals/data-formats/codecs.txt
ccho-mongodb: Maybe I'm missing something about how we should use admonitions (note, warning, etc.), but here's my decision process:

I've only used warning or note blocks for describing some unexpected behavior whether it's a difference between how the driver and server behave or a common user mistake.

This specific exception type is a feature that helps a user identify what went wrong during runtime. I'm not sure we should be using an admonition to let a user know about this.

Let me know if this makes sense and/or whether I'm missing something.

source/fundamentals/data-formats/codecs.txt
ccho-mongodb: Good catch, will update.

source/fundamentals/data-formats/codecs.txt
ccho-mongodb: I think it depends on factors including intent, feasibility, maintainability, and audience.

In this case:

- The intent is not to walk the user through the source code, but rather to mention they can trace the code themselves independently for further reading/learning. We've already explained one example of its usage above.

- For feasibility, in this case it's multiple lines of code, so linking to one line could be deceptive more than helpful, making the user believe that the reference might only be in one place.

- For maintainability, we need to be extra careful about linking to line numbers in source code since those can change in point releases without us knowing.

- As for audience, this paragraph is explicitly mentioning driver source code to allow users to determine whether this is the type of learning they want to pursue. 

Do you disagree with any of these points? And can you think of a better solution that addresses your request?

source/fundamentals/data-formats/codecs.txt
ccho-mongodb: I was thinking that "construct" only applies to objects/instances, so that could be implied. Since it could be misinterpreted, I'll reword this to "instantiate a Codec".

source/fundamentals/data-formats/codecs.txt
biniona-mongodb: From a technical perspective, I think we can circumvent the maintainability and feasibility concerns (points 2 and 3) by linking to multiple lines of a file at a specific commit hash. Example: https://github.com/mongodb/mongo-java-driver/blob/bd522224a66418c89595601263265da510843206/bson/src/main/org/bson/codecs/DocumentCodecProvider.java#L40-L63.

In regards to point 1, I think readers will have a more productive time tracing the source code with a definite starting point. Often for me the hardest thing about reading source code is gaining my first foothold. Having an explicit place to start/end has in the past been extremely helpful for me in beginning the process of studying an unfamiliar project.

In regards to point 4, I think there might have been a miscommunication due to me incorrectly selecting the lines I wished to comment on. The only things I think could benefit from a change are lines 268 and 269 such that the `href`s link to line numbers of a specific commit hash:

`- :github:`DocumentCodecProvider <mongodb/mongo-java-driver/blob/master/bson/src/main/org/bson/codecs/DocumentCodecProvider.java>`
- :github:`DocumentCodec <mongodb/mongo-java-driver/blob/master/bson/src/main/org/bson/codecs/DocumentCodec.java>``

source/fundamentals/data-formats/codecs.txt
ccho-mongodb: That was my first thought, but when considering the following, I decided it was better not to do so for the following reasons:
1) Since we can't give links monospace formatting, we would need to provide additional information about the link as we do (or should do) with other links.
2) Keep the reader focused on what we're demonstrating which in this case doesn't offer information to understand the example other than maybe the type that getEncoderClass() returns in PowerStatusCodec. The content referenced in the link on line 201 features a link back to PowerStatusCodec including a description.

Let me know if that makes sense or whether you have a better solution, and whether this is a blocker for you.

source/fundamentals/data-formats/codecs.txt
ccho-mongodb: I don't think we should try to show and explain all the potential exceptions, other than those that are ones that might expect a user to frequently encounter. A few reasons for this:

- Java IDEs like Eclipse and IntelliJ generate try/catch block code for any of the potential exception types.
- There are a lot of different exception types in the methods we document, so describing each of them in detail beyond what's offered in the API documentation we link to has all the disadvantages we've discussed in denormalizing API documentation in reference documentation applies.

source/fundamentals/data-formats/codecs.txt
terakilobyte: I think this makes sense, especially since it isn't surprising.

source/fundamentals/data-formats/codecs.txt
ccho-mongodb: > In regards to point 4, I think there might have been a miscommunication due to me incorrectly selecting the lines I wished to comment on. The only things I think could benefit from a change are lines 268 and 269 such that the hrefs link to line numbers of a specific commit hash. These are the two lines I think could be changed:

I think those are the same links I'm referring to as they're the only ones that link to the driver repository. In your recommended lines to highlight, I'm not sure whether those are sufficient since the variable that contains the reference is used in more methods than the highlighted ones. I'm not sure a reader can understand the intent without explanation either.

Unfortunately, linking to a commit hash doesn't make it more maintainable than linking to a branch.  Since we're presenting documentation for specific versions, we would need to ensure we're showing them the "current implementation for this version" rather than "an implementation in a version, but not necessarily this version". Since a referenced commit is not guaranteed to be the same across PATCH versions, we can't guarantee those line numbers are valid in their MINOR version (for definitions of MINOR and PATCH, see semver.org).

Current issues that we've identified in this conversation:
- the user might not find the links to source code helpful without more guidance
- as document creators, it is difficult to guide some users through all the details of the source code (especially with just line numbers).
- as document creators, it's difficult to come up with a way to make documentation that guides a user through specific lines of source code maintainable

Initially, I thought it could be helpful, but I think from hearing your feedback, maybe I should do one or more of the following:

1) omit the link to the source code
2) replace it with a contrived example of how it might be able to used in the decode() method of a Codec
3) omit the section

I'll see if someone on the Java team wants to weigh in on what their thoughts are in an external review.

biniona-mongodb: After hearing your response, I agree that linking to a specific commit hash is not a good idea. I like the problems you have outlined. How to effectively utilize source code as a teaching tool could be a worthwhile conversation with DBX engineering teams. Maybe certain sections of the source are heavily commented and designed to be read as educational material? Not sure how to solve this. 

Resolving this comment as I am no longer requesting any changes.

terakilobyte (approved):
LGTM

source/fundamentals/data-formats/codecs.txt
biniona-mongodb: Sounds good, thanks for the response. Will try to avoid links to API docs in opening paragraphs in my writing going forward.

source/fundamentals/data-formats/codecs.txt
biniona-mongodb: Thanks for the response. I think those reasons make sense. Going forward, I will try to critically assess linking to examples introduced elsewhere in the document as opposed to linking to referenced examples as a rule.

source/fundamentals/data-formats/codecs.txt
biniona-mongodb: Sounds good. You are correct that we did not reach resolution [on point 4 of the Writing Hints and Tips for Drivers](https://docs.google.com/document/d/1SxlDvx4MYKBE6-LYGp5i5uuHfY1ZUYYlhc6syEdta5M/edit). I'll try and restart conversation on the doc.

Marking comment as resolved as no changes are requested.

biniona-mongodb (approved):

ccho-mongodb:
Hi @jyemin and @rozza , this PR is ready for your review.

I was also wondering if you could provide an opinion on whether mentioning that Document uses BsonTypeClassMap, and linking to DocumentCodecProvider and DocumentCodec in the source was worthwhile or if it's better to omit it: https://github.com/mongodb/docs-java/pull/65/files#diff-f3632dc4c3e534fc767d59725b984a96a9352c9ceddbfa768316c762fd155c80R268-R272

Thanks in advance!

jyemin:
@rozza can you take first crack at this one?

source/includes/fundamentals/code-snippets/PowerStatusCodec.java
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    @Override
+    public Class<String> getEncoderClass() {
+        return String.class;
+    }
+
+    @Override
+    public String decode(BsonReader reader, DecoderContext decoderContext) {
+        Boolean value = reader.readBoolean();
```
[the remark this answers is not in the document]
rozza: eg:

```
package fundamentals.monolightcodec;

import org.bson.BsonReader;
import org.bson.BsonType;
import org.bson.BsonWriter;
import org.bson.codecs.Codec;
import org.bson.codecs.DecoderContext;
import org.bson.codecs.EncoderContext;


// start class
public class PowerStatusCodec implements Codec<String> {

    @Override
    public void encode(BsonWriter writer, String value, EncoderContext encoderContext) {
        if (value.equals("on")) {
            writer.writeBoolean(Boolean.TRUE);;
        } else if (value.equals("off")) {
            writer.writeBoolean(Boolean.FALSE);
        } else {
            writer.writeNull();
        }
    }

    @Override
    public Class<String> getEncoderClass() {
        return String.class;
    }

    @Override
    public String decode(BsonReader reader, DecoderContext decoderContext) {
        BsonType bsonType = reader.readBsonType();
        switch (bsonType) {
            case NULL:
                reader.readNull();
                return null;
            case BOOLEAN:
                boolean value = reader.readBoolean();
               return value ? "on" : "off";
            default:
                throw new IllegalStateException("Unexpected BsonType: " + bsonType);
        }
    }
}
// end class
```

source/fundamentals/data-formats/codecs.txt
```diff
@@ -11,38 +11,340 @@ Codecs
...
+- :ref:`PowerStatusCodec <codecs-powerstatus-codec>`, our sample ``Codec``
+  that converts certain Java strings to BSON booleans.
+
+You can retrieve the ``Codec`` instances from the ``CodecRegistry`` instance
+from the prior example using the following code:
+
+.. code-block:: java
+   :copyable: true
+
+   Codec<String> powerStatus = codecRegistry.get(String.class);
```
rozza, line 164: Nit: Naming convention - `powerStatusCodec`

source/includes/fundamentals/code-snippets/MonolightCodecProvider.java
```diff
@@ -0,0 +1,35 @@
...
+
+    @Override
+    @SuppressWarnings("unchecked")
+    public <T> Codec<T> get(Class<T> clazz, CodecRegistry registry) {
+        if (clazz == Monolight.class) {
+            return (Codec<T>) new MonolightCodec(registry);
+        } else if (clazz == Integer.class) {
+            return (Codec<T>) integerCodec;
+        } else if (clazz == String.class) {
+            return (Codec<T>) powerStatusCodec;
```
rozza, line 28: This should be treated with caution - as it will handle *all* string values.  

So if a user was using the codec registry to decode a `Document` then any string values would encoded using the powerStatusCodec!

source/includes/fundamentals/code-snippets/MonolightCodec.java
```diff
@@ -0,0 +1,57 @@
...
+
+    @Override
+    public Monolight decode(BsonReader reader, DecoderContext decoderContext) {
+        Monolight monolight = new Monolight();
+
+        reader.readStartDocument();
+        while (reader.readBsonType() != BsonType.END_OF_DOCUMENT) {
+            String fieldName = reader.readName();
+            if (fieldName.equals("powerStatus")) {
+                monolight.setPowerStatus(powerStatusCodec.decode(reader, decoderContext));
```
rozza, line 40: I think although we have a `powerStatusCodec` this codec would handle its own encoding for such a simple task. But I understand that it could be a more complex type.

source/includes/fundamentals/code-snippets/MonolightCodecProvider.java
```diff
@@ -0,0 +1,35 @@
...
+        integerCodec = new IntegerCodec();
+        powerStatusCodec = new PowerStatusCodec();
+    }
+
+    @Override
+    @SuppressWarnings("unchecked")
+    public <T> Codec<T> get(Class<T> clazz, CodecRegistry registry) {
+        if (clazz == Monolight.class) {
+            return (Codec<T>) new MonolightCodec(registry);
+        } else if (clazz == Integer.class) {
```
rozza, line 25: Its an anti pattern for a provider to handle multiple types. Especially, if they have their own codecs.  In this scenario you would use `fromCodecs(new IntegerCodec(), new PowerStatusCodec());`

rozza (changes requested):
I think this looks good, however there is a trap with the `PowerStatusCodec`. It represents **_all_** strings it could quite easily cause a user to get in a pickle.

So what suggestions would I give a user to avoid this?

* You could use a special type an Enum would be great for the power status
* Have the `MonolightCodec` handle the encoding / decoding of its types.  It could use a type map like `Document` and `BsonDocument` does.

I probably would look towards using an Enum for the power status then it wouldn't trip anyone up.

source/includes/fundamentals/code-snippets/PowerStatusCodec.java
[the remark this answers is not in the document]
rozza: See my comment about why using `Codec<String>` could cause an issue here.

source/includes/fundamentals/code-snippets/MonolightCodecProvider.java
ccho-mongodb: I figured that would happen, but is there a convention that I can use to selectively encode strings? Would I need to create a conditional that checked the value of the string to determine whether to use a PowerStatusCodec or a StringCodec?

source/includes/fundamentals/code-snippets/MonolightCodecProvider.java
ccho-mongodb: Do you mean I should be instantiating a new CodecRegistry (using the factory method you provided) and then fetching the `IntegerCodec` or `PowerStatusCodec` from it?

source/includes/fundamentals/code-snippets/MonolightCodec.java
ccho-mongodb: If this doesn't seem necessary, I can rework the example to exclude the PowerStatusCodec. My intent was to build up the example by referencing the PowerStatusCodec presented in an earlier section, and demonstrate how you might use it to decode a field but it could be too contrived/simple. Let me know what you think.

source/includes/fundamentals/code-snippets/MonolightCodecProvider.java
rozza: The CodecRegistry is a global registry. So the convention would probably be for the MonolightCodec would handle the power status conversion internally.

source/includes/fundamentals/code-snippets/MonolightCodecProvider.java
rozza: Not quite, its just that a custom CodecProvider typically would only handle a single type and not multiple types as shown here.

source/includes/fundamentals/code-snippets/MonolightCodec.java
rozza: I think to fix there are two choices for a quick fix: 

1) Keep PowerStatusCodec, just don't register it in the global codec registry and instantiate it here.
2) Change the type for the field to an Enum, that way there will be no global registry clash.

ccho-mongodb:
Thanks for the review!

> * You could use a special type an Enum, which would be great for the power status
> ...
> I probably would look towards using an Enum for the power status then it wouldn't trip anyone up.

I wanted to avoid using an Enum since I would possibly contradict suggested usage of the EnumCodec here:
https://github.com/mongodb/docs-java/blob/master/source/fundamentals/data-formats/pojo-customization.txt#L647
However, if it seems too contrived and confusing to use it for pre-determined String values, I can rewrite this to use it.

> * Have the `MonolightCodec` handle the encoding / decoding of its types.  It could use a type map like `Document` and `BsonDocument` does.

Do you mean something like `BsonTypeCodecMap`? I guess I was trying to avoid demonstrating codecs with a minimal example implementation, but if it is typical to require a mapping, maybe I should switch to describing the `DocumentCodecProvider` similar to what's described in the current documentation here https://mongodb.github.io/mongo-java-driver/4.2/bson/codecs/

Let me know what you think about this and the questions I left in my responses. Thanks!

ccho-mongodb:
Hi @rozza, I think I've addressed the items you've pointed out, but could you take a look before I incorporate it into this PR and update the text around it? I've drafted the code in the following gist:

https://gist.github.com/ccho-mongodb/e571f45796c17cdbe0fad5209bb92e7b

Notable changes:
- I used an Enum instead of a String to represent the PowerStatus
- I registered the Codecs with what I think you're referring to as the global provider. I'm not that certain how to use them in a more local scope if not directly in the get() method of the Provider.
- The decode() method in PowerStatusCodec uses readBoolean rather than readBsonType() since I was getting an exception that the Codec was attempting to the value and not the type.
- I'm using the IntegerCodec rather than using the int32 read/write methods

Let me know what you think when you get the chance, thanks!

rozza:
The gist is looking good. Couple of notes:

```
MonolightCodecProvider monolightCodecProvider = new MonolightCodecProvider();
CodecRegistry MonolightFieldRegistry = CodecRegistries.fromCodecs(new IntegerCodec(), new PowerStatusCodec());

CodecRegistry codecRegistry = CodecRegistries.fromProviders(MonolightFieldRegistry, monolightCodecProvider);
```

Can be replaced by:

```
CodecRegistries.fromRegistries(CodecRegistries.fromProviders(monolightCodecProvider), 
                               CodecRegistries.fromCodecs(new IntegerCodec(), new PowerStatusCodec()),
                               MongoClientSettings.DEFAULT_CODEC_REGISTRY);
```

`MonolightCodecProvider.java` doesn't need the IntegerCodec & PowerStatusCodec.

`PowerStatus` encode, doesn't need to writeNull - as it can't be null its an enum and decode shouldn't return a null.  I would set a default value in the `Monolight` class so it always has a value. Or in MonolightCodec don't write the field if the value is null.

ccho-mongodb:
Thanks @rozza, I made updates based on your feedback. Please re-review when you get the chance, thanks!

source/includes/fundamentals/code-snippets/Monolight.java
```diff
@@ -0,0 +1,35 @@
+package fundamentals.monolightcodec;
+
+// start class
+public class Monolight {
+    private PowerStatus powerStatus = PowerStatus.OFF;
+    private Integer colorTemperature;
+
+    public Monolight() {}
+
+    // ...
```
rozza, line 10: Nit: Are these comments needed?

source/includes/fundamentals/code-snippets/PowerStatusCodec.java
```diff
@@ -0,0 +1,37 @@
...
+        if (value != null) {
+            writer.writeBoolean(value.equals(PowerStatus.ON) ? Boolean.TRUE : Boolean.FALSE);
+        }
+    }
+
+    @Override
+    public PowerStatus decode(BsonReader reader, DecoderContext decoderContext) {
+        Boolean value = reader.readBoolean();
+
+        if (value) {
```
rozza, line 24: Suggest change to: 

```
return reader.readBoolean() ? PowerStatus.ON : PowerStatus.OFF;
```

source/fundamentals/data-formats/codecs.txt
```diff
@@ -11,38 +11,360 @@ Codecs
...
+When working with POJOs, consider using the  ``PojoCodecProvider`` to
+minimize duplicate code to convert commonly-used data types and customize
+the behavior. See our
+:doc:`POJO Customization guide </fundamentals/data-formats/pojo-customization>`
+for more information.
+
+For more information about the classes and interfaces in this section, see the
+:java-docs:`CodecProvider API documentation <apidocs/bson/org/bson/codecs/configuration/CodecProvider.html>`.
 
 BsonTypeClassMap
```
rozza, line 232: This section doesn't need to be nested in CodecProvider.

`BsonTypeClassMap` is really an implementation detail for Codec's that contain container classes like `Document`, `Map` & `Iterables` and its only used for decoding of types from Bson into these container types.  Perhaps add a clarification at the start?

rozza (changes requested):
Couple of minor nits/suggestion/comments

Very nearly there now :)

source/includes/fundamentals/code-snippets/Monolight.java
ccho-mongodb: They're there for formatting the example. We use a directive called "literalinclude" which inserts all the text from a file, optionally between two comments. The "start class" and "end class" denote the start and end of the included text, and the "// ..." signifies additional but excluded code.
In the documentation example, I only wanted to show the fields and constructor, not all the getters/setters/toString methods. See the attached to see how this renders on the page:

Screen Shot 2021-04-30 at 10 10 23 AM

Let me know if you have any recommended changes here.

source/fundamentals/data-formats/codecs.txt
ccho-mongodb: Good to know, updating the section.

ccho-mongodb:
@rozza ready for another look when you get the chance.

rozza (approved):
:+1: LGTM!

PR #66, opened 2021-03-24
DOCSP-15135 crud read sort update

biniona-mongodb:
## Pull Request Info
Migrate Fundamentals > Builders > Sorts page from  https://jira.mongodb.org/browse/DOCSP-9585 into Fundamentals > CRUD > Read > Sort and redo builders section to more explicitly focus on `Sorts` classs. 

### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-15135

https://docs-mongodbcom-staging.corp.mongodb.com/d9aff57/java/docsworker-xlarge/DOCSP-15135-CRUD-Read-Sort-Update/fundamentals/crud/read-operations/sort/

https://docs-mongodbcom-staging.corp.mongodb.com/d9aff57/java/docsworker-xlarge/DOCSP-15135-CRUD-Read-Sort-Update/fundamentals/builders/sort/

### Self-Review Checklist

### DIFF
--- a/source/fundamentals/builders/sort.txt
+++ b/source/fundamentals/builders/sort.txt
@@ -47,7 +47,7 @@ driver, consider reading our
 The Sorts Class
 ---------------
 
-``Sorts`` class is a builder that provides static factory methods for all sort
+The ``Sorts`` class is a builder that provides static factory methods for all sort
 criteria operators supported by MongoDB. These methods return a ``Bson`` object
 that you can pass to the  ``sort()`` method of a ``FindIterable`` instance or to 
 ``Aggregates.sort()``. If you want to learn more about ``Aggregates``, see our 
@@ -170,7 +170,7 @@ Text Score
 
 To specify a sort by the text score of a text search, use the
 ``Sorts.metaTextScore()`` static factory method. For a detailed usage example
-showing how to specify a sort using ``Sorts.metaTextScore()``, see 
+showing how to specify sort criteria using ``Sorts.metaTextScore()``, see 
 :ref:`the text search section of our sorting fundamentals guide <sorts-crud-text-search>`.
 
 For more information, see the
--- a/source/fundamentals/builders/sort.txt
+++ b/source/fundamentals/builders/sort.txt
@@ -39,9 +39,9 @@ driver, consider reading our
 
 .. warning::
 
-   MongoDB does not provide guaranteed sort order in the event of ties, or
-   documents with identical values in the field used for sorting. For a detailed
-   discussion of ties see the
+   MongoDB does not provide guaranteed sort order in the event of **ties**. A
+   tie occurs when two or more documents contain identical values in the
+   field being used for a sort. For a detailed discussion of ties see the
    :ref:`handling ties section of our guide on sorting <sorts-crud-handling-ties>`
 
 The Sorts Class
--- a/source/fundamentals/builders/sort.txt
+++ b/source/fundamentals/builders/sort.txt
@@ -198,9 +198,11 @@ The output of the example above should look something like this:
 Text Score
 ----------
 
-To specify a sort by the text score of a text search, use the
-``Sorts.metaTextScore()`` static factory method. For a detailed usage example
-showing how to specify sort criteria using ``Sorts.metaTextScore()``, see 
+You can sort text search results by their text score, a value that indicates how
+closely a search result matches your search string. To specify a sort by the
+text score of a text search, use the ``Sorts.metaTextScore()`` static factory
+method. For a detailed example showing how to specify sort criteria using
+the ``Sorts.metaTextScore()`` method, see 
 :ref:`the text search section of our sorting fundamentals guide <sorts-crud-text-search>`.
 
 For more information, see the
--- a/source/fundamentals/builders/sort.txt
+++ b/source/fundamentals/builders/sort.txt
@@ -86,7 +86,7 @@ The following two code snippets are equivalent. However, we recommend you use th
    // Not Recommended
    collection.find().sort(new Document("_id", 1));
 
-The code snippets above both sort the documents in the
+Both code snippets above sort the documents in the
 :ref:`sort_example collection <sorts-crud-sort-example>`  
 in the following order: 
    
--- a/source/fundamentals/builders/sort.txt
+++ b/source/fundamentals/builders/sort.txt
@@ -132,8 +132,8 @@ The output of the example above should look something like this:
    {"_id": 4, "letter": "b", "food": "coffee with sugar"}
    ...
 
-Combining Criteria
-------------------
+Combining Sort Criteria
+-----------------------
 
 To combine sort criteria, use the ``Sorts.orderBy()`` static factory
 method. The ``orderBy()`` method builds sort criteria that apply passed 
--- a/source/fundamentals/builders/sort.txt
+++ b/source/fundamentals/builders/sort.txt
@@ -123,10 +123,7 @@ on the ``_id`` field:
    
    collection.find().sort(descending("_id"));
 
-
-The code snippet above returns the documents in the
-:ref:`sort_example collection <sorts-builders-sort-example>`  
-in the following order: 
+The output of the example above should look something like this: 
 
 .. code-block:: json
 
@@ -142,9 +139,9 @@ To combine sort criteria, use the ``Sorts.orderBy()`` static factory
 method. The ``orderBy()`` method builds sort criteria that apply passed 
 in sort criteria from left to right in the event of ties.
 
-In the following code snippet, we use the ``orderBy()`` method to combine a
-descending sort on the ``letter`` field with an ascending sort on the
-``_id`` field.
+The following example sorts the documents in the 
+:ref:`sort_example collection <sorts-builders-sort-example>` in descending order
+on the ``letter`` field and ascending order on the ``_id`` field:
 
 .. code-block:: java
 
@@ -157,9 +154,7 @@ descending sort on the ``letter`` field with an ascending sort on the
    Bson orderBySort = orderBy(descending("letter"), ascending("_id"));
    collection.find().sort(orderBySort);
 
-The code snippet above returns the documents in the
-:ref:`sort_example collection <sorts-builders-sort-example>`
-in the following order: 
+The output of the example above should look something like this: 
 
 .. code-block:: json
 
@@ -170,13 +165,13 @@ in the following order:
    {"_id": 3, "letter": "a", "food": "maple syrup"}
    {"_id": 5, "letter": "a", "food": "milk and cookies"}
 
-Meta Text Score
----------------
+Text Score
+----------
 
 To specify a sort by the text score of a text search, use the
-``Sorts.metaTextScore()`` static factory method. For a detailed example of
-specifying a sort using ``Sorts.metaTextScore()``, see 
-:ref:`the text search section of our sorting fundamentals guide <sorts-crud-text-search>`
+``Sorts.metaTextScore()`` static factory method. For a detailed usage example
+showing how to specify a sort using ``Sorts.metaTextScore()``, see 
+:ref:`the text search section of our sorting fundamentals guide <sorts-crud-text-search>`.
 
 For more information, see the
 :java-docs:`Sorts class API documentation </apidocs/mongodb-driver-core/com/mongodb/client/model/Sorts.html>`.
--- a/source/fundamentals/builders/sort.txt
+++ b/source/fundamentals/builders/sort.txt
@@ -171,7 +171,8 @@ in sort criteria from left to right in the event of ties.
 
 The following example sorts the documents in the 
 :ref:`sort_example collection <sorts-builders-sort-example>` in descending order
-on the ``letter`` field and ascending order on the ``_id`` field:
+on the ``letter`` field, and in the event of a tie, ascending order on the
+``_id`` field:
 
 .. code-block:: java
 
--- a/source/fundamentals/builders/sort.txt
+++ b/source/fundamentals/builders/sort.txt
@@ -84,7 +84,7 @@ The following two code snippets are equivalent, however we recommend you use the
    // <MongoCollection setup code here>
    
    //Not Recommended
-   collection.find().sort(new Document("_id", 1)));
+   collection.find().sort(new Document("_id", 1));
 
 The code snippets above both sort the documents in the
 :ref:`sort_example collection <sorts-crud-sort-example>`  
--- a/source/fundamentals/builders/sort.txt
+++ b/source/fundamentals/builders/sort.txt
@@ -76,7 +76,7 @@ The following two code snippets are equivalent, however we recommend you use the
 
    // <MongoCollection setup code here>
    
-   //Recommended
+   // Recommended
    collection.find().sort(ascending("_id"));
 
 .. code-block:: java
--- a/source/fundamentals/builders/sort.txt
+++ b/source/fundamentals/builders/sort.txt
@@ -60,36 +60,6 @@ that you can pass to the  ``sort()`` method of a ``FindIterable`` instance or to
 class, see our 
 :doc:`guide on the Aggregates builder </fundamentals/builders/aggregates>`.
 
-The following two code snippets are equivalent. However, we recommend you use the
-``Sorts`` builder to specify your sort criteria:
-
-.. code-block:: java
-
-   import static com.mongodb.client.model.Sorts.ascending;
-
-   // <MongoCollection setup code here>
-   
-   // Recommended
-   collection.find().sort(ascending("_id"));
-
-.. code-block:: java
-
-   // <MongoCollection setup code here>
-   
-   // Not Recommended
-   collection.find().sort(new Document("_id", 1));
-
-Both code snippets above sort the documents in the
-:ref:`sort_example collection <sorts-crud-sort-example>`  
-in the following order: 
-   
-.. code-block:: json
-
-   {"_id": 1, "letter": "c", "food": "coffee with milk"}
-   {"_id": 2, "letter": "a", "food": "donuts and coffee"}
-   {"_id": 3, "letter": "a", "food": "maple syrup"}
-   ...
-
 For more information about the classes and interfaces in this section, see the
 following API documentation:
    
--- a/source/fundamentals/builders/sort.txt
+++ b/source/fundamentals/builders/sort.txt
@@ -65,7 +65,8 @@ The Sorts Class
 The ``Sorts`` class is a builder that provides static factory methods for all sort
 criteria operators supported by MongoDB. These methods return a ``Bson`` object
 that you can pass to the  ``sort()`` method of a ``FindIterable`` instance or to 
-``Aggregates.sort()``. If you want to learn more about ``Aggregates``, see our 
+``Aggregates.sort()``. If you want to learn more about the ``Aggregates``
+class, see our 
 :doc:`guide on the Aggregates builder </fundamentals/builders/aggregates>`.
 
 The following two code snippets are equivalent. However, we recommend you use the
--- a/source/fundamentals/builders/sort.txt
+++ b/source/fundamentals/builders/sort.txt
@@ -36,14 +36,6 @@ If you want to learn the fundamentals of sorting in the MongoDB Java
 driver, consider reading our
 :doc:`guide on sorting </fundamentals/crud/read-operations/sort/>`.
 
-.. warning::
-
-   MongoDB does not guarantee sort order in the event of **ties**. A
-   tie occurs when two or more documents contain identical values in the
-   field being used to order your results. For a detailed discussion of ties, 
-   see the
-   :ref:`handling ties section of our guide on sorting <sorts-crud-handling-ties>`.
-
 .. _sorts-builders-sort-example:
 
 The examples on this page use a sample collection, ``sort_example``, that
--- a/source/fundamentals/builders/sort.txt
+++ b/source/fundamentals/builders/sort.txt
@@ -128,8 +128,9 @@ Combining Sort Criteria
 -----------------------
 
 To combine sort criteria, use the ``Sorts.orderBy()`` static factory
-method. The ``orderBy()`` method builds sort criteria that apply passed 
-in sort criteria from left to right in the event of ties.
+method. This method constructs an object containing an ordered list of sort
+criteria. When performing the sort, if the leftmost sort criteria results in a
+tie, the sort uses the next sort criteria in the list to determine the order.
 
 The following example sorts the documents in the 
 :ref:`sample collection <sorts-builders-sort-sample>` in descending order
--- a/source/fundamentals/builders/sort.txt
+++ b/source/fundamentals/builders/sort.txt
@@ -36,9 +36,9 @@ If you want to learn the fundamentals of sorting in the MongoDB Java
 driver, consider reading our
 :doc:`guide on sorting </fundamentals/crud/read-operations/sort/>`.
 
-.. _sorts-builders-sort-example:
+.. _sorts-builders-sort-sample:
 
-The examples on this page use a sample collection, ``sort_example``, that
+The examples on this page use a sample collection that
 contains the following documents:
 
 .. code-block:: json
@@ -77,7 +77,7 @@ factory method. Pass ``Sorts.ascending()``
 the name of the field you need to sort on.
 
 The following example sorts the documents in the 
-:ref:`sort_example collection <sorts-builders-sort-example>` by ascending order
+:ref:`sample collection <sorts-builders-sort-sample>` by ascending order
 on the ``_id`` field:
 
 .. code-block:: java
@@ -104,7 +104,7 @@ To specify a descending sort, use the ``Sorts.descending()`` static factory
 method. Pass ``Sorts.descending()`` the name of the field you need to sort on.
 
 The following example sorts the documents in the 
-:ref:`sort_example collection <sorts-builders-sort-example>` in descending order
+:ref:`sample collection <sorts-builders-sort-sample>` in descending order
 on the ``_id`` field:
 
 .. code-block:: java
@@ -132,7 +132,7 @@ method. The ``orderBy()`` method builds sort criteria that apply passed
 in sort criteria from left to right in the event of ties.
 
 The following example sorts the documents in the 
-:ref:`sort_example collection <sorts-builders-sort-example>` in descending order
+:ref:`sample collection <sorts-builders-sort-sample>` in descending order
 on the ``letter`` field, and in the event of a tie, ascending order on the
 ``_id`` field:
 
--- a/source/fundamentals/builders/sort.txt
+++ b/source/fundamentals/builders/sort.txt
@@ -39,7 +39,7 @@ driver, consider reading our
 
 .. warning::
 
-   MongoDB does not provide guaranteed sort order in the event of **ties**. A
+   MongoDB does not guarantee sort order in the event of **ties**. A
    tie occurs when two or more documents contain identical values in the
    field being used for a sort. For a detailed discussion of ties see the
    :ref:`handling ties section of our guide on sorting <sorts-crud-handling-ties>`.
--- a/source/fundamentals/builders/sort.txt
+++ b/source/fundamentals/builders/sort.txt
@@ -67,7 +67,7 @@ that you can pass to the  ``sort()`` method of a ``FindIterable`` instance or to
 ``Aggregates.sort()``. If you want to learn more about ``Aggregates``, see our 
 :doc:`guide on the Aggregates builder </fundamentals/builders/aggregates>`.
 
-The following two code snippets are equivalent, however we recommend you use the
+The following two code snippets are equivalent. However, we recommend you use the
 ``Sorts`` builder to specify your sort criteria:
 
 .. code-block:: java
--- a/source/fundamentals/builders/sort.txt
+++ b/source/fundamentals/builders/sort.txt
@@ -41,7 +41,7 @@ driver, consider reading our
 
    MongoDB does not guarantee sort order in the event of **ties**. A
    tie occurs when two or more documents contain identical values in the
-   field being used for a sort. For a detailed discussion of ties see the
+   field being used for a sort. For a detailed discussion of ties, see the
    :ref:`handling ties section of our guide on sorting <sorts-crud-handling-ties>`.
 
 .. _sorts-builders-sort-example:
--- a/source/fundamentals/builders/sort.txt
+++ b/source/fundamentals/builders/sort.txt
@@ -44,20 +44,10 @@ driver, consider reading our
    field being used for a sort. For a detailed discussion of ties see the
    :ref:`handling ties section of our guide on sorting <sorts-crud-handling-ties>`
 
-The Sorts Class
----------------
-
-The ``Sorts`` class is a builder that provides static factory methods for all sort
-criteria operators supported by MongoDB. These methods return a ``Bson`` object
-that you can pass to the  ``sort()`` method of a ``FindIterable`` instance or to 
-``Aggregates.sort()``. If you want to learn more about ``Aggregates``, see our 
-:doc:`guide on the Aggregates builder </fundamentals/builders/aggregates>`.
-
 .. _sorts-builders-sort-example:
 
-The following examples show you how to use the methods
-provided by the ``Sorts`` class to sort your queries. The examples use a
-sample collection, ``sort_example``, that contains the following documents:
+The examples on this page use a sample collection, ``sort_example``, that
+contains the following documents:
 
 .. code-block:: json
 
@@ -67,6 +57,15 @@ sample collection, ``sort_example``, that contains the following documents:
    {"_id": 5, "letter": "a", "food": "milk and cookies"}
    {"_id": 2, "letter": "a", "food": "donuts and coffee"}
    {"_id": 6, "letter": "c", "food": "maple donut"}
+
+The Sorts Class
+---------------
+
+The ``Sorts`` class is a builder that provides static factory methods for all sort
+criteria operators supported by MongoDB. These methods return a ``Bson`` object
+that you can pass to the  ``sort()`` method of a ``FindIterable`` instance or to 
+``Aggregates.sort()``. If you want to learn more about ``Aggregates``, see our 
+:doc:`guide on the Aggregates builder </fundamentals/builders/aggregates>`.
 
 For more information about the classes and interfaces in this section, see the
 following API documentation:
--- a/source/fundamentals/builders/sort.txt
+++ b/source/fundamentals/builders/sort.txt
@@ -67,6 +67,36 @@ that you can pass to the  ``sort()`` method of a ``FindIterable`` instance or to
 ``Aggregates.sort()``. If you want to learn more about ``Aggregates``, see our 
 :doc:`guide on the Aggregates builder </fundamentals/builders/aggregates>`.
 
+The following two code snippets are equivalent, however we recommend you use the
+``Sorts`` builder to specify your sort criteria:
+
+.. code-block:: java
+
+   import static com.mongodb.client.model.Sorts.ascending;
+
+   // <MongoCollection setup code here>
+   
+   //Recommended
+   collection.find().sort(ascending("_id"));
+
+.. code-block:: java
+
+   // <MongoCollection setup code here>
+   
+   //Not Recommended
+   collection.find().sort(new Document("_id", 1)));
+
+The code snippets above both sort the documents in the
+:ref:`sort_example collection <sorts-crud-sort-example>`  
+in the following order: 
+   
+.. code-block:: json
+
+   {"_id": 1, "letter": "c", "food": "coffee with milk"}
+   {"_id": 2, "letter": "a", "food": "donuts and coffee"}
+   {"_id": 3, "letter": "a", "food": "maple syrup"}
+   ...
+
 For more information about the classes and interfaces in this section, see the
 following API documentation:
    
--- a/source/fundamentals/builders/sort.txt
+++ b/source/fundamentals/builders/sort.txt
@@ -83,7 +83,7 @@ The following two code snippets are equivalent, however we recommend you use the
 
    // <MongoCollection setup code here>
    
-   //Not Recommended
+   // Not Recommended
    collection.find().sort(new Document("_id", 1));
 
 The code snippets above both sort the documents in the
--- a/source/fundamentals/builders/sort.txt
+++ b/source/fundamentals/builders/sort.txt
@@ -152,7 +152,7 @@ on the ``_id`` field:
    
    collection.find().sort(descending("_id"));
 
-The output of the example above should look something like this: 
+The above example should output something like this: 
 
 .. code-block:: json
 
--- a/source/fundamentals/builders/sort.txt
+++ b/source/fundamentals/builders/sort.txt
@@ -42,7 +42,7 @@ driver, consider reading our
    MongoDB does not provide guaranteed sort order in the event of **ties**. A
    tie occurs when two or more documents contain identical values in the
    field being used for a sort. For a detailed discussion of ties see the
-   :ref:`handling ties section of our guide on sorting <sorts-crud-handling-ties>`
+   :ref:`handling ties section of our guide on sorting <sorts-crud-handling-ties>`.
 
 .. _sorts-builders-sort-example:
 
--- a/source/fundamentals/builders/sort.txt
+++ b/source/fundamentals/builders/sort.txt
@@ -40,24 +40,18 @@ driver, consider reading our
 .. warning::
 
    MongoDB does not provide guaranteed sort order in the event of ties, or
-   documents with identical values on the field being sorted. For a detailed
+   documents with identical values in the field used for sorting. For a detailed
    discussion of ties see the
    :ref:`handling ties section of our guide on sorting <sorts-crud-handling-ties>`
 
 The Sorts Class
 ---------------
 
-The :java-docs:`Sorts </apidocs/mongodb-driver-core/com/mongodb/client/model/Sorts.html>`
-class is a builder that provides static factory methods for all sort criteria
-operators supported by MongoDB. These methods return a
-:java-docs:`Bson </apidocs/bson/org/bson/conversions/Bson.html>`
-object that you can pass to the 
-:java-docs:`sort() </apidocs/mongodb-driver-sync/com/mongodb/client/FindIterable.html#sort(org.bson.conversions.Bson)>`
-method of a ``FindIterable`` instance or to 
-:java-docs:`Aggregates.sort() </apidocs/mongodb-driver-core/com/mongodb/client/model/Aggregates.html#sort(org.bson.conversions.Bson)>`.
-If you want to learn more about ``Aggregates``, see our 
+``Sorts`` class is a builder that provides static factory methods for all sort
+criteria operators supported by MongoDB. These methods return a ``Bson`` object
+that you can pass to the  ``sort()`` method of a ``FindIterable`` instance or to 
+``Aggregates.sort()``. If you want to learn more about ``Aggregates``, see our 
 :doc:`guide on the Aggregates builder </fundamentals/builders/aggregates>`.
-
 
 .. _sorts-builders-sort-example:
 
@@ -73,6 +67,15 @@ sample collection, ``sort_example``, that contains the following documents:
    {"_id": 5, "letter": "a", "food": "milk and cookies"}
    {"_id": 2, "letter": "a", "food": "donuts and coffee"}
    {"_id": 6, "letter": "c", "food": "maple donut"}
+
+For more information about the classes and interfaces in this section, see the
+following API documentation:
+   
+- :java-docs:`Sorts </apidocs/mongodb-driver-core/com/mongodb/client/model/Sorts.html>` 
+- :java-docs:`Bson </apidocs/bson/org/bson/conversions/Bson.html>`
+- :java-docs:`FindIterable </apidocs/mongodb-driver-sync/com/mongodb/client/FindIterable.html>`
+- :java-docs:`Aggregates </apidocs/mongodb-driver-core/com/mongodb/client/model/Aggregates.html>`.
+
 
 Ascending
 ---------
--- a/source/fundamentals/builders/sort.txt
+++ b/source/fundamentals/builders/sort.txt
@@ -37,6 +37,13 @@ If you want to learn the fundamentals of sorting in the MongoDB Java
 driver, consider reading our
 :doc:`guide on sorting </fundamentals/crud/read-operations/sort/>`.
 
+.. warning::
+
+   MongoDB does not provide guaranteed sort order in the event of ties, or
+   documents with identical values on the field being sorted. For a detailed
+   discussion of ties see the
+   :ref:`handling ties section of our guide on sorting <sorts-crud-handling-ties>`
+
 The Sorts Class
 ---------------
 
@@ -130,7 +137,7 @@ Combining Criteria
 
 To combine sort criteria, use the ``Sorts.orderBy()`` static factory
 method. The ``orderBy()`` method builds sort criteria that apply passed 
-in sort criteria from left to right in the event of ties. 
+in sort criteria from left to right in the event of ties.
 
 In the following code snippet, we use the ``orderBy()`` method to combine a
 descending sort on the ``letter`` field with an ascending sort on the
@@ -165,8 +172,8 @@ Meta Text Score
 
 To specify a sort by the text score of a text search, use the
 ``Sorts.metaTextScore()`` static factory method. For a detailed example of
-specifying a sort using ``Sorts.metaTextScore()``, see our 
-:doc:`sorting fundamentals guide </fundamentals/crud/read-operations/sort/#text-search>`.
+specifying a sort using ``Sorts.metaTextScore()``, see 
+:ref:`the text search section of our sorting fundamentals guide <sorts-crud-text-search>`
 
 For more information, see the
 :java-docs:`Sorts class API documentation </apidocs/mongodb-driver-core/com/mongodb/client/model/Sorts.html>`.
--- a/source/fundamentals/builders/sort.txt
+++ b/source/fundamentals/builders/sort.txt
@@ -41,7 +41,8 @@ driver, consider reading our
 
    MongoDB does not guarantee sort order in the event of **ties**. A
    tie occurs when two or more documents contain identical values in the
-   field being used for a sort. For a detailed discussion of ties, see the
+   field being used to order your results. For a detailed discussion of ties, 
+   see the
    :ref:`handling ties section of our guide on sorting <sorts-crud-handling-ties>`.
 
 .. _sorts-builders-sort-example:
--- a/source/fundamentals/builders/sort.txt
+++ b/source/fundamentals/builders/sort.txt
@@ -29,9 +29,8 @@ Builders are classes provided by the MongoDB Java driver that help you construct
 :java-docs:`Bson <apidocs/bson/org/bson/conversions/Bson.html>` objects. 
 To learn more, see our :doc:`guide on builders </fundamentals/builders/>`.
 
-You should read this guide if you would like to:
-
-* Use builders to specify sort criteria for your queries.
+You should read this guide if you would like to use builders to specify sort
+criteria for your queries.
 
 If you want to learn the fundamentals of sorting in the MongoDB Java
 driver, consider reading our
--- a/source/fundamentals/crud/read-operations/sort.txt
+++ b/source/fundamentals/crud/read-operations/sort.txt
@@ -220,8 +220,9 @@ code:
    
    collection.find().sort(ascending("letter"));
 
-Since multiple documents contain the value "a" for the field on which we perform the sort,
-the following documents could be returned in any order:
+Since multiple documents that matched our query contain the value "a" for the
+field on which we perform the sort, the following documents could be returned
+in any order:
 
 .. code-block:: json
 
--- a/source/fundamentals/crud/read-operations/sort.txt
+++ b/source/fundamentals/crud/read-operations/sort.txt
@@ -230,9 +230,9 @@ document returned could be any of the following documents:
    {"_id": 5, "letter": "a", "food": "milk and cookies"}
    {"_id": 2, "letter": "a", "food": "donuts and coffee"}
 
-If you need guaranteed sort order for documents that
-have fields with identical values, you can specify additional fields to sort
-on in the event of a tie.
+If you need to guarantee a specific sort order for documents that have fields
+with identical values, you can specify additional fields to sort on in the event
+of a tie.
 
 We can specify an ascending sort on the ``letter`` field followed by the
 ``_id`` field as follows:
--- a/source/fundamentals/crud/read-operations/sort.txt
+++ b/source/fundamentals/crud/read-operations/sort.txt
@@ -262,8 +262,9 @@ Combining Sort Criteria
 ~~~~~~~~~~~~~~~~~~~~~~~
 
 To combine sort criteria, use the ``Sorts.orderBy()`` static factory
-method. The ``orderBy()`` method builds sort criteria that apply passed 
-in sort criteria from left to right in the event of ties. 
+method. This method constructs an object containing an ordered list of sort criteria. 
+When performing the sort, if the leftmost sort criteria results in a tie, the sort uses the
+next sort criteria in the list to determine the order.
 
 In the following code snippet, we use the ``orderBy()`` method to combine a
 descending sort on the ``letter`` field with an ascending sort on the
--- a/source/fundamentals/crud/read-operations/sort.txt
+++ b/source/fundamentals/crud/read-operations/sort.txt
@@ -302,7 +302,7 @@ Text Search
 You can specify the order of the results of a 
 :manual:`text search </text-search/>` by how closely they match your
 search string. Each of your search results has a
-:manual:`text score </reference/operator/aggregation/meta/#exp._S_meta>`, a
+:manual:`text score </reference/operator/aggregation/meta/#exp._S_meta>` which is a
 numerical value indicating how well that result matches your search. 
 Use the ``Sorts.metaTextScore()`` static factory method to build your sort
 criteria to sort by the text score.
--- a/source/fundamentals/crud/read-operations/sort.txt
+++ b/source/fundamentals/crud/read-operations/sort.txt
@@ -31,10 +31,10 @@ You should read this guide if you would like to:
 * Sort on the text score of a
   :manual:`text search </core/text-search-operators/>`.
 
-.. _sorts-crud-sort-example:
+.. _sorts-crud-sort-sample:
 
-The examples in this guide use a sample collection, ``sort_example``, that
-contains the following documents:
+The examples in this guide use a sample collection that contains the following
+documents:
   
 .. code-block:: json
   
@@ -79,7 +79,7 @@ as follows:
    collection.aggregate(Arrays.asList(Aggregates.sort(ascending("_id"))));
 
 The code snippets above both sort the documents in the
-:ref:`sort_example collection <sorts-crud-sort-example>` from smallest to
+:ref:`sample collection <sorts-crud-sort-sample>` from smallest to
 largest value of the ``_id`` field: 
 
 .. code-block:: json
@@ -148,7 +148,7 @@ documents in your collection, sorted from smallest to largest on the specified
 field name. 
 
 In the following code example, we use the ``ascending()`` method to sort the
-:ref:`sort_example collection <sorts-crud-sort-example>`  
+:ref:`sample collection <sorts-crud-sort-sample>`  
 by the ``_id`` field:
 
 .. code-block:: java
@@ -191,7 +191,7 @@ The following code snippet shows how to specify a descending sort on the
 
 
 The code snippet above returns the documents in the
-:ref:`sort_example collection <sorts-crud-sort-example>`  
+:ref:`sample collection <sorts-crud-sort-sample>`  
 in descending order: 
 
 .. code-block:: json
@@ -209,7 +209,7 @@ Handling Ties
 A tie occurs when two or more documents have a field with identical values.
 MongoDB does not guarantee sort order in the event of ties. For example, suppose
 we encounter a tie when applying a sort to the
-:ref:`sort_example collection <sorts-crud-sort-example>` using the following
+:ref:`sample collection <sorts-crud-sort-sample>` using the following
 code:
 
 .. code-block:: java
@@ -245,7 +245,7 @@ We can specify an ascending sort on the ``letter`` field followed by the
    collection.find().sort(ascending("letter", "_id"));
 
 The code snippet above returns the documents in the
-:ref:`sort_example collection <sorts-crud-sort-example>`  
+:ref:`sample collection <sorts-crud-sort-sample>`  
 in the following order: 
 
 .. code-block:: json
@@ -281,7 +281,7 @@ descending sort on the ``letter`` field with an ascending sort on the
    collection.find().sort(orderBySort);
 
 The code snippet above returns the documents in the
-:ref:`sort_example collection <sorts-crud-sort-example>`
+:ref:`sample collection <sorts-crud-sort-sample>`
 in the following order: 
 
 .. code-block:: json
@@ -315,14 +315,14 @@ criteria to sort by the text score.
 
 In the following code example, we show how you can use the
 ``Sorts.metaTextScore()`` method to sort the results of a text
-search on the :ref:`sort_example collection <sorts-crud-sort-example>`.
+search on the :ref:`sample collection <sorts-crud-sort-sample>`.
 The code example uses the :doc:`Filters </fundamentals/builders/filters>`,
 :doc:`Indexes </fundamentals/builders/indexes>`, and
 :doc:`Projections </fundamentals/builders/projections>` builders.
 The code example performs the following actions:
 
 #. Creates a text index for your
-   :ref:`sort_example collection <sorts-crud-sort-example>`
+   :ref:`sample collection <sorts-crud-sort-sample>`
    on the ``food`` field. If you call ``createIndex()`` specifying an index that
    already exists on the collection, the operation does not create a new index.
 #. Runs your text search for the phrase "maple donut".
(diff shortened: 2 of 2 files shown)


biniona-mongodb:
It could be confusing to the reader that "Handling Ties" and "Combining Sort Criteria" are separate sections. They overlap in what they cover. You can use orderBy to handle a tie -- the only difference is that you can specify disparate orders.

Sometimes there are multiple links going to the same place within a section which you probably don't need.
Similar to Kailie's comment, instead of featuring a section "The Sorts Class", just reference the Sort Builder page since that is what the class you're describing is. The disparities in the current description makes it difficult for the reader to relate the two.

Not sure if you meant to fill in this info, but it could use some explanation of FindIterable and Aggregates.sort() and connecting "Bson" documents with the Document Java class.

Try to keep the reader focused on the content of the section rather than linking all different places. I try to do this by referencing the key concepts at the bottom of the page and clearly identifying them.

The text search example description isn't that easy to relate to the code example that follows.  Maybe you can split it into smaller pieces? For example, you wouldn't be creating an index every time you wanted to run the search.

source/fundamentals/crud/read-operations/sort.txt
```diff
@@ -4,135 +4,396 @@ Sort Results
...
+   {"_id": 2, "letter": "a", "food": "donuts and coffee"}
+   {"_id": 6, "letter": "c", "food": "maple donut"}
+
+Methods For Sorting
+-------------------
+
+You can sort your results using the  ``sort()`` method of a ``FindIterable``
+instance or by using the  ``Aggregates.sort()`` static factory method in an
+aggregation pipeline. Both of these methods take objects that implement the
+``Bson`` interface and specify sort criteria as arguments. For information
```
kyuan-mongodb, line 53:
```suggestion
:ref:`Bson <bson>` interface and specify sort criteria as arguments.
```
The sentence after can be removed.

source/fundamentals/crud/read-operations/sort.txt
```diff
@@ -4,135 +4,396 @@ Sort Results
...
+
+.. _sorts-crud-find-iterable-example:
+
+You can use the ``sort()`` method of a ``FindIterable`` instance as follows:
+
+.. code-block:: java
+
+   // <MongoCollection setup code here>
+   
+   collection.find().sort(new Document("_id", 1)));
```
kyuan-mongodb, line 65: Suggestion: I think you should mention what 1 and -1 means.  The reader might not make the connection with your equivalent example.

source/fundamentals/crud/read-operations/sort.txt
```diff
@@ -4,135 +4,396 @@ Sort Results
...
+
+For more information on the benefits of using builders, see our
+:doc:`guide on builders </fundamentals/builders/>`. For more information on the
+``Sorts`` builder class, see our
+:doc:`guide on the Sorts builder </fundamentals/builders/sort/>`.    
+
+For more information about the classes and interfaces in this section, see the
+following API documentation:
+
+- :java-docs:`FindIterable </apidocs/mongodb-driver-sync/com/mongodb/client/FindIterable.html>`
```
kyuan-mongodb, line 114: Suggestion: Instead of having a bulleted list, you can link the java-docs where the class/interface is mentioned.  This avoids having to put it again at the bottom.

source/fundamentals/builders/sort.txt
```diff
@@ -261,82 +165,13 @@ in the following order:
...
-:manual:`text score </reference/operator/aggregation/meta/#exp._S_meta>`, a
-numerical value indicating how well that result matches your search. 
-Use the ``Sorts.metaTextScore()`` static factory method to build your sort
-criteria to sort by the text score.
-
-.. warning:: Make Sure to Create a Text Index
-
-   You need a :manual:`text index </core/index-text/>` on your collection to perform a text search. See the server manual documentation for more
-   information on how to
-   :manual:`create a text index </core/index-text/#create-text-index>`.
```
kyuan-mongodb, line 279: Suggestion: I think "create a text index" should link to the builders > indexes page where it creates a text index

kyuan-mongodb (changes requested):
Overall, I think this is pretty good!  Just some suggestions to improve the pages.

source/fundamentals/builders/sort.txt
biniona-mongodb: Thank you for the note. The [Text Indexes Section](https://docs-mongodbcom-staging.corp.mongodb.com/86d2198/java/docsworker-xlarge/DOCSP-15135-CRUD-Read-Sort-Update/fundamentals/builders/indexes/#text-indexes) does not provide a description of what a text index is used for. My goal in linking to the server documentation is to help readers gain an understanding of how text indexes are used in case they are not familiar with them, and I do not believe the Text indexes section provides that information. Please let me know if you disagree with this assessment.

kyuan-mongodb: Agreed, but is there a reason you're linking to the same page twice?  I think the first sentence on its own is good enough in that case.

source/fundamentals/crud/read-operations/sort.txt
biniona-mongodb: This style was recommended in by Chris Cho [here](https://github.com/mongodb/docs-java/pull/65#discussion_r598887796) and [here](https://mongodb.slack.com/archives/GETCHDLCB/p1616618748074300?thread_ts=1616614407.070400&cid=GETCHDLCB) (see the second to the last bullet). That is the background for including the bulleted list. Please let me know if you disagree with this position, and if you do we should have a team meeting to reach consensus on style in regards to this issue.

source/fundamentals/crud/read-operations/sort.txt
biniona-mongodb: Thank you for the note. I disagree with this suggestion as the link would not provide context to what it is linking to, in addition the suggested link is not linking to the documentation of the BSON interface but to the Bson fundamentals page.

source/fundamentals/crud/read-operations/sort.txt
```diff
@@ -4,135 +4,396 @@ Sort Results
...
+   :depth: 2
+   :class: singlecol
+
+Overview
+--------
+
+In this guide, we show you how to **sort** your results.
+
+Sorting orders the documents returned from your query by your specified 
+**sort criteria**. Sort criteria are the rules you give MongoDB that describe
```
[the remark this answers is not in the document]
biniona-mongodb: Thank you for the note. I believe this suggestion is incorrect, as MongoDB performs the sort. This stance is supported by a technical reviewing having approved this phrasing in a previous review.

source/fundamentals/builders/sort.txt
```diff
@@ -71,86 +68,52 @@ sample collection, ``sort_example``, that contains the following documents:
    {"_id": 2, "letter": "a", "food": "donuts and coffee"}
    {"_id": 6, "letter": "c", "food": "maple donut"}
 
+For more information about the classes and interfaces in this section, see the
```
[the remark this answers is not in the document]
biniona-mongodb: Thank you for the note. This phrasing was modeled after Chris Cho's phrasing [here ](https://docs-mongodbcom-staging.corp.mongodb.com/398c0f5/java/docsworker-xlarge/DOCSP-12161-fundamentals-codecs/fundamentals/data-formats/codecs/#codec). See [here](https://github.com/mongodb/docs-java/pull/66#discussion_r601822352) for response to other comment.

source/fundamentals/builders/sort.txt
```diff
@@ -32,30 +32,27 @@ To learn more, see our :doc:`guide on builders </fundamentals/builders/>`.
...
-class is a builder that provides static factory methods for all sort criteria
-operators supported by MongoDB. These methods return a
-:java-docs:`Bson </apidocs/bson/org/bson/conversions/Bson.html>`
-object that you can pass to the 
-:java-docs:`sort() </apidocs/mongodb-driver-sync/com/mongodb/client/FindIterable.html#sort(org.bson.conversions.Bson)>`
-method of a ``FindIterable`` instance or to 
-:java-docs:`Aggregates.sort() </apidocs/mongodb-driver-core/com/mongodb/client/model/Aggregates.html#sort(org.bson.conversions.Bson)>`.
-If you want to learn more about ``Aggregates``, see our 
+The ``Sorts`` class is a builder that provides static factory methods for all sort
+criteria operators supported by MongoDB. These methods return a ``Bson`` object
```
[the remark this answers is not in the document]
biniona-mongodb: Thanks for the note. See https://github.com/mongodb/docs-java/pull/66#discussion_r601822352. Also, `Bson` object should link to the `Bson` interface API documentation.

source/fundamentals/builders/sort.txt
```diff
@@ -32,30 +32,27 @@ To learn more, see our :doc:`guide on builders </fundamentals/builders/>`.
...
-The :java-docs:`Sorts </apidocs/mongodb-driver-core/com/mongodb/client/model/Sorts.html>`
-class is a builder that provides static factory methods for all sort criteria
-operators supported by MongoDB. These methods return a
-:java-docs:`Bson </apidocs/bson/org/bson/conversions/Bson.html>`
-object that you can pass to the 
-:java-docs:`sort() </apidocs/mongodb-driver-sync/com/mongodb/client/FindIterable.html#sort(org.bson.conversions.Bson)>`
-method of a ``FindIterable`` instance or to 
-:java-docs:`Aggregates.sort() </apidocs/mongodb-driver-core/com/mongodb/client/model/Aggregates.html#sort(org.bson.conversions.Bson)>`.
-If you want to learn more about ``Aggregates``, see our 
+The ``Sorts`` class is a builder that provides static factory methods for all sort
```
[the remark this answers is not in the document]
biniona-mongodb: See https://github.com/mongodb/docs-java/pull/66#discussion_r601822352

source/fundamentals/builders/sort.txt
```diff
@@ -32,30 +32,27 @@ To learn more, see our :doc:`guide on builders </fundamentals/builders/>`.
 You should read this guide if you would like to:
 
 * Use builders to specify sort criteria for your queries.
-* Perform ascending sorts and descending sorts.
-* Combine sort criteria.
-* Sort on the text score of a
-  :manual:`text search </core/text-search-operators/>`.
 
 If you want to learn the fundamentals of sorting in the MongoDB Java
```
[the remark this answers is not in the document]
biniona-mongodb: Thanks for the note. This language was chosen in a 1-1 with Chris. The reason for it was that Overview sections should use an imperative tone sparingly. This section should help a reader decide if this page is for them, "This page will show you Z. if you want X, read this article".

source/fundamentals/builders/sort.txt
```diff
@@ -32,30 +32,27 @@ To learn more, see our :doc:`guide on builders </fundamentals/builders/>`.
 You should read this guide if you would like to:
```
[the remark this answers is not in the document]
biniona-mongodb: Thanks for the note. In a meeting, me and Chris decided that overview sections will have the formal structure of defining key terms, laying out what the article will say using the phrasingn "you should read this guide if you...", and then providing an alternate location to go to if this is not the right place. If you disagree with this structure, we should have a team meeting to decide on we will structure overview paragraphs going forward.

source/fundamentals/builders/sort.txt
biniona-mongodb: One link describes what a text index is, the other describes how to create one. Although they are both on the same page, I think having both gives readers a way to understand what I am describing (what is a text index?) and how to proceed given the situation (how do I make one?).

source/fundamentals/builders/sort.txt
```diff
@@ -32,30 +32,27 @@ To learn more, see our :doc:`guide on builders </fundamentals/builders/>`.
...
+The ``Sorts`` class is a builder that provides static factory methods for all sort
+criteria operators supported by MongoDB. These methods return a ``Bson`` object
+that you can pass to the  ``sort()`` method of a ``FindIterable`` instance or to 
+``Aggregates.sort()``. If you want to learn more about ``Aggregates``, see our 
 :doc:`guide on the Aggregates builder </fundamentals/builders/aggregates>`.
 
-
 .. _sorts-builders-sort-example:
 
 The following examples show you how to use the methods
```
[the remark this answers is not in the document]
biniona-mongodb: Thank you for the note. After rereading, I agree this sentence is unhelpful. I have removed the above sentence, and moved the sample collection out of the `Sorts Class` section.

Fix: https://github.com/mongodb/docs-java/pull/66/commits/098201d7ed944bacdcdbd7ec5853384d0c55693e

source/fundamentals/crud/read-operations/sort.txt
biniona-mongodb: Thank you for the suggestion. Your suggestion raises a lot of interesting questions. I am not sure what the right call is here, as we are trying to push users towards the builders style for creating sort specifications. Not sure how in depth we should go into MQL? Should we explain a code snippet that we are actively trying to discourage users from using? Would be curious to see what other people on the team have to say to this point?

@ccho-mongodb, @terakilobyte, any thoughts? Should we explain patterns we are trying to have users avoid?

kyuan-mongodb (commented):
Looks good so far!

source/fundamentals/crud/read-operations/sort.txt
ccho-mongodb: I am trying to figure out what you (Alek) want to do since you're presenting both the MQL here and Sorts builder further down. If you want to keep both, I think Kailie's suggestion is a valid point. You could explain it in as little as a sentence and link back to the server manual for more info.

My preference is not to show both on this page because it extends the amount of explanation needed and can easily be missed because it requires vertical scrolling and is further separated by the Aggregates code example. The only page I might provide this comparison is on the Sort builders page to show equivalency.

biniona-mongodb: Thank you for the suggestion. After implementing the rewrite, I agree that the `Sorts` builder page is a more natural location for the equivalency example. Fix: https://github.com/mongodb/docs-java/pull/66/commits/a7c3b22a9f311aba0269920c9763ff42a08ee906

source/fundamentals/builders/sort.txt
```diff
@@ -32,35 +32,22 @@ To learn more, see our :doc:`guide on builders </fundamentals/builders/>`.
...
-:java-docs:`Bson </apidocs/bson/org/bson/conversions/Bson.html>`
-object that you can pass to the 
-:java-docs:`sort() </apidocs/mongodb-driver-sync/com/mongodb/client/FindIterable.html#sort(org.bson.conversions.Bson)>`
-method of a ``FindIterable`` instance or to 
-:java-docs:`Aggregates.sort() </apidocs/mongodb-driver-core/com/mongodb/client/model/Aggregates.html#sort(org.bson.conversions.Bson)>`.
-If you want to learn more about ``Aggregates``, see our 
-:doc:`guide on the Aggregates builder </fundamentals/builders/aggregates>`.
+.. warning::
 
+   MongoDB does not provide guaranteed sort order in the event of **ties**. A
```
cbush, line 42: You can often replace verb + past participle adjective with the verb:
```suggestion
   MongoDB does not guarantee sort order in the event of **ties**. A
```

source/fundamentals/builders/sort.txt
```diff
@@ -32,35 +32,22 @@ To learn more, see our :doc:`guide on builders </fundamentals/builders/>`.
...
+   tie occurs when two or more documents contain identical values in the
+   field being used for a sort. For a detailed discussion of ties see the
+   :ref:`handling ties section of our guide on sorting <sorts-crud-handling-ties>`
```
cbush, line 45:
```suggestion
   :ref:`handling ties section of our guide on sorting <sorts-crud-handling-ties>`.
```

source/fundamentals/builders/sort.txt
```diff
@@ -32,35 +32,22 @@ To learn more, see our :doc:`guide on builders </fundamentals/builders/>`.
...
-:java-docs:`sort() </apidocs/mongodb-driver-sync/com/mongodb/client/FindIterable.html#sort(org.bson.conversions.Bson)>`
-method of a ``FindIterable`` instance or to 
-:java-docs:`Aggregates.sort() </apidocs/mongodb-driver-core/com/mongodb/client/model/Aggregates.html#sort(org.bson.conversions.Bson)>`.
-If you want to learn more about ``Aggregates``, see our 
-:doc:`guide on the Aggregates builder </fundamentals/builders/aggregates>`.
+.. warning::
 
+   MongoDB does not provide guaranteed sort order in the event of **ties**. A
+   tie occurs when two or more documents contain identical values in the
+   field being used for a sort. For a detailed discussion of ties see the
```
cbush, line 44: "Field being used for a sort" feels a bit clunky. Maybe "in the sorted field"?

cbush, line 44:
```suggestion
   field being used for a sort. For a detailed discussion of ties, see the
```

source/fundamentals/builders/sort.txt
```diff
@@ -71,86 +58,91 @@ sample collection, ``sort_example``, that contains the following documents:
...
+criteria operators supported by MongoDB. These methods return a ``Bson`` object
+that you can pass to the  ``sort()`` method of a ``FindIterable`` instance or to 
+``Aggregates.sort()``. If you want to learn more about ``Aggregates``, see our 
+:doc:`guide on the Aggregates builder </fundamentals/builders/aggregates>`.
 
-The ``Sorts`` class provides methods for specifying the direction of your sort.
-The direction of your sort can either be **ascending** or **descending**.
-An ascending sort orders your results from smallest to largest. A
-descending sort orders your results from largest to smallest.
+The following two code snippets are equivalent, however we recommend you use the
```
cbush, line 70: Avoid joining two sentences with a comma:
```suggestion
The following two code snippets are equivalent. However, we recommend you use the
```

source/fundamentals/builders/sort.txt
```diff
@@ -71,86 +58,91 @@ sample collection, ``sort_example``, that contains the following documents:
...
+.. code-block:: java
 
-* Numbers: 1, 2, 3, 43, 43, 55, 120
-* Dates: 1990-03-10, 1995-01-01, 2005-10-30, 2005-12-21 
-* Words (ASCII): Banana, Dill, carrot, cucumber, hummus
+   import static com.mongodb.client.model.Sorts.ascending;
+
+   // <MongoCollection setup code here>
+   
+   //Recommended
```
cbush, line 79: Purely personal preference:
```suggestion
   // Recommended
```

source/fundamentals/builders/sort.txt
```diff
@@ -71,86 +58,91 @@ sample collection, ``sort_example``, that contains the following documents:
...
 
-Here are some examples of data sorted in descending order:
+.. code-block:: java
 
-* Numbers: 100, 30, 12, 12, 9, 3, 1
-* Dates: 2020-01-01, 1998-12-11, 1998-12-10, 1975-07-22 
-* Words (reverse ASCII): pear, grapes, apple, Cheese
+   // <MongoCollection setup code here>
+   
+   //Not Recommended
```
cbush, line 86:
```suggestion
   // Not Recommended
```

source/fundamentals/builders/sort.txt
```diff
@@ -71,86 +58,91 @@ sample collection, ``sort_example``, that contains the following documents:
...
+   collection.find().sort(new Document("_id", 1));
+
+The code snippets above both sort the documents in the
```
cbush, line 89: The "both" here feels a bit redundant. You could also say "Both code snippets..."
```suggestion
The code snippets above sort the documents in the
```

source/fundamentals/builders/sort.txt
```diff
@@ -160,10 +152,7 @@ The following code snippet shows how to specify a descending sort on the
    
    collection.find().sort(descending("_id"));
 
-
-The code snippet above returns the documents in the
-:ref:`sort_example collection <sorts-builders-sort-example>`  
-in the following order: 
+The output of the example above should look something like this:
```
cbush, line 155: You can always tidy up a sentence by moving the subject around and using verbs to convey the other relevant information:
```suggestion
The above example should output something like this: 
```

source/fundamentals/crud/read-operations/sort.txt
```diff
@@ -4,135 +4,382 @@ Sort Results
...
```
cbush, line 19: Watch out for dropping the preposition on indirect objects:
```suggestion
**sort criteria**. Sort criteria are the rules you give to MongoDB that describe
```

source/fundamentals/crud/read-operations/sort.txt
```diff
@@ -4,135 +4,382 @@ Sort Results
...
+how you would like your data ordered. Some examples of sort criteria are:
```
cbush, line 20:
```suggestion
how you would like your data to be ordered. Some examples of sort criteria are:
```
Or:
```suggestion
how you would like to order your data. Some examples of sort criteria are:
```

source/fundamentals/crud/read-operations/sort.txt
cbush: Maybe you could link to the javadoc for Bson here instead?

source/fundamentals/crud/read-operations/sort.txt
```diff
@@ -4,135 +4,382 @@ Sort Results
...
+In the above code snippets, we specify our sort criteria using the ``Sorts``
+builder class. While it is possible to specify sort criteria using any class
+that implements the ``Bson`` interface, we recommend that you specify sort
+criteria through the ``Sorts`` builder. For more information on the benefits of
+using builders, see our :doc:`guide on builders </fundamentals/builders/>`. For
+more information on the ``Sorts`` builder class, see our
+:doc:`guide on the Sorts builder </fundamentals/builders/sort/>`.    
+
+For more information about the classes and interfaces in this section, see the
+following API documentation:
```
cbush, line 98: This section actually does have a lot of "For more information,...". Maybe you could use a bulleted list or something like that to arrange this information.

cbush (approved):
Looks good. Some minor comments you might want to consider.

source/fundamentals/builders/sort.txt
biniona-mongodb: Thank you for pointing this out. Going forward, I need to figure out a system for catching these types of punctuation errors.

source/fundamentals/builders/sort.txt
biniona-mongodb: Thank you for the note.

Going forward, I will watch out for verb + past participle usage. Can this check be automated?

source/fundamentals/builders/sort.txt
biniona-mongodb: Thank you for pointing this out. Need to come up with a better system for grammar checking.

source/fundamentals/builders/sort.txt
biniona-mongodb: Thank you for the note. This is the style we have been using as well. I think I need to update my Intellij settings to automatically lead code comments with spaces.

source/fundamentals/crud/read-operations/sort.txt
biniona-mongodb: Thanks for the note Chris. After looking around online, it seems like there is mixed opinion on whether or not non-pronoun indirect objects require a preposition. [This site](https://www.chompchomp.com/terms/indirectobject.htm) only recommends prepositions for pronoun indirect objects. This seems like the [difference between a prepositional phrase and an indirect object](really-learn-english.com/english-direct-object-indirect-object-and-object-of-the-preposition.html#:~:text=How%20can%20you%20tell%20if,followed%20by%20the%20direct%20object.)? Not sure, would appreciate discussion on this so that going forward I do not make this mistake.

cbush: Not a big deal, but it reads a bit sloppy to me without when you have a direct object right there as well.

source/fundamentals/crud/read-operations/sort.txt
biniona-mongodb: Thank you for the note. Would you consider this a style edit or a grammar edit? Taking the suggestion, however both sentences sound correct to my ear.

source/fundamentals/crud/read-operations/sort.txt
biniona-mongodb: Sounds good. My ear doesn't pick up on the sloppiness, so I'll try and keep an eye out for direct object oddities going forward.

source/fundamentals/builders/sort.txt
biniona-mongodb: Thank you for the suggestion. I am not sure I agree with this edit, as using a period makes me read it as two separate thoughts (where in my head it is one thought "code snippets do the same thing but one is preferred"). Will commit suggestion, but would appreciate follow up discussion.

source/fundamentals/builders/sort.txt
biniona-mongodb: Thank you for the note.  I agree this phrasing could be improved. think I prefer "Both code snippets", as in my head it stresses equivalence more heavily.

```suggestion
Both code snippets above sort the documents in the
```

source/fundamentals/builders/sort.txt
biniona-mongodb: Thank you for the note. I agree that phrasing is much better. I think I got too into the habit of saying "above should look something like this". Will try to reuse text more carefully.

source/fundamentals/builders/sort.txt
biniona-mongodb: Thanks for the note. I agree my current language is clunky. "In the sorted field" sounds slightly misleading to me as I fear a user could mistake that to mean the "fields are being sorted" or "only some fields get sorted".  What about "field being used to order your results"?

Change: https://github.com/mongodb/docs-java/pull/66/commits/ce1fe60d3bd9e8acad5e6b37898d73f800747796

source/fundamentals/crud/read-operations/sort.txt
biniona-mongodb: Thanks for the note. I do not understand what change you are suggesting, [as a bulleted list follows line 98](https://github.com/mongodb/docs-java/pull/66/files/a5dd2bb61d540bf9526f7792274a21d70c91f313#diff-b4955d68d249252f5472f742c01958cda1df25f3541a79865dd4d3e61314e836R101). Are you saying that a different part of this section could benefit from being formatted as a bulleted list?

source/fundamentals/builders/sort.txt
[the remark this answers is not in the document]
ccho-mongodb: To be clear, I think both the suggestion and your current wording are technically imperative (i.e. giving the reader a command) language, so you're not reducing it either way. As you said, we want to use the overview section to help a reader decide where to go by presenting them with "if you want X, read [Y]." While most definitions of "imperative" direct you to its simplest form "Do X", if you dig further, you might stumble upon Kant's "hypothetical imperative":

> Hypothetical imperatives tell us how to act in order to achieve a specific goal and the commandment of reason applies only conditionally, e.g. "I must study to get a degree."

"See X to learn Y" and "If you want X, read [Y]" are equivalent, hypothetical imperatives.

source/fundamentals/builders/sort.txt
cbush: Sure, sounds good.

source/fundamentals/crud/read-operations/sort.txt
cbush: Note the above paragraph repeats "For more information..." several times. I'm wondering if there's a way to tidy it up a bit.

source/fundamentals/crud/read-operations/sort.txt
```diff
@@ -4,135 +4,382 @@ Sort Results
...
+   {"_id": 4, "letter": "b", "food": "coffee with sugar"}
+   {"_id": 1, "letter": "c", "food": "coffee with milk"}
+   {"_id": 6, "letter": "c", "food": "maple donut"}
+
+Combining Sort Criteria
+~~~~~~~~~~~~~~~~~~~~~~~
+
+To combine sort criteria, use the ``Sorts.orderBy()`` static factory
+method. The ``orderBy()`` method builds sort criteria that apply passed 
+in sort criteria from left to right in the event of ties.
```
ccho-mongodb, line 264: 1) The first two sentences refer to the method two different ways which might cause some disconnect for the reader. Instead, we can use a pronoun to avoid repeating the terminology.

2) By "passed in" (prepositional phrase), I think you meant "passed-in" (adjective). However, I don't think this is the greatest adjective to describe parameters. We can omit it and the sentence still communicates the same message.

3) "builds sort criteria that apply passed in sort criteria" doesn't make much sense to me. Are there two different types of sort criteria? How do I distinguish between them?

 
```suggestion
To combine sort criteria, use the ``Sorts.orderBy()`` static factory
method. This method constructs an object containing an ordered list of sort criteria. 
When performing the sort, if the leftmost sort criteria results in a tie, the sort uses the
next sort criteria in the list to determine the order.
```

source/fundamentals/crud/read-operations/sort.txt
```diff
@@ -4,135 +4,382 @@ Sort Results
...
+
+.. _sorts-crud-text-search:
+
+Text Search
+-----------
+
+You can specify the order of the results of a 
+:manual:`text search </text-search/>` by how closely they match your
+search string. Each of your search results has a
+:manual:`text score </reference/operator/aggregation/meta/#exp._S_meta>`, a
```
ccho-mongodb, line 302:
```suggestion
:manual:`text score </reference/operator/aggregation/meta/#exp._S_meta>` which is a
```

source/fundamentals/crud/read-operations/sort.txt
```diff
@@ -4,135 +4,382 @@ Sort Results
...
+numerical value indicating how well that result matches your search.
```
ccho-mongodb, line 303:
```suggestion
numerical value that indicates how closely that text value matches your search. 
```

source/fundamentals/crud/read-operations/sort.txt
```diff
@@ -4,135 +4,382 @@ Sort Results
...
+   {"_id": 3, "letter": "a", "food": "maple syrup", "score": 0.75}
+
+.. note:: MongoDB 4.4 or later ``$meta`` Behavior
+
+  When using MongoDB 4.4 or later, projecting ``Projections.metaTextScore()`` 
+  into your ``FindIterable`` instance is not necessary to sort on the text
+  score. In addition, MongoDB 4.4 or later disregards the field name you specify
+  in a ``$meta`` text score aggregation operation used in a sort. This means
+  that the field name argument you pass to ``Sorts.metaTextScore()`` is ignored
+  in MongoDB 4.4 or later.
```
ccho-mongodb, line 370: Could you find a way to explain this without repeating "MongoDB 4.4 or later" beyond the title and maybe the first sentence? It only seems to distract from the content.

source/fundamentals/crud/read-operations/sort.txt
```diff
@@ -4,135 +4,382 @@ Sort Results
...
+The code example uses the :doc:`Filters </fundamentals/builders/filters>`,
+:doc:`Indexes </fundamentals/builders/indexes>`, and
+:doc:`Projections </fundamentals/builders/projections>` builders.
+The code example performs the following actions:
+
+#. Creates a text index for your
+   :ref:`sort_example collection <sorts-crud-sort-example>`
+   on the ``food`` field. In a production application you would not
+   want to call the ``createIndex()`` method every search, but it is alright for
+   the purpose of this example.
```
ccho-mongodb, line 326: Let's omit this. There are plenty of other things we wouldn't do every search in our examples in a production environment, so it's not necessary to mention that. You could instead mention that no new index is created by this method if the index already exists.

ccho-mongodb (changes requested):

source/fundamentals/crud/read-operations/sort.txt
```diff
@@ -4,135 +4,382 @@ Sort Results
...
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+Overview
+--------
+
+In this guide, we show you how to **sort** your results.
+
+Sorting orders the documents returned from your query by your specified
```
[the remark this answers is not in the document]
biniona-mongodb: Thank you for the note Chris. On rereading, I agree that this changing of the form of the word is not beneficial for learning. Fix:

Going forward, I will try to keep the form of words consistent between usage and definition. A consistent form doesn't make readers have to do the extra work of translating between the two forms of the word.

source/fundamentals/crud/read-operations/sort.txt
```diff
@@ -4,135 +4,382 @@ Sort Results
...
+   :depth: 2
+   :class: singlecol
+
+Overview
+--------
+
+In this guide, we show you how to **sort** your results.
+
+Sorting orders the documents returned from your query by your specified 
+**sort criteria**. Sort criteria are the rules you give to MongoDB that describe
```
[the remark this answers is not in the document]
biniona-mongodb: Thank you for the note. I agree, I also think "pass to" is more commonly used. Committing suggested change.

Going forward, I will use "pass to" when discussing arguments, and will generally try to be more critical in selecting the best/most widely accepted term for an operation.

source/fundamentals/crud/read-operations/sort.txt
```diff
@@ -4,135 +4,382 @@ Sort Results
...
+   {"_id": 2, "letter": "a", "food": "donuts and coffee"}
+   {"_id": 6, "letter": "c", "food": "maple donut"}
+
+Methods For Sorting
+-------------------
+
+You can sort your results using the  ``sort()`` method of a ``FindIterable``
+instance or by using the  ``Aggregates.sort()`` static factory method in an
+aggregation pipeline. Both of these methods take objects that implement the
+``Bson`` interface and specify sort criteria as arguments. For information
```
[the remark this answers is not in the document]
biniona-mongodb: Thanks for the note. I agree this sentence is difficult to understand. I removed the "and specify sort criteria as arguments" , and changed "take" to "receive". The sentence now reads as follows, and it sounds better to my ear.

> Both of these methods receive objects that implement the
``Bson`` interface as arguments.

I think "specify sort criteria" made the sentence do too much work, and sort criteria are discussed later in the section. Let me know if you disagree.

Fix:

source/fundamentals/crud/read-operations/sort.txt
```diff
@@ -4,135 +4,382 @@ Sort Results
...
+on the ``Bson`` interface, see our 
+:doc:`guide on BSON </fundamentals/data-formats/document-data-format-bson/>`.
```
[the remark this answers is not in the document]
biniona-mongodb: Thanks for a note. I agree an introductory sentence here is beneficial. Fix: https://github.com/mongodb/docs-java/pull/66/commits/6768e8ae97456001b7755c82ba6610f2c8394673

Going forward I will try and notice when I am introducing a new concept, and provide adequate context for unfamiliar readers.

source/fundamentals/crud/read-operations/sort.txt
```diff
@@ -4,135 +4,382 @@ Sort Results
...
+
+The output of the code example above should look something like this: 
+
+.. code-block:: json
+
+   {"_id": 6, "letter": "c", "food": "maple donut", "score": 1.5}
+   {"_id": 2, "letter": "a", "food": "donuts and coffee", "score": 0.75}
+   {"_id": 3, "letter": "a", "food": "maple syrup", "score": 0.75}
+
+.. note:: MongoDB 4.4 or later ``$meta`` Behavior
```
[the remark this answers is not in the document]
biniona-mongodb: Thanks for the note. I agree, leading with version can be unhelpful and readers may glaze over text that has to do with version numbers.

Going forward, I will try to specify place behavior descriptions before version numbers such that readers better focus on what is being stressed in a section.

source/fundamentals/crud/read-operations/sort.txt
```diff
@@ -4,135 +4,382 @@ Sort Results
...
+   {"_id": 1, "letter": "c", "food": "coffee with milk"}
+   {"_id": 2, "letter": "a", "food": "donuts and coffee"}
+   {"_id": 3, "letter": "a", "food": "maple syrup"}
+   ...
+
+Descending
+~~~~~~~~~~
+
+To specify a descending sort, use the ``Sorts.descending()`` static factory
+method. Pass ``Sorts.descending()`` the name of the field you need to sort on.
```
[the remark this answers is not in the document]
biniona-mongodb: Thank you for the note. I agree this phrasing is more clear, and thank you for catching my not using our convention for referencing methods I.E. "The ``method()`` method".  I am committing your suggested changes, and applying similar edits to the ascending section here: https://github.com/mongodb/docs-java/pull/66/commits/03b8f04f01b0c06c0112f8e8fcd9a0ff7cd1d88f

source/fundamentals/crud/read-operations/sort.txt
```diff
@@ -4,135 +4,382 @@ Sort Results
...
-
-Call the ``sort()`` function and specify the order with ``ascending()`` to ensure that the
-operation returns books in alphabetical order by title:
+   import static com.mongodb.client.model.Sorts.ascending;
 
+   // <MongoCollection setup code here>
+   
+   collection.find().sort(ascending("_id"));
+ 
+You can use the ``Aggregates.sort()`` method as follows:
```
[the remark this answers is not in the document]
biniona-mongodb: Thanks for the note. I agree some context could be helpful. I am hesitant to discuss the `Aggregates` class at this stage of the article, as `Sorts` and builders more generally are discussed in the next paragraph. I have tried to add context by saying that `Aggregates.sort()` is "within an aggregation pipeline". Let me know if you think this solves the problem or is too subtle: https://github.com/mongodb/docs-java/pull/66/commits/6768e8ae97456001b7755c82ba6610f2c8394673

source/fundamentals/crud/read-operations/sort.txt
```diff
@@ -4,135 +4,382 @@ Sort Results
...
+   import com.mongodb.client.model.Aggregates;
+   import static com.mongodb.client.model.Sorts.ascending;
+
+   // <MongoCollection setup code here>
+   
+   collection.aggregate(Arrays.asList(Aggregates.sort(ascending("_id"))));
+
+The code snippets above both sort the documents in the
+:ref:`sort_example collection <sorts-crud-sort-example>`  
+in the following order:
```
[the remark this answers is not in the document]
biniona-mongodb: Thanks for the note. I didn't think to do this, but I agree at this early stage of the page I think this reinforcement makes a lot of sense. Fix: https://github.com/mongodb/docs-java/pull/66/commits/accd64cb458ec403b8bcf31d8713dcda533ad754

Going forward, I will look for opportunities to reinforce reader's understanding/expectations through meaningful transition text.

source/fundamentals/crud/read-operations/sort.txt
```diff
@@ -4,135 +4,382 @@ Sort Results
...
-This sample code prints your query results in ascending order by length. 
-When there are two documents with the same ``length`` value, it resolves the tie using the 
-``title`` field you specified as the second parameter.
+   import static com.mongodb.client.model.Sorts.ascending;
+
+   // <MongoCollection setup code here>
+   
+   Bson idSort = ascending("_id");
+   List<Document> results = new ArrayList<>();
+   collection.find().sort(idSort).into(results);
```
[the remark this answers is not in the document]
biniona-mongodb: Thanks for the note. Implementing the suggested edit. I think on my first pass I though the assignment was helpful for readers to connect builders to the `Bson` interface, but now I think the "Sorting Methods" section discusses that in enough depth that the extra variable assignment will not help readers. Fix: 

Going forward, I will try to maintain parallel structure between code snippets and code examples and will explain differences if  the two have divergent structure.

source/fundamentals/crud/read-operations/sort.txt
```diff
@@ -4,135 +4,382 @@ Sort Results
...
+.. code-block:: json
 
-In the returned documents, notice how the two books with the same 
-length - "Infinite Jest" and "A Dance with Dragons" -
-appear in the reverse of the order shown in the previous example:
+   {"_id": 3, "letter": "a", "food": "maple syrup"}
+   {"_id": 5, "letter": "a", "food": "milk and cookies"}
+   {"_id": 2, "letter": "a", "food": "donuts and coffee"}
+
+If you need guaranteed sort order for documents that
```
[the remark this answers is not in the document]
biniona-mongodb: Thanks for the note. While my ear doesn't hear a problem in either "guarenteed sort order" (noun) or "guarentee (verb) sort order", after reading your comment a few times I can see that there may be opportunity for more clarity if I avoid "guarenteed sort order" (noun) phrasing. Changed to 

>  If you need to guarantee a specific sort order

such that no additional terms are introduced. Fix:

source/fundamentals/builders/sort.txt
```diff
@@ -261,82 +195,13 @@ in the following order:
...
+----------
 
-  When using MongoDB 4.4 or later, projecting ``Projections.metaTextScore()`` 
-  into your ``FindIterable`` instance is not necessary to sort on the text
-  score. In addition, MongoDB 4.4 or later disregards the field name you specify
-  in a ``$meta`` text score aggregation operation used in a sort. This means
-  that the field name argument you pass to ``Sorts.metaTextScore()`` is ignored
-  in MongoDB 4.4 or later.
+To specify a sort by the text score of a text search, use the
+``Sorts.metaTextScore()`` static factory method. For a detailed usage example
```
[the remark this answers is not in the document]
biniona-mongodb: Thanks for the note. I changed this to just "for a detailed example". Going forward, I will try to avoid overloading terms in the documentation. Fix: https://github.com/mongodb/docs-java/pull/66/commits/4ca761f59685fc5ab3e5adc47d36cd702f3548bf

source/fundamentals/builders/sort.txt
```diff
@@ -172,70 +162,16 @@ in the following order:
...
 method. The ``orderBy()`` method builds sort criteria that apply passed 
-in sort criteria from left to right in the event of ties. 
+in sort criteria from left to right in the event of ties.
 
-In the following code snippet, we use the ``orderBy()`` method to combine a
-descending sort on the ``letter`` field with an ascending sort on the
-``_id`` field.
+The following example sorts the documents in the 
+:ref:`sort_example collection <sorts-builders-sort-example>` in descending order
+on the ``letter`` field and ascending order on the ``_id`` field:
```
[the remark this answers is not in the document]
biniona-mongodb: Thank you for the note. I agree adding "and in the event of a tie" makes the sentence more clear. Fix:

Going forward, I will try to improve my ability to hear these types of opportunities to improve the clarity or sentences.

source/fundamentals/builders/sort.txt
```diff
@@ -32,35 +32,23 @@ To learn more, see our :doc:`guide on builders </fundamentals/builders/>`.
...
-
-The :java-docs:`Sorts </apidocs/mongodb-driver-core/com/mongodb/client/model/Sorts.html>`
-class is a builder that provides static factory methods for all sort criteria
-operators supported by MongoDB. These methods return a
-:java-docs:`Bson </apidocs/bson/org/bson/conversions/Bson.html>`
-object that you can pass to the 
-:java-docs:`sort() </apidocs/mongodb-driver-sync/com/mongodb/client/FindIterable.html#sort(org.bson.conversions.Bson)>`
-method of a ``FindIterable`` instance or to 
-:java-docs:`Aggregates.sort() </apidocs/mongodb-driver-core/com/mongodb/client/model/Aggregates.html#sort(org.bson.conversions.Bson)>`.
-If you want to learn more about ``Aggregates``, see our
```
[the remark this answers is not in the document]
biniona-mongodb: Thank you for the note. I agree, `Aggregates` requires more context. Fix:

Going forward, I will try to refine this area of reader empathy, and not count on internal formatting conventions to effectively convey meaning.

source/fundamentals/builders/sort.txt
```diff
@@ -32,35 +32,23 @@ To learn more, see our :doc:`guide on builders </fundamentals/builders/>`.
 You should read this guide if you would like to:
 
 * Use builders to specify sort criteria for your queries.
```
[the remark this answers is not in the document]
biniona-mongodb: Thanks for the note. Fix: https://github.com/mongodb/docs-java/pull/66/commits/164972f20dc5d3c875fca859681da60ebad0a8de

Going forward, I will avoid single item lists. I will try to refine my ability to correctly use formatting conventions.

source/fundamentals/builders/sort.txt
```diff
@@ -32,35 +32,23 @@ To learn more, see our :doc:`guide on builders </fundamentals/builders/>`.
...
-:java-docs:`sort() </apidocs/mongodb-driver-sync/com/mongodb/client/FindIterable.html#sort(org.bson.conversions.Bson)>`
-method of a ``FindIterable`` instance or to 
-:java-docs:`Aggregates.sort() </apidocs/mongodb-driver-core/com/mongodb/client/model/Aggregates.html#sort(org.bson.conversions.Bson)>`.
-If you want to learn more about ``Aggregates``, see our 
-:doc:`guide on the Aggregates builder </fundamentals/builders/aggregates>`.
+.. warning::
 
+   MongoDB does not guarantee sort order in the event of **ties**. A
+   tie occurs when two or more documents contain identical values in the
+   field being used to order your results. For a detailed discussion of ties,
```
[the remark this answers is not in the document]
biniona-mongodb: Thanks for the note. I added the warning because tied sort behavior seemed to me the main thing about sorting that could trip up a developer in their app, and they have skimmed the intro and not be aware of the other article. 

This probably is assuming too much however, and focusing this guide around the builder and less around sorting is probably for the best. I will remove the warning. Fix:

source/fundamentals/crud/read-operations/sort.txt
```diff
@@ -4,135 +4,382 @@ Sort Results
...
-        .sort(orderBy(ascending("length"), descending("title")))
-        .iterator();
+   import static com.mongodb.client.model.Sorts.ascending;
+
+   // <MongoCollection setup code here>
+   
+   collection.find().sort(ascending("letter"));
+
+Since multiple documents contain "a" on the field we are sorting on, the first
+document returned could be any of the following documents:
```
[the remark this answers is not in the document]
biniona-mongodb: Thank you for the note. I agree this phrasing is significantly better. I will "on the field we are sorting on" and "document returned could be any of the following documents" are awkward and overloaded sentences. I will try and get better at spotting poor phrasing like this going forward. Committing suggestion.

source/fundamentals/crud/read-operations/sort.txt
biniona-mongodb: Thank you for the note. I had a lot of trouble with this introduction, and this phrasing is much better than what I had. I think my second sentence was doing too much work, and your breaking it into two sentences significantly improved clarity. 

Going forward, I will look for opportunities to use pronouns to  avoid repeating terminology when it does not cause problems with clarity. 

Going forward, I will try and avoid using the same terms in slightly different ways in the same sentence. A few of your notes have touched on my doing this in different places.

source/fundamentals/crud/read-operations/sort.txt
biniona-mongodb: Thanks for the note. Using "which" with no comma to make [ this a restrictive clause](https://ell.stackexchange.com/questions/134665/which-vs-that-and-comma-before-which) sounds better to my ear and seems to improve clarity. Committing suggestion.

source/fundamentals/crud/read-operations/sort.txt
biniona-mongodb:
```suggestion
numerical value that indicates how closely that result matches your search. 
```

Thanks for the note. I agree that "closely" is a better word than "well" here, and I agree that "indicating" should be split into "that indicates", however I disagree that "result" should be changed to "text value". The reason for this is that the text search is by document rather than by field, so I think it is more accurate to that the text score is saying how well a search matched an individual result.

source/fundamentals/builders/sort.txt
```diff
@@ -71,86 +59,91 @@ sample collection, ``sort_example``, that contains the following documents:
...
-* Dates: 2020-01-01, 1998-12-11, 1998-12-10, 1975-07-22 
-* Words (reverse ASCII): pear, grapes, apple, Cheese
+   // <MongoCollection setup code here>
+   
+   // Not Recommended
+   collection.find().sort(new Document("_id", 1));
+
+Both code snippets above sort the documents in the
+:ref:`sort_example collection <sorts-crud-sort-example>`  
+in the following order:
```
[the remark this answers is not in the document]
biniona-mongodb: Thanks for the note. I am removing the equivalency example as suggested. I think it is getting too out of scope to discuss `Bson` documents in any depth, and most likely isn't all that helpful to readers. Fix: https://github.com/mongodb/docs-java/pull/66/commits/a21b7a8b39f91a627142ef7d8bfda65b1ee8ace5

source/fundamentals/builders/sort.txt
```diff
@@ -71,86 +59,91 @@ sample collection, ``sort_example``, that contains the following documents:
...
-* Words (ASCII): Banana, Dill, carrot, cucumber, hummus
+   import static com.mongodb.client.model.Sorts.ascending;
+
+   // <MongoCollection setup code here>
+   
+   // Recommended
+   collection.find().sort(ascending("_id"));
 
-Here are some examples of data sorted in descending order:
+.. code-block:: java
```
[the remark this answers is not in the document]
biniona-mongodb: Thanks for the note. I have removed the equivalency example -- see https://github.com/mongodb/docs-java/pull/66#discussion_r602642042

source/fundamentals/crud/read-operations/sort.txt
biniona-mongodb: Thanks for the note.  Agree that mentioning the behavior of the `createIndex()` method is probably more helpful information.  Fix: https://github.com/mongodb/docs-java/pull/66/commits/17cb5967fd954884643786a57b977cb852607297

Going forward, I will try to better refine my sense for what is/isn't helpful context to include in an example.

source/fundamentals/crud/read-operations/sort.txt
biniona-mongodb: Thanks for the clarification Chris.B and Chris.C. In order to both keep the link to the sorts guide closer to its context and mitigate the clunkiness caused by too much repetition of the phrase "For more information", I removed the link to the Builders guide. This positions the sorts builders link immediately after the `Sorts` class is mentioned, and prevents a triple "For more information". Let me know if you find this solutions satisfactory. Fix: https://github.com/mongodb/docs-java/pull/66/commits/57da0467c871d74a2d0596b17988856df84478dc

source/fundamentals/crud/read-operations/sort.txt
biniona-mongodb: Thanks for the suggestion. In a discussion with Chris.C, I was told that as a general rule we are trying to avoid inline Java Doc links and instead group them at the end of the section. This discussions where this took place are linked to in https://github.com/mongodb/docs-java/pull/66#discussion_r601822352. I should have probably listed that as well in my earlier comment in this thread, apologies.

source/fundamentals/crud/read-operations/sort.txt
biniona-mongodb: Thank you for the note. I am having trouble cutting down "MongoDB 4.4 or later" usage under two times in the note as one time MongoDB 4.4 needs to be an object and the other time it needs to be the subject. One time the reader is using it, the other time the server itself is ignoring some input. I  agree that including "MongoDB 4.4 or later" multiple times is cumbersome, but I am not sure how to get it cleaner than this fix: https://github.com/mongodb/docs-java/pull/66/commits/13a9cd3f88619242918961478e8c9ec0ae5ee063

ccho-mongodb: How about introducing the paragraph with the condition and explaining that everything that follows is related to that condition? E.g. "The structure of text searches has changed for MongoDB 4.4 and later. When projecting ... In addition, the field name you specify in a $meta text score aggregation operation used in the sort is ignored..."

source/fundamentals/crud/read-operations/sort.txt
```diff
@@ -4,135 +4,384 @@ Sort Results
...
+
+.. code-block:: json
+
+   {"_id": 6, "letter": "c", "food": "maple donut", "score": 1.5}
+   {"_id": 2, "letter": "a", "food": "donuts and coffee", "score": 0.75}
+   {"_id": 3, "letter": "a", "food": "maple syrup", "score": 0.75}
+
+.. note:: ``$meta`` Behavior in MongoDB 4.4 or Later
+
+  If you are using MongoDB 4.4 or later, projecting
```
ccho-mongodb, line 367: This should be indented 3 spaces to line up with the "n" in "note" similar to what you have for the warning admonition you used earlier.

source/fundamentals/builders/sort.txt
```diff
@@ -71,86 +50,62 @@ sample collection, ``sort_example``, that contains the following documents:
...
-* Numbers: 100, 30, 12, 12, 9, 3, 1
-* Dates: 2020-01-01, 1998-12-11, 1998-12-10, 1975-07-22 
-* Words (reverse ASCII): pear, grapes, apple, Cheese
+For more information about the classes and interfaces in this section, see the
+following API documentation:
+   
+- :java-docs:`Sorts </apidocs/mongodb-driver-core/com/mongodb/client/model/Sorts.html>` 
+- :java-docs:`Bson </apidocs/bson/org/bson/conversions/Bson.html>`
+- :java-docs:`FindIterable </apidocs/mongodb-driver-sync/com/mongodb/client/FindIterable.html>`
+- :java-docs:`Aggregates </apidocs/mongodb-driver-core/com/mongodb/client/model/Aggregates.html>`.
```
ccho-mongodb, line 69:
```suggestion
- :java-docs:`Aggregates </apidocs/mongodb-driver-core/com/mongodb/client/model/Aggregates.html>`
```

source/fundamentals/builders/sort.txt
```diff
@@ -29,38 +29,17 @@ Builders are classes provided by the MongoDB Java driver that help you construct
...
-If you want to learn more about ``Aggregates``, see our 
-:doc:`guide on the Aggregates builder </fundamentals/builders/aggregates>`.
-
-
 .. _sorts-builders-sort-example:
 
-The following examples show you how to use the methods
-provided by the ``Sorts`` class to sort your queries. The examples use a
-sample collection, ``sort_example``, that contains the following documents:
+The examples on this page use a sample collection, ``sort_example``, that
```
ccho-mongodb, line 41: Do we ever use the collection name "sort_example" in any of the snippets? I wasn't able to find any instances on this page, so maybe we can omit the name and just refer to it as the example collection?

source/fundamentals/crud/read-operations/sort.txt
ccho-mongodb: > the text search is by document rather than by field
Isn't it by one or more fields as specified by the text index? https://docs.mongodb.com/manual/text-search/#text-index

source/fundamentals/crud/read-operations/sort.txt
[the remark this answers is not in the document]
ccho-mongodb: "gauranteed sort order" or "guarantee" by itself as a noun sounds like a feature that we offer, but we're trying to describe how the user can "guarantee" something by specifying the proper criteria in their sort fields.

source/fundamentals/crud/read-operations/sort.txt
biniona-mongodb: Yes, it is by one or more fields as specified by the text index. My choice of word "document" earlier was imprecise, apologies. I believe the word "result" better conveys the possible multi-field search happening in a `$text` operation better than "text value" which to me implies a single string value in a field. If you disagree with this assessment, I can change "result" to "text value", as both seem appropriate.

source/fundamentals/crud/read-operations/sort.txt
biniona-mongodb: Thanks for the note. I'm surprised this did not raise any errors (that I am aware of anyway). Going forward, I will make these checks manually. Fix: https://github.com/mongodb/docs-java/pull/66/commits/fc5d2ad9552aa5f7de81850a8090b66337683917

source/fundamentals/crud/read-operations/sort.txt
```diff
@@ -4,135 +4,384 @@ Sort Results
...
+
+#. Creates a text index for your
+   :ref:`sort_example collection <sorts-crud-sort-example>`
+   on the ``food`` field. If the index specified in your call to the
+   ``createIndex()`` method already exists, calling it will not create a new
+   index.
+#. Runs your text search for the phrase "maple donut".
+#. Projects text scores into your query results as the
+   ``score`` field. This projection is optional if your MongoDB instance is
+   running MongoDB 4.4 or later.
```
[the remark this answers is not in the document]
biniona-mongodb: Thanks for the note. After rereading, I can see how including this information disrupts the logical flow of information, and is largely unhelpful. Fix: https://github.com/mongodb/docs-java/pull/66/commits/37235712a517f331822f6146e58937dca900931b

source/fundamentals/builders/sort.txt
biniona-mongodb: Thanks for the note. You are right, it is not referred to. Fix: https://github.com/mongodb/docs-java/pull/66/commits/3273cb0926c61881026ca6d759d4520857076100

Going forward, I will try to be more aware of what needs an explicit name, and try to be more aware of when names are helpful/unhelpful.

source/fundamentals/crud/read-operations/sort.txt
```diff
@@ -4,135 +4,382 @@ Sort Results
...
+The following subsections show how to specify these sort criteria.
+
+Ascending
+~~~~~~~~~
+
+To specify an ascending sort, use the ``Sorts.ascending()`` static
+factory method. Pass the ``Sorts.ascending()`` method
+the name of the field you need to sort in ascending order.
+
+The ``ascending()`` method can be used as follows:
```
ccho-mongodb, line 136: For a moment, I thought I was looking at the builders section on Sorts and not how to use sort() because of the emphasis on using the ascending() method. I think this description needs to clearly distinguish between the functionality of sort() and ascending(). First, I'd suggest framing the code sample below as how to use the sort() method, using the ascending helper, to order the matched results.

source/fundamentals/crud/read-operations/sort.txt
```diff
@@ -4,135 +4,382 @@ Sort Results
...
 
-You can also sort by multiple fields in different orders, using ``Sorts.orderBy()``. 
-Consider this example where you first sort on ``length`` in an ascending order, and then on
-``title`` in a descending order:
+.. _sorts-crud-handling-ties:
+
+Handling Ties
+~~~~~~~~~~~~~
+
+A tie occurs when two or more documents have a field with identical values.
```
ccho-mongodb, line 209: I think it could be helpful for the reader if you further qualified the documents and field. First, the documents must be matching results on which the sort operates and second, the field must be one that determines the order. If that doesn't fit into a single sentence, maybe you can use a bulleted list to describe the details of each condition.

source/fundamentals/crud/read-operations/sort.txt
```diff
@@ -4,135 +4,382 @@ Sort Results
...
-    MongoCursor<Document> cursor = collection.find()
-        .sort(orderBy(ascending("length"), descending("title")))
-        .iterator();
+   import static com.mongodb.client.model.Sorts.ascending;
+
+   // <MongoCollection setup code here>
+   
+   collection.find().sort(ascending("letter"));
+
+Since multiple documents contain the value "a" for the field on which we perform the sort,
```
ccho-mongodb, line 223: Would it help for this to be more specific about what the "multiple documents" are? Since you're calling an operation, maybe they could be "documents that matched the query" or "results of the query" or other.

source/fundamentals/crud/read-operations/sort.txt
```diff
@@ -4,135 +4,382 @@ Sort Results
...
+
+Combining Sort Criteria
+~~~~~~~~~~~~~~~~~~~~~~~
+
+To combine sort criteria, use the ``Sorts.orderBy()`` static factory
+method. This method constructs an object containing an ordered list of sort
+criteria. When performing the sort, if the leftmost sort criteria results in a
+tie, the sort uses the next sort criteria in the list to determine the order.
+
+In the following code snippet, we use the ``orderBy()`` method to combine a
```
ccho-mongodb, line 268: Since you've described oerderBy() using terms like "ordered list" and "determine the order", the term "combine" used here could be confusing. Maybe you can use different terminology to explain that the sorts you are passing are considered in order?

ccho-mongodb (changes requested):
Just a few comments left, otherwise looks good.

source/fundamentals/crud/read-operations/sort.txt
biniona-mongodb: Thanks for the note. After rereading, I can see how this sentence can be clearer by specifying the  more about the necessary conditions to produce a tie. I have added a clause specifying that a tie is tied to the field being used to order your results. Fix:

Going forward, I will try to refine my senses for spotting these unclear/incomplete statements.

source/fundamentals/crud/read-operations/sort.txt
biniona-mongodb: Thanks for the note. I agree, "multiple documents" is unproductively vague. I have added clarifying language. fix: https://github.com/mongodb/docs-java/pull/66/commits/0d419400d1da8dcf44314d7a571718bf2e6fe068

Going forward, I will try to keep an eye for these types of unclear subjects. "Multiple Documents" is too abstract.

source/fundamentals/crud/read-operations/sort.txt
biniona-mongodb: Thanks for the note. I agree, introducing the term "combine" adds unnecessary complexity for learners. Fix: https://github.com/mongodb/docs-java/pull/66/commits/86eef3bcef3c19d3dfd6cdea3fb63f8069624710

Going forward, I will try to be mindful of introducing new terms in an unproductive way, and will try to build empathy for the challenges learners face in being confronted with a new term.

source/fundamentals/crud/read-operations/sort.txt
```diff
@@ -4,135 +4,384 @@ Sort Results
...
+Methods For Sorting
+-------------------
+
+You can sort results retrieved by a query and you can sort results 
+within an aggregation pipeline. To sort your query results, use the
+``sort()`` method of a ``FindIterable`` instance. To sort your results within an 
+aggregation pipeline, use the ``Aggregates.sort()`` static factory method. Both
+of these methods receive objects that implement the ``Bson`` interface as
+arguments. For information on the ``Bson`` interface, see our 
+:doc:`guide on BSON </fundamentals/data-formats/document-data-format-bson/>`.
```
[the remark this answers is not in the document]
biniona-mongodb: Thanks for the note. After making the switch, I agree changing this link to the API documentation probably provides more helpful information to the user. This relates to https://github.com/mongodb/docs-java/pull/66#discussion_r601800649, but I think is a slightly different suggestion as it is about the content of the link more so than an inline change. Fix: 

Going forward, I will try to be more critical about which external links provide the most helpful information.

source/fundamentals/crud/read-operations/sort.txt
biniona-mongodb: Thanks for the suggestion. I agree, changing this to be more explicitly about the `sort()` method is probably more helpful to learners. Fix: https://github.com/mongodb/docs-java/pull/66/commits/a53492f53a1540d0f44c053437612da8a373423d

Going forward, I will try to maintain the focus of the page and better align my writing to the expectations of readers.

ccho-mongodb (approved):
LGTM, resolved any unresolved conversations I could find, and left a couple nitpicks.

biniona-mongodb:
@jyemin can I have a review of this PR?

jyemin (commented):
I'm not sure what I'm looking at here. On the first page reference in the description, I see the sentence:

> The above sort() method returns a FindIterable object containing the documents in your collection, sorted from smallest to largest on the specified field name.

But I see that same sentence deleted in this PR.  Please advise.  Am I looking at the correct rendered view?

biniona-mongodb:
@jyemin sorry for the confusion. In this PR `Fundamentals > Builders > Sorts` page  has a large portion of its content moved to `Fundamentals > CRUD > Read > Sorts`. The sentence you are referring to was deleted on the `Builders > Sorts` page [here](https://github.com/mongodb/docs-java/pull/66/files#diff-ba4eb0892fe0d9ae6303bf2d13d2c12612eb49117994435ff4b87b3d50847e7eL115) and added to the `CRUD > Read > Sorts` page [here](https://github.com/mongodb/docs-java/pull/66/files#diff-b4955d68d249252f5472f742c01958cda1df25f3541a79865dd4d3e61314e836R147). Please let me know if you would like any additional clarification. 

I believe the staging links are up to date, as both are point to the "ee675c6" hash.

source/fundamentals/crud/read-operations/sort.txt
```diff
@@ -4,135 +4,385 @@ Sort Results
...
-the first field by which to sort and the ``title`` field as the second field. In the event 
-that the ``length`` field comparison results in a tie, the operation compares the tied results 
-using the ``title`` field:
+   import static com.mongodb.client.model.Sorts.ascending;
+
+   // <MongoCollection setup code here>
+
+   collection.find().sort(ascending("<field name>"));
 
+The above ``sort()`` method returns a ``FindIterable`` object containing the
```
jyemin, line 147: This could be misleading to a Java developer.  A Java developer reading that an `Iterable` contains a bunch of objects might be led to think that the objects are all in memory, when in fact the FindIterable is just a logical collection (that doesn't actually contain any documents at all in memory)  that should be streamed via a cursor.  I'm not sure how to rephrase it but maybe you can think of an alternative?  Or we can discuss on Slack.

biniona-mongodb: Thanks for the note Jeff. I agree that "containing" is misleading here. Thanks for catching this and helping me better understand cursors in MongoDB. Do you think this edit fixes the issue (I received help [on this edit from Chris Bush in this Slack thread](https://mongodb.slack.com/archives/GETCHDLCB/p1617377512125500)): https://github.com/mongodb/docs-java/pull/66/commits/d9aff574d889e5104c5dc23f6f7861a425be339e

jyemin: Yep!

jyemin (approved):
LGTM!

PR #72, opened 2021-04-05
DOCSP-9561 crud cursor

kyuan-mongodb:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-9561

https://docs-mongodbcom-staging.corp.mongodb.com/befcf60/java/docsworker-xlarge/DOCSP-9561-CRUDCursor/fundamentals/crud/read-operations/cursor/

### Self-Review Checklist

### DIFF
--- a/source/fundamentals/crud/read-operations/cursor.txt
+++ b/source/fundamentals/crud/read-operations/cursor.txt
@@ -6,7 +6,7 @@ Access Data From a Cursor
 
 Read operations that return multiple documents do not immediately return
 all values matching the query. Because a query can potentially match
-large sets of documents, we need to be able to access or store the
+large number of documents, we need to be able to access or store the
 matched documents. 
 
 This page uses an initiating method, ``find()`` to show how to access
@@ -15,9 +15,11 @@ data from a :java-docs:`FindIterable
 
 The ``find()`` method iterates through all the documents in your
 collection to find documents that match your query and returns an
-instance of a ``FindIterable``. A ``FindIterable`` consists of documents
-matched by your search criteria and allows you to further specify which
-documents to see by setting paramaters though methods. 
+instance of a ``FindIterable``. 
+
+A ``FindIterable`` consists of documents matched by your search criteria
+and allows you to further specify which documents to see by setting
+paramaters though methods. 
 
 Terminal Methods
 ----------------
@@ -39,7 +41,7 @@ method to retireve the first document in your query results:
    :end-before: end firstExample
 
 This method is often used when your query filter will match one
-document, as when filtering by a unique index. 
+document, such as when filtering by a unique index. 
 
 Into
 ~~~~
@@ -61,12 +63,14 @@ Cursor
 
 Use the :java-docs:`cursor() <apidocs/mongodb-driver-sync/com/mongodb/client/MongoIterable.html#cursor()>`
 method to iterate through fetched documents and ensure that the cursor
-closes if there is an early termination.
+closes if there is an early termination:
 
 .. code-block:: java
    :copyable: true
    
    MongoCursor<Document> cursor = collection.find().cursor();
+
+For more information on how to ensure a cursor closes, see the :ref:`cursor cleanup section <cursor_cleanup>`.
 
 Usage Patterns
 --------------
@@ -128,8 +132,8 @@ Close
 
 Use the
 :java-docs:`close() <apidocs/mongodb-driver-sync/com/mongodb/client/MongoCursor.html#close()>`
-method to free up a cursor's consumption of resources in both the client
-application and the MongoDB server:
+method in a finally block to free up a cursor's consumption of resources
+in both the client application and the MongoDB server:
 
 .. literalinclude:: /includes/fundamentals/code-snippets/Cursor.java
    :language: java
--- a/source/fundamentals/crud/read-operations/cursor.txt
+++ b/source/fundamentals/crud/read-operations/cursor.txt
@@ -7,24 +7,53 @@ Access Data From a Cursor
 Read operations that return multiple documents do not immediately return
 all values matching the query. Because a query can potentially match
 large sets of documents, these operations rely upon a
-:java-docs:`MongoCursor
-<apidocs/mongodb-driver-sync/com/mongodb/client/MongoCursor.html>` 
-and a :java-docs:`FindIterable
-<apidocs/mongodb-driver-sync/com/mongodb/client/FindIterable.html>` .
+:java-docs:`FindIterable
+<apidocs/mongodb-driver-sync/com/mongodb/client/FindIterable.html>`.
 
-A ``MongoCursor`` fetches documents in batches to reduce both memory
-consumption and network bandwidth usage. It can iterate through the
-fetched documents and ensures that the cursor closes if there is an
-early termination. 
-
-A ``FindIterable`` can also iterate through the fetched documents.
-It also allows you to further specify which documents you see by setting
-paramaters though methods. 
+A ``FindIterable`` iterates through all the documents in your collection
+to find documents that match your search criteria. It also allows you to
+further specify which documents you see by setting paramaters though methods. 
 
 This page uses an initiating method, ``find()`` to show how to access
-data from a ``MongoCursor`` and ``FindIterable``. 
+data from a ``FindIterable``. 
 
-The following shows the ``find()`` method as a cursor:
+Terminal Methods
+----------------
+
+Execute database operations on a ``FindIterable`` using terminal
+methods. Terminal methods return a single value after iterating thorugh
+all the results.
+
+First
+~~~~~
+
+Use the :java-docs:`first() <apidocs/mongodb-driver-sync/com/mongodb/client/MongoIterable.html#first()>`
+method to retireve the first document in your query results:
+
+.. literalinclude:: /includes/fundamentals/code-snippets/Cursor.java
+   :language: java
+   :dedent:
+   :start-after: begin firstExample
+   :end-before: end firstExample
+
+Into
+~~~~
+
+Use the :java-docs:`into() <apidocs/mongodb-driver-sync/com/mongodb/client/MongoIterable.html#into(A)>`
+method to store your query results in an ``ArrayList``: 
+
+.. literalinclude:: /includes/fundamentals/code-snippets/Cursor.java
+   :language: java
+   :dedent:
+   :start-after: begin intoExample
+   :end-before: end intoExample
+   
+Cursor
+~~~~~~
+
+Use the :java-docs:`cursor() <apidocs/mongodb-driver-sync/com/mongodb/client/MongoIterable.html#cursor()>`
+method to iterate through fetched documents and ensure that the cursor
+closes if there is an early termination.
 
 .. code-block:: java
    :copyable: true
@@ -43,7 +72,7 @@ Functional Iteration
 
 Pass a function to the 
 `forEach() <https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Iterable.html?is-external=true#forEach(java.util.function.Consumer)>`_ 
-method of any cursor to iterate through results in a functional style:
+method of a ``FindIterable`` to iterate through results in a functional style:
 
 .. literalinclude:: /includes/fundamentals/code-snippets/Cursor.java
    :language: java
@@ -81,37 +110,6 @@ method to retrieve the next available document from the cursor:
    :dedent:
    :start-after: begin manualIteration
    :end-before: end manualIteration
-
-
-Terminal Methods
-----------------
-
-Terminal methods perform on a ``FindIterable`` to execute database operations.
-
-First
-~~~~~
-
-Use the :java-docs:`first() <apidocs/mongodb-driver-sync/com/mongodb/client/MongoIterable.html#first()>`
-method to retireve the first document in your ``FindIterable``:
-
-.. literalinclude:: /includes/fundamentals/code-snippets/Cursor.java
-   :language: java
-   :dedent:
-   :start-after: begin firstExample
-   :end-before: end firstExample
-
-Into
-~~~~
-
-Use the :java-docs:`into() <apidocs/mongodb-driver-sync/com/mongodb/client/MongoIterable.html#into(A)>`
-method to iterate through all your documents and to store your query
-results in an ``ArrayList``: 
-
-.. literalinclude:: /includes/fundamentals/code-snippets/Cursor.java
-   :language: java
-   :dedent:
-   :start-after: begin intoExample
-   :end-before: end intoExample
 
 .. _cursor_cleanup:
 
--- a/source/fundamentals/crud/read-operations/cursor.txt
+++ b/source/fundamentals/crud/read-operations/cursor.txt
@@ -6,10 +6,10 @@ Access Data From a Cursor
 
 Read operations that return multiple documents do not immediately return
 all values matching the query. Because a query can potentially match
-large sets of documents, these operations rely upon an object called
-:java-docs:`MongoCursor </apidocs/mongodb-driver-sync/com/mongodb/client/MongoCursor.html>`. 
+large sets of documents, these operations rely upon an object called a
+:java-docs:`MongoCursor </apidocs/mongodb-driver-sync/com/mongodb/client/MongoCursor.html>`.
 
-A MongoCursor fetches documents in batches to reduce both memory
+A ``MongoCursor`` fetches documents in batches to reduce both memory
 consumption and network bandwidth usage. 
 
 The following functions use cursors internally, and return the results of the operations:
@@ -31,14 +31,14 @@ The following shows ``collection.find()`` as a cursor:
 Cursor Paradigms
 ----------------
 
-You can work with cursors using **cursor paradigms**.
-Most cursor paradigms allow you to access query results one document at
-a time, abstracting away network and caching logic. 
+You work with cursors using **cursor paradigms**. Most cursor paradigms
+allow you to access query results one document at a time, abstracting
+away network and caching logic. 
 
 .. warning::
 
    Do not combine different cursor paradigms on a single cursor.
-   Operations such as ``hasNext()`` and ``forEachRemaining()``
+   Operations such as ``next()`` and ``forEachRemaining()``
    each predictably modify the original cursor. If you mix these calls
    on a single cursor, you may receive unexpected results.
 
@@ -51,7 +51,7 @@ a time, abstracting away network and caching logic.
 For Each Functional Iteration
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-You can pass a function to the 
+Pass a function to the 
 `forEachRemaining() <https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Iterator.html?is-external=true#forEachRemaining(java.util.function.Consumer)>`_ 
 method of any cursor to iterate through results in a functional style:
 
@@ -64,7 +64,7 @@ method of any cursor to iterate through results in a functional style:
 Manual Iteration
 ~~~~~~~~~~~~~~~~
 
-You can use the :java-docs:`hasNext() </apidocs/mongodb-driver-sync/com/mongodb/client/MongoCursor.html#hasNext()>`
+Use the :java-docs:`hasNext() </apidocs/mongodb-driver-sync/com/mongodb/client/MongoCursor.html#hasNext()>`
 method to check if there are any documents available in the cursor, and then use the
 :java-docs:`next() </apidocs/mongodb-driver-sync/com/mongodb/client/MongoCursor.html#next()>`
 method to retrieve the next available document from the cursor:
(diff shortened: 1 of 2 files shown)


source/fundamentals/crud/read-operations/cursor.txt
```diff
@@ -5,4 +5,91 @@ Access Data From a Cursor
...
+----------------------
+
+Close
+~~~~~
+
+MongoCursors consume memory and network resources both in the client
+application and in the connected instance of MongoDB. Use
+:java-docs:`close() </apidocs/mongodb-driver-sync/com/mongodb/client/MongoCursor.html#close()>`
+to free up a cursor's resources in both the client application
+and the MongoDB server:
```
terakilobyte, line 88: suggestion: simplify

This paragraph feels awkward. In on sentence the words "both in" are used, and in the second sentence "in both" is used. Perhaps they can be combined?

> ``MongoCursor``s consume memory and network resources. Use ...

terakilobyte (changes requested):
Content is looking good but I think it can be cleaned up. I left some suggestions and comments.

Additionally, please look through the page and make sure you are formatting class names correctly.

source/fundamentals/crud/read-operations/cursor.txt
```diff
@@ -5,4 +5,91 @@ Access Data From a Cursor
...
+----------------
+
+You can work with cursors using **cursor paradigms**.
+Most cursor paradigms allow you to access query results one document at
+a time, abstracting away network and caching logic. 
+
+.. warning::
+
+   Do not combine different cursor paradigms on a single cursor.
+   Operations such as ``hasNext()`` and ``forEachRemaining()``
```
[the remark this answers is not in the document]
kyuan-mongodb: Good catch, you're right, it's ``next()`` that modifies

source/fundamentals/crud/read-operations/cursor.txt
```diff
@@ -5,4 +5,91 @@ Access Data From a Cursor
...
+   each predictably modify the original cursor. If you mix these calls
+   on a single cursor, you may receive unexpected results.
+
+.. note::
+
+   When you reach the last result through iteration, the cursor is
+   exhausted which means it ceases to respond to methods
```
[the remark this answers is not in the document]
kyuan-mongodb: According to this PR: https://github.com/mongodb/docs-node/pull/94, it ceases to respond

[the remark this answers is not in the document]
terakilobyte: But isn't this Java? That's a node PR

source/fundamentals/crud/read-operations/cursor.txt
```diff
@@ -5,4 +5,90 @@ Access Data From a Cursor
...
+:java-docs:`next() </apidocs/mongodb-driver-sync/com/mongodb/client/MongoCursor.html#next()>`
+method to retrieve the next available document from the cursor:
+
+.. literalinclude:: /includes/fundamentals/code-snippets/Cursor.java
+   :language: java
+   :dedent:
+   :start-after: begin manualIteration
+   :end-before: end manualIteration
+
+Cursor Utility Methods
```
terakilobyte, line 78: This heading feels out of place. Methods implies there's more than one, but close is the only one in the topic area. Maybe the heading could be "Close Cursors"?

terakilobyte (changes requested):
A few more things I noticed

source/fundamentals/crud/read-operations/cursor.txt
```diff
@@ -5,4 +5,90 @@ Access Data From a Cursor
...
+
+.. code-block:: java
+   :copyable: true
+   
+   collection.find().cursor();
+
+Cursor Paradigms
+----------------
+
+You work with cursors using **cursor paradigms**. Most cursor paradigms
```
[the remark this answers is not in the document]
kyuan-mongodb: I'll remove "most" since there's only 2 in the section

source/fundamentals/crud/read-operations/cursor.txt
```diff
@@ -5,4 +5,90 @@ Access Data From a Cursor
...
+allow you to access query results one document at a time, abstracting
+away network and caching logic. 
+
+.. warning::
+
+   Do not combine different cursor paradigms on a single cursor.
+   Operations such as ``next()`` and ``forEachRemaining()``
+   each predictably modify the original cursor. If you mix these calls
+   on a single cursor, you may receive unexpected results.
```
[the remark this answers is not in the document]
kyuan-mongodb: Just tested it and it is possible, but nothing out of the ordinary happens.  It prints all documents as expected and throws no error when the functional iteration runs.  I'll remove this since it doesn't apply here either.

source/fundamentals/crud/read-operations/cursor.txt
kyuan-mongodb: I feel that changing it's name wouldn't sound as nice, so I'll make the heading singular

terakilobyte: How about `Cursor Cleanup` since that is what it does?

terakilobyte (approved):
lgtm

kyuan-mongodb:
Hi @jyemin and @rozza , this PR is ready for review

source/fundamentals/crud/read-operations/cursor.txt
```diff
@@ -5,4 +5,77 @@ Access Data From a Cursor
...
+all values matching the query. Because a query can potentially match
+large sets of documents, these operations rely upon an object called a
+:java-docs:`MongoCursor </apidocs/mongodb-driver-sync/com/mongodb/client/MongoCursor.html>`.
+
+A ``MongoCursor`` fetches documents in batches to reduce both memory
+consumption and network bandwidth usage. 
+
+The following functions use cursors internally, and return the results of the operations:
+
+- ``collection.find()``
```
jyemin, line 17: There are others as well.  Perhaps don't present this as an exhaustive list.

source/fundamentals/crud/read-operations/cursor.txt
```diff
@@ -5,4 +5,77 @@ Access Data From a Cursor
...
+- ``collection.aggregate()``
+- ``collection.listIndexes()``
```
jyemin, line 19: These two in particular are not commonly used.

source/fundamentals/crud/read-operations/cursor.txt
```diff
@@ -5,4 +5,77 @@ Access Data From a Cursor
...
+- ``db.listCollectionNames()``
+
+Append ``.cursor()`` to use these functions as cursors.
```
jyemin, line 22: nit: They're called methods in java.

source/fundamentals/crud/read-operations/cursor.txt
```diff
@@ -5,4 +5,77 @@ Access Data From a Cursor
...
+Append ``.cursor()`` to use these functions as cursors.  
+
+The following shows ``collection.find()`` as a cursor:
+
+.. code-block:: java
+   :copyable: true
+   
+   collection.find().cursor();
+
+Cursor Paradigms
```
jyemin, line 31: I'm not sure these rise to the level of paradigms. Maybe just usage patterns?

source/fundamentals/crud/read-operations/cursor.txt
jyemin, line 22: This is a weird phrasing.  I think we need to point out a few things here:

* The initiating methods (find, aggregate) don't actually start any operations on the database.  It's not until the application calls one of the terminal methods that anything happens.
* The terminal methods include: cursor(), but also into() and first(). These latter two can be quite useful.
* The initiating methods implement the Iterable interface, so you can also use forEach to iterate.  This is generally used in preference to forEachRemaining on Iterator.
* The initiating methods implement the Iterable interface, so you can technically use the iterator method to iterate, but you should not do so because Iterator doesn't let you ensure that the underlying cursor is closed if for some reason you stop iterating before the underlying cursor is exhausted.  Instead, you should use the cursor() method, which returns a MongoCursor.  MongoCursor implements Iterable, but also Closeable.  So you can iterate the cursor, and also ensure that the cursor is closed in case of early termination (either in a finally block or a try-with-resources statement)

source/fundamentals/crud/read-operations/cursor.txt
```diff
@@ -5,4 +5,77 @@ Access Data From a Cursor
...
+
+Work with cursors using **cursor paradigms**. Cursor paradigms
+allow you to access query results one document at a time, abstracting
+away network and caching logic. 
+
+Functional Iteration
+~~~~~~~~~~~~~~~~~~~~~
+
+Pass a function to the 
+`forEachRemaining() <https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Iterator.html?is-external=true#forEachRemaining(java.util.function.Consumer)>`_
```
jyemin, line 42: See above about forEach method on Iterable

source/fundamentals/crud/read-operations/cursor.txt
```diff
@@ -5,4 +5,77 @@ Access Data From a Cursor
...
+.. literalinclude:: /includes/fundamentals/code-snippets/Cursor.java
+   :language: java
+   :dedent:
+   :start-after: begin manualIteration
+   :end-before: end manualIteration
+
+Cursor Cleanup
+--------------
+
+Close
```
jyemin, line 68: We should show examples of calling close according to Java idioms:

1. Using try-with-resources
2. Using a finally block

jyemin (changes requested):

source/fundamentals/crud/read-operations/cursor.txt
```diff
@@ -48,8 +44,21 @@ method of any cursor to iterate through results in a functional style:
...
+.. note::
+
+   The ``find()`` method implements the Iterable interface. We often see the
+   ``forEach()`` method used to iterate through results:
+
+   .. code-block:: java
+
+      collection.find().forEach(doc -> System.out.println(doc.toJson()));
+
+   Using this method is not preferable because you can't ensure that
```
jyemin, line 56: This actually isn't true.  The driver's Iterable implementation of forEach ensures the cursor is closed. We might, however, might want to warn against use of the enhanced for loop syntax:

    for (Document cur : collection.find()) {
         ..
    }

If there is an exception thrown before iteration completes, the cursor will _not_ be closed.

If you want to see the implementation, it's [here](https://github.com/mongodb/mongo-java-driver/blob/master/driver-sync/src/main/com/mongodb/client/internal/MongoIterableImpl.java#L120):

source/includes/fundamentals/code-snippets/Cursor.java
```diff
@@ -44,10 +44,20 @@ private void manualIteration(){
...
+    }
 
-        // begin close 
-        cursor.close();
-        // end close 
+    private void closeExample(){
+        // begin closeExample
+        MongoCursor<Document> cursor = collection.find().cursor();
+        
+        try{
```
jyemin, line 53: Nit: space before open curly brace.

jyemin, line 53: Do you want to show try-with-resources (introduced in Java 7)

    try (MongoCursor<Document> cursor = collection.find().cursor()) {
         while (cursor.hasNext()){
            System.out.println(cursor.next().toJson());
    }

jyemin (changes requested):
Looking better. Still would like you to address this comment: https://github.com/mongodb/docs-java/pull/72#discussion_r610595584

source/includes/fundamentals/code-snippets/Cursor.java
kyuan-mongodb: hm, doing this would mean the finally block wouldn't work to close the cursor.  Would I in this case put the close after the while loop in the try block?

jyemin: If you do that you don't need the finally block, or the call to close.  See https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html

source/fundamentals/crud/read-operations/cursor.txt
kyuan-mongodb: I believe I've addressed each bullet in my latest commit.  Please let me know if I've missed anything and I'll be sure to fix the page.

source/fundamentals/crud/read-operations/cursor.txt
```diff
@@ -5,4 +5,107 @@ Access Data From a Cursor
 .. default-domain:: mongodb
 
 Read operations that return multiple documents do not immediately return
-all values matching the query.
+all values matching the query. Because a query can potentially match
+large sets of documents, these operations rely upon an object called a
+:java-docs:`MongoCursor </apidocs/mongodb-driver-sync/com/mongodb/client/MongoCursor.html>`.
```
rozza, line 10: nit: This produces a url with a double `//` `https://mongodb.github.io/mongo-java-driver/4.2//apidocs/mongodb-driver-sync/com/mongodb/client/MongoCursor.html` .

Change to: `` `:java-docs:`MongoCursor <apidocs/mongodb-driver-sync/com/mongodb/client/MongoCursor.html>` ``

rozza (approved):
LGTM - one nit

source/fundamentals/crud/read-operations/cursor.txt
jyemin: I'm not seeing the first, second, or third bullet points addressed anywhere.

kyuan-mongodb: The first and second bullet, I mention them in lines 17 and 18.  The third bullet, I have in the warning box.

I didn't want to put your bullet word-for-word, so I took parts of each to incorporate into the page.  Would you like me to take more from each bullet?

jyemin: I'll go through them one by one:

* I think it's important to mention that it's not until a terminal method is executed that a database operation is performed.  If not on this page, then somewhere.
* I would like to see into() and first() described as well, with examples
* I still see forEachRemaining described, and not forEach.  I think forEach should be described, with an example, but not forEachRemaining
* This one is good.  It's the one with the warning box.

source/includes/fundamentals/code-snippets/Cursor.java
```diff
@@ -0,0 +1,113 @@
...
+        // end explicitCloseExample
+    }
+
+    private void closeExample(){
+        // begin implicitCloseExample
+        MongoCursor<Document> cursor = collection.find().cursor();
+        
+        try {
+            while (cursor.hasNext()){
+            System.out.println(cursor.next().toJson());
```
jyemin, line 92: Indent four spaces

source/fundamentals/crud/read-operations/cursor.txt
```diff
@@ -5,4 +5,140 @@ Access Data From a Cursor
...
+
+Use a `try-with-resources statement
+<https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html>`_
+to automatically free up a cursor's consumption of resources in both the
+client application and the MongoDB server:
+
+.. literalinclude:: /includes/fundamentals/code-snippets/Cursor.java
+   :language: java
+   :dedent:
+   :start-after: begin implicitCloseExample
```
jyemin, line 143: The example included here doesn't use try-with-resources statement anymore.  It looks like you deleted that example, the one that looked like:

   try(MongoCursor<Document> cursor = collection.find().cursor()) {
        // iterate
   }

source/includes/fundamentals/code-snippets/Cursor.java
```diff
@@ -0,0 +1,113 @@
...
+    private void intoExample(){
+        // begin intoExample
+        List<Document> results = new ArrayList<>();
+        FindIterable<Document> iterable = collection.find();
+        iterable.into(results);
+        System.out.println(results);
+        // end intoExample
+    }
+
+    private void manualIteration(){
```
jyemin, line 73: This example should use a finally block in which `close()` is called.  It's never correct to use a cursor as it is in this example, since the call to `close()` will be skipped if there's an exception

source/fundamentals/crud/read-operations/cursor.txt
```diff
@@ -5,4 +5,140 @@ Access Data From a Cursor
...
+
+Execute database operations on a ``FindIterable`` using terminal
+methods. Terminal methods return a single value after iterating thorugh
+all the documents in a collection.
+
+First
+~~~~~
+
+Use the :java-docs:`first() <apidocs/mongodb-driver-sync/com/mongodb/client/MongoIterable.html#first()>`
+method to retireve the first document in your query results:
```
jyemin, line 31: Consider adding something like:

"Most often this method is used when you know that the query filter will only match one document, as when filtering by `_id` or some other field with a unique index."

source/fundamentals/crud/read-operations/cursor.txt
```diff
@@ -5,4 +5,140 @@ Access Data From a Cursor
...
+further specify which documents to see by setting paramaters though methods. 
+
+This page uses an initiating method, ``find()`` to show how to access
+data from a ``FindIterable``. 
+
+Terminal Methods
+----------------
+
+Execute database operations on a ``FindIterable`` using terminal
+methods. Terminal methods return a single value after iterating thorugh
```
jyemin, line 24: thorugh -> through

source/fundamentals/crud/read-operations/cursor.txt
```diff
@@ -5,4 +5,140 @@ Access Data From a Cursor
...
+all the documents in a collection.
```
jyemin, line 25: This seems inaccurate, as `first` does not iterate through all the documents, just until it finds a match to the filter.

source/fundamentals/crud/read-operations/cursor.txt
```diff
@@ -5,4 +5,140 @@ Access Data From a Cursor
...
+   :language: java
+   :dedent:
+   :start-after: begin firstExample
+   :end-before: end firstExample
+
+Into
+~~~~
+
+Use the :java-docs:`into() <apidocs/mongodb-driver-sync/com/mongodb/client/MongoIterable.html#into(A)>`
+method to store your query results in an ``ArrayList``:
```
jyemin, line 43: ArrayList => List

source/fundamentals/crud/read-operations/cursor.txt
```diff
@@ -5,4 +5,140 @@ Access Data From a Cursor
...
+.. literalinclude:: /includes/fundamentals/code-snippets/Cursor.java
+   :language: java
+   :dedent:
+   :start-after: begin firstExample
+   :end-before: end firstExample
+
+Into
+~~~~
+
+Use the :java-docs:`into() <apidocs/mongodb-driver-sync/com/mongodb/client/MongoIterable.html#into(A)>`
```
jyemin, line 42: Consider adding something like:

"Most often this method is used when you know your query filter will return a small number of documents that will easily fit into available memory"

source/fundamentals/crud/read-operations/cursor.txt
jyemin, line 24: Not sure what is meant by `return a single value`.  I guess MongoCursor is a single value, but it's confusing to think of it that way since you have to further iterate through it.  Perhaps something like:

"Terminal methods execute the operation on the MongoDB server after all parameters controlling the operation have been configured on the `Iterable` instance.

source/fundamentals/crud/read-operations/cursor.txt
```diff
@@ -5,4 +5,140 @@ Access Data From a Cursor
 .. default-domain:: mongodb
 
 Read operations that return multiple documents do not immediately return
-all values matching the query.
+all values matching the query. Because a query can potentially match
+large sets of documents, these operations rely upon a
+:java-docs:`FindIterable
+<apidocs/mongodb-driver-sync/com/mongodb/client/FindIterable.html>`.
```
jyemin, line 11: It's going to be confusing if this page seems to be limited to using FindIterable.  I thought the idea was to make it a general page about iterating cursors, and use `find` as the example.  But everything here applies equally to the other methods that return instances of `MongoIterable`, e.g. aggregate, listCollections.

I think this needs to be generalized.

source/fundamentals/crud/read-operations/cursor.txt
```diff
@@ -5,4 +5,140 @@ Access Data From a Cursor
 .. default-domain:: mongodb
 
 Read operations that return multiple documents do not immediately return
-all values matching the query.
+all values matching the query. Because a query can potentially match
+large sets of documents, these operations rely upon a
```
jyemin, line 9: > Because a query can potentially match large sets of documents

Now that this page is also describing `first` and `into`, this language should be reconsidered.

jyemin (changes requested):
I thought we were getting close on this one, but it's turning out to be a tricky page to get right!

Let me know if you want to discuss any of this in the Slack channel.

source/fundamentals/crud/read-operations/cursor.txt
kyuan-mongodb: I originally wrote only about methods within MongoCursor since the page is called "Access Data From a Cursor".  But with the terminal methods added in, it shifted its focus to a FindIterable.  

I've generalized the introduction a bit more. Let me know what you think of the current edits.

jyemin: I like the edits.  Please consider adding a note indicating that everything below applies equally to other iterables, esp AggregateIterable

jyemin (approved):
One small addition to consider, but I'm going to LGTM as is.

Thanks for sticking with me on this one.  It was tricky, but it's a big improvement over what we have now on GitHub pages.  Someone actually just asked for this exact documentation in the #java Slack channel, and I had to just write the code for them there because there are no examples in the current docs!

PR #76, opened 2021-04-22
Docsp 9572 logging

biniona-mongodb:
## Pull Request Info
Adds Fundamentals > Logging Page

### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-9572

https://docs-mongodbcom-staging.corp.mongodb.com/4949a9a/java/docsworker-xlarge/DOCSP-9572-Logging/fundamentals/logging/

### Self-Review Checklist

### DIFF
--- a/source/fundamentals/logging.txt
+++ b/source/fundamentals/logging.txt
@@ -353,9 +353,9 @@ For a concrete example, this is what a logger hierarchy looks like in code.
 A logger inherits the properties of its ancestor logger and can define
 its own. You can think of this as similar to class inheritance in Java. 
 
-The makers of the MongoDB Java driver have defined the following logger names to
-organize different logging events in the driver. Here are the logger names
-defined in the driver and the logging events they correspond with. 
+The MongoDB Java driver defines the following logger names to organize different
+logging events in the driver. Here are the logger names defined in the driver
+and the logging events they correspond with. 
 
 * ``org.mongodb.driver.authenticator`` : authentication
 * ``org.mongodb.driver.client`` : events related to ``MongoClient`` instances
--- a/source/fundamentals/logging.txt
+++ b/source/fundamentals/logging.txt
@@ -58,8 +58,7 @@ can't find the ``slf4j-api`` package, the driver logs the following warning with
 
    WARNING: SLF4J not found on the classpath.  Logging is disabled for the 'org.mongodb.driver' component
 
-To set up a logger, you must include the following in your project's dependency
-list.
+To set up a logger, you must include the following in your project.
 
 * The ``slf4j-api`` package
 * a logging framework 
@@ -67,9 +66,10 @@ list.
 
 .. note:: 
 
-    The dependency resolution features of Gradle and Maven allow us to install
-    all three of these dependencies through a single package for some
-    logging frameworks. You will see this in the example below.
+    For the most popular logging frameworks, there is often a single binding
+    package that lists the ``slf4j-api`` and the logging framework as
+    dependencies. This means that you can set up a logger by adding one package
+    to your project's dependency list. You will see this in the example below.
 
 A binding is a piece of code that connects the ``slf4j-api`` package with a
 logging framework. The example below shows how to bind the ``slf4j-api`` package
--- a/source/fundamentals/logging.txt
+++ b/source/fundamentals/logging.txt
@@ -52,7 +52,7 @@ For more information on SLF4J,
 To set up a logger you must add the ``slf4j-api`` package, a logging framework, and a
 **binding** to your project's dependency list. A binding is a piece of code that
 connects the ``slf4j-api`` package with a logging framework. The example below shows
-how to bind the `slfj-api`to the two most popular logging frameworks, Log4j2 and
+how to bind the ``slfj-api`` package to the two most popular logging frameworks, Log4j2 and
 Logback.
 
 Setting up a logger is optional. When you start your application the MongoDB
@@ -101,8 +101,8 @@ tab corresponding to the logging framework you would like to use in your project
            <version>1.2.3</version>
          </dependency>
 
-      Once you have installed the above dependencies, try connecting to a
-      MongoDB instance and retrieving a document with the following code.
+      Once you have installed the above dependencies, connect to your
+      MongoDB instance and retrieve a document with the following code.
 
       .. code-block:: java
 
--- a/source/fundamentals/logging.txt
+++ b/source/fundamentals/logging.txt
@@ -380,10 +380,12 @@ Example - Logger Names
 ~~~~~~~~~~~~~~~~~~~~~~
 
 This example shows how to change the log level for a specific driver logger.
+We set the root logger to "OFF" and the  ``org.mongodb.connection`` logger to
+"INFO". This will cause the logger to only output messages related to connecting
+to a MongoDB instance. 
 
-In the following example, we set our root logger to "OFF" and the 
-``org.mongodb.connection`` logger to "INFO". This will cause your logger to only output
-messages related to connecting to your MongoDB instance. 
+Select the tab corresponding to the logging framework you are using in your
+project.
 
 .. tabs::
 
--- a/source/fundamentals/logging.txt
+++ b/source/fundamentals/logging.txt
@@ -439,6 +439,7 @@ project.
       Set your ``log4j2.xml`` file to the following.
 
       .. code-block:: xml
+         :emphasize-lines: 9
 
          <?xml version="1.0" encoding="UTF-8"?>
          <Configuration status="INFO">
--- a/source/fundamentals/logging.txt
+++ b/source/fundamentals/logging.txt
@@ -175,11 +175,8 @@ Configure Your Logger
 To configure your logger, you must configure the logging framework bound to
 SLF4J.
 
-The following subsection provides a concrete example showing how to configure
-your logger.
-
-Configure Your Log Level
-~~~~~~~~~~~~~~~~~~~~~~~~
+Example
+~~~~~~~
 
 In this subsection we show how to configure your logger's **log level**.
 
--- a/source/fundamentals/logging.txt
+++ b/source/fundamentals/logging.txt
@@ -254,22 +254,18 @@ project.
       .. code-block:: xml
 
          <configuration>
-         <appender name="CONSOLE"
-                   class="ch.qos.logback.core.ConsoleAppender">
-           <!-- deny all events with a level below INFO, that is TRACE and DEBUG -->
-           <filter class="ch.qos.logback.classic.filter.ThresholdFilter">
-             <level>INFO</level>
-           </filter>
-           <encoder>
-             <pattern>
-               %-4relative [%thread] %-5level %logger{30} - %msg%n
-             </pattern>
-           </encoder>
-         </appender>
-         <root level="DEBUG">
-           <appender-ref ref="CONSOLE" />
-         </root>
-         </configuration>
+            <appender name="CONSOLE"
+                     class="ch.qos.logback.core.ConsoleAppender">
+               <encoder>
+                  <pattern>
+                        %-4relative [%thread] %-5level %logger{30} - %msg%n
+                  </pattern>
+               </encoder>
+            </appender>
+            <root level="INFO">
+               <appender-ref ref="CONSOLE" />
+            </root>
+      </configuration>
 
       To test that your logger configuration was successful, run the following
       code.
@@ -283,9 +279,6 @@ project.
          ...
          1317 [cluster-ClusterId{value='<your cluster id>', description='null'}-<your connection uri>] INFO  org.mongodb.driver.cluster - Discovered replica set primary <your connection uri>
          1568 [main] INFO  org.mongodb.driver.connection - Opened connection [connectionId{localValue:7, serverValue:<server value>}] to <your connection uri>  
-
-      The above ``logback.xml`` file is a copy of the
-      `Threshold Filter Example in chapter seven of the Logback Manual <https://logback.qos.ch/manual/filters.html#thresholdFilter>`__.
 
       For more information on configuring Logback, see the 
       `the Logback Manual <https://logback.qos.ch/manual/configuration.html>`__.
--- a/source/fundamentals/logging.txt
+++ b/source/fundamentals/logging.txt
@@ -149,7 +149,8 @@ tab corresponding to the logging framework you would like to use in your project
            <version>2.14.1</version>
          </dependency>
 
-      Once you have installed the above dependencies, try logging an error.
+      Once you have installed the above dependencies, log an error using the
+      following code.
 
       .. code-block:: java
 
--- a/source/fundamentals/logging.txt
+++ b/source/fundamentals/logging.txt
@@ -174,7 +174,7 @@ Example
 In this subsection we show how to configure your logger's **log level**.
 
 A logger's log level specifies a lower bound for how urgent a message must be
-for the logger to output it.
+for the logger to output that message.
 
 This example shows how to configure your logger's log level to INFO.
 Select the tab corresponding to the logging framework you are using in your
--- a/source/fundamentals/logging.txt
+++ b/source/fundamentals/logging.txt
@@ -38,8 +38,8 @@ You will learn how to:
 Set Up a Logger
 ---------------
 
-In this section, we provide background on the dependencies necessary to set up a
-logger and walk through an example logger setup. 
+This section gives background on the dependencies necessary to set up a
+logger and provides an example logger setup. 
 
 Background
 ~~~~~~~~~~
--- a/source/fundamentals/logging.txt
+++ b/source/fundamentals/logging.txt
@@ -104,11 +104,10 @@ tab corresponding to the logging framework you would like to use in your project
          .. tab:: Maven
             :tabid: maven
                         
-            Add the following dependencies to your ``pom.xml`` file.
+            Add the following dependency to your ``pom.xml`` file.
    
             .. code-block:: xml
    
-               <!-- This package contains an SLF4J-Logback binding -->
                <dependency>
                  <groupId>ch.qos.logback</groupId>
                  <artifactId>logback-classic</artifactId>
@@ -118,7 +117,7 @@ tab corresponding to the logging framework you would like to use in your project
          .. tab:: Gradle
             :tabid: gradle
    
-            Add the following dependencies to your ``build.gradle`` file.
+            Add the following dependency to your ``build.gradle`` file.
    
             .. code-block:: none
    
@@ -126,7 +125,7 @@ tab corresponding to the logging framework you would like to use in your project
                   implementation 'ch.qos.logback:logback-classic:1.2.3'
                }
 
-      Once you have installed the above dependencies, connect to your
+      Once you have installed the above dependency, connect to your
       MongoDB instance and retrieve a document with the following code.
 
       .. include:: /includes/fundamentals/logging.rst
@@ -152,11 +151,10 @@ tab corresponding to the logging framework you would like to use in your project
          .. tab:: Maven
             :tabid: maven
                         
-            Add the following dependencies to your ``pom.xml`` file.
+            Add the following dependency to your ``pom.xml`` file.
    
             .. code-block:: xml
    
-               <!-- SLF4J-Log4j2 binding -->
                <dependency>
                  <groupId>org.apache.logging.log4j</groupId>
                  <artifactId>log4j-slf4j-impl</artifactId>
@@ -166,7 +164,7 @@ tab corresponding to the logging framework you would like to use in your project
          .. tab:: Gradle
             :tabid: gradle
    
-            Add the following dependencies to your ``build.gradle`` file.
+            Add the following dependency to your ``build.gradle`` file.
    
             .. code-block:: none
    
@@ -174,7 +172,7 @@ tab corresponding to the logging framework you would like to use in your project
                   implementation 'org.apache.logging.log4j:log4j-slf4j-impl:2.14.1'
                }
     
-      Once you have installed the above dependencies, log an error using the
+      Once you have installed the above dependency, log an error using the
       following code.
 
       .. code-block:: java
@@ -354,10 +352,6 @@ For a concrete example, this is what a logger hierarchy looks like in code.
 
 A logger inherits the properties of its ancestor logger and can define
 its own. You can think of this as similar to class inheritance in Java. 
-
-Your logging framework and SLF4J do not define logger names. The people who work
-with a logging framework and SLF4J choose logging names that are meaningful in the context
-of how they are using them.
 
 The makers of the MongoDB Java driver have defined the following logger names to
 help organize different log events in the driver. Here are the logger names
--- a/source/fundamentals/logging.txt
+++ b/source/fundamentals/logging.txt
@@ -295,7 +295,6 @@ project.
       The Log4j2 framework defines the following log levels. You can see the
       log levels listed below in order from most urgent to least urgent: 
       
-      - OFF
       - FATAL
       - ERROR
       - WARN
--- a/source/fundamentals/logging.txt
+++ b/source/fundamentals/logging.txt
@@ -58,11 +58,18 @@ can't find the ``slf4j-api`` package, the driver logs the following warning with
 
    WARNING: SLF4J not found on the classpath.  Logging is disabled for the 'org.mongodb.driver' component
 
-To set up a logger, you must add the following to your project's dependency list.
+To set up a logger, you must include the following in your project's dependency
+list.
 
 * The ``slf4j-api`` package
 * a logging framework 
 * a **binding**
+
+.. note:: 
+
+    The dependency resolution features of Gradle and Maven allow us to install
+    all three of these dependencies through a single package for some
+    logging frameworks. You will see this in the example below.
 
 A binding is a piece of code that connects the ``slf4j-api`` package with a
 logging framework. The example below shows how to bind the ``slf4j-api`` package
@@ -329,7 +336,7 @@ Logger Names
 
 Your logger uses logger names to help organize different logging events. Logger
 names are strings that form a hierarchy. A logger is an ancestor of another if
-its name followed by a "." is a prefix of another logger's name. For example,
+its name followed by a ``"."`` is a prefix of another logger's name. For example,
 ``"grandparent"`` is an ancestor of ``"grandparent.parent"`` which is an
 ancestor of ``"grandparent.parent.child"``.
 
@@ -350,7 +357,7 @@ its own. You can think of this as similar to class inheritance in Java.
 
 Your logging framework and SLF4J do not define logger names. The people who work
 with a logging framework and SLF4J choose logging names that are meaningful in the context
-of how they are using SLF4J or their logging framework.
+of how they are using them.
 
 The makers of the MongoDB Java driver have defined the following logger names to
 help organize different log events in the driver. Here are the logger names
--- a/source/fundamentals/logging.txt
+++ b/source/fundamentals/logging.txt
@@ -381,7 +381,7 @@ Example - Names
 
 This example shows how to change the log level for a specific driver logger.
 We set the root logger to "OFF" and the  ``org.mongodb.connection`` logger to
-"INFO". This will cause the logger to only output messages related to connecting
+"INFO". This will cause the application to only log messages related to connecting
 to a MongoDB instance. 
 
 Select the tab corresponding to the logging framework you are using in your
--- a/source/fundamentals/logging.txt
+++ b/source/fundamentals/logging.txt
@@ -357,19 +357,15 @@ The makers of the MongoDB Java driver have defined the following logger names to
 organize different logging events in the driver. Here are the logger names
 defined in the driver and the logging events they correspond with. 
 
-* ``org.mongodb.driver`` : root logger for the MongoDB Java driver
-* ``org.mongodb.driver.cluster`` : monitoring of your MongoDB servers
+* ``org.mongodb.driver.authenticator`` : authentication
+* ``org.mongodb.driver.client`` : events related to ``MongoClient`` instances
+* ``org.mongodb.driver.cluster`` : monitoring of MongoDB servers
 * ``org.mongodb.driver.connection`` : connections and connection pools
-* ``org.mongodb.driver.protocol`` : protocol message sent to and received from your MongoDB server
-* ``org.mongodb.driver.protocol.insert`` : insert messages and responses
-* ``org.mongodb.driver.protocol.update`` : update messages and responses
-* ``org.mongodb.driver.protocol.delete`` : delete messages and responses
-* ``org.mongodb.driver.protocol.query`` : query messages and responses
-* ``org.mongodb.driver.protocol.getmore`` : getmore messages and responses
-* ``org.mongodb.driver.protocol.killcursor`` : killcursor messages and responses
-* ``org.mongodb.driver.protocol.command`` : command messages and responses
+* ``org.mongodb.driver.connection.tls`` : TLS/SSL
+* ``org.mongodb.driver.operation`` : operations, including logging related to automatic retries
+* ``org.mongodb.driver.protocol`` : commands sent to and replies received from MongoDB servers
 * ``org.mongodb.driver.uri`` : connection string parsing
-* ``org.mongodb.driver.management`` : JMX (Java Management Extensions)'
+* ``org.mongodb.driver.management`` : JMX (Java Management Extensions)
 
 For more information on configuring your logger using logger names,
 see the `official Logback configuration guide
--- a/source/fundamentals/logging.txt
+++ b/source/fundamentals/logging.txt
@@ -191,6 +191,14 @@ tab corresponding to the logging framework you would like to use in your project
 
          12:35:00.438 [main] ERROR <my package path> - Logging an Error
 
+      .. note::
+
+         The default log level of Log4J2 is ERROR. This means that running
+         standard operations on the MongoDB Java driver will not produce output
+         from Log4J2 without configuration. To learn how to set your Log4J2
+         logger's log level, see the 
+         :ref:`example in the Configure Your Logger section of this page <configure-log-level>` .
+
       For more information on Log4j2, see the
       `Log4j2 manual <https://logging.apache.org/log4j/2.x/manual/>`__.
 
@@ -212,6 +220,8 @@ Example
 This example shows how to configure your logger's log level to INFO.
 Select the tab corresponding to the logging framework you are using in your
 project.
+
+.. _configure-log-level:
 
 .. tabs::
 
--- a/source/fundamentals/logging.txt
+++ b/source/fundamentals/logging.txt
@@ -165,14 +165,20 @@ tab corresponding to the logging framework you would like to use in your project
 Logger Names
 ------------
 
-Your logger organizes different loggers by logger name. Logger names are
-strings that form a hierarchy. A logger is an ancestor of another if its name
-followed by a "." is a prefix of another logger's name. For example, ``"grandparent"`` is
-an ancestor of ``"grandparent.parent"`` which is an ``"grandparent.parent.child"``.
+Your logger uses logger names to help organize different logging events. Logger
+names are strings that form a hierarchy. A logger is an ancestor of another if
+its name followed by a "." is a prefix of another logger's name. For example,
+``"grandparent"`` is an ancestor of ``"grandparent.parent"`` which is an
+ancestor of ``"grandparent.parent.child"``.
 
 For a concrete example, this is what a logger hierarchy looks like in code.
 
 .. code-block:: java
+
+   import org.slf4j.Logger;
+   import org.slf4j.LoggerFactory;
+
+   ...
 
    Logger logger_parent = LoggerFactory.getLogger("parent");
    Logger logger_child = LoggerFactory.getLogger("parent.child");
@@ -180,9 +186,9 @@ For a concrete example, this is what a logger hierarchy looks like in code.
 A logger inherits the properties of its ancestor logger and can define
 its own. You can think of this as similar to class inheritance in Java. 
 
-Your logging framework does not define logger names. The people who work
-with a logging framework choose logging names that are meaningful in the context
-of how they are using the logging framework.
+Your logging framework and SLF4J do not define logger names. The people who work
+with a logging framework and SLF4J choose logging names that are meaningful in the context
+of how they are using SLF4J or their logging framework.
 
 The makers of the MongoDB Java driver have defined the following logger names to
 help organize different log events in the driver. Here are the logger names
--- a/source/fundamentals/logging.txt
+++ b/source/fundamentals/logging.txt
@@ -248,7 +248,7 @@ project.
 
          <configuration>
          <appender name="CONSOLE"
-           class="ch.qos.logback.core.ConsoleAppender">
+                   class="ch.qos.logback.core.ConsoleAppender">
            <!-- deny all events with a level below INFO, that is TRACE and DEBUG -->
            <filter class="ch.qos.logback.classic.filter.ThresholdFilter">
              <level>INFO</level>
@@ -397,8 +397,7 @@ messages related to connecting to your MongoDB instance.
 
          <configuration>
             <appender name="CONSOLE"
-                     class="ch.qos.logback.core.ConsoleAppender">
-               <!-- deny all events with a level below INFO, that is TRACE and DEBUG -->
+                      class="ch.qos.logback.core.ConsoleAppender">
                <filter class="ch.qos.logback.classic.filter.ThresholdFilter">
                   <level>INFO</level>
                </filter>
--- a/source/fundamentals/logging.txt
+++ b/source/fundamentals/logging.txt
@@ -106,12 +106,6 @@ tab corresponding to the logging framework you would like to use in your project
            <artifactId>logback-classic</artifactId>
            <version>1.2.3</version>
          </dependency>
-         <!-- Logback core functionality -->
-         <dependency>
-           <groupId>ch.qos.logback</groupId>
-           <artifactId>logback-core</artifactId>
-           <version>1.2.3</version>
-         </dependency>
 
       Once you have installed the above dependencies, connect to your
       MongoDB instance and retrieve a document with the following code.
--- a/source/fundamentals/logging.txt
+++ b/source/fundamentals/logging.txt
@@ -149,12 +149,6 @@ tab corresponding to the logging framework you would like to use in your project
    
             .. code-block:: xml
    
-               <!-- SLF4J API -->
-               <dependency>
-                 <groupId>org.slf4j</groupId>
-                 <artifactId>slf4j-api</artifactId>
-                 <version>1.7.30</version>
-               </dependency>
                <!-- SLF4J-Log4j2 binding -->
                <dependency>
                  <groupId>org.apache.logging.log4j</groupId>
@@ -170,7 +164,6 @@ tab corresponding to the logging framework you would like to use in your project
             .. code-block:: none
    
                dependencies {
-                  implementation 'org.slf4j:slf4j-api:1.7.30'
                   implementation 'org.apache.logging.log4j:log4j-slf4j-impl:2.14.1'
                }
     
--- a/source/fundamentals/logging.txt
+++ b/source/fundamentals/logging.txt
@@ -52,7 +52,7 @@ For more information on SLF4J,
 To set up a logger you must add the ``slf4j-api`` package, a logging framework, and a
 **binding** to your project's dependency list. A binding is a piece of code that
 connects the ``slf4j-api`` package with a logging framework. The example below shows
-how to bind the ``slfj-api`` package to the two most popular logging frameworks, Log4j2 and
+how to bind the ``slf4j-api`` package to the two most popular logging frameworks, Log4j2 and
 Logback.
 
 Setting up a logger is optional. When you start your application the MongoDB
@@ -204,7 +204,7 @@ project.
       - DEBUG
       - TRACE 
 
-      To configure your Logback logger to log all messages of level INFO or
+      To configure your logger to log all messages of level INFO or
       higher, set your ``logback.xml`` file to the following.
 
       .. code-block:: xml
@@ -254,7 +254,7 @@ project.
       - TRACE
       - ALL 
 
-      To configure your Log4j2 logger to log all messages of level INFO or
+      To configure your logger to log all messages of level INFO or
       higher, set your ``log4j2.xml`` file to the following.
 
       .. code-block:: xml
--- a/source/fundamentals/logging.txt
+++ b/source/fundamentals/logging.txt
@@ -255,7 +255,7 @@ project.
 
          <configuration>
             <appender name="CONSOLE"
-                     class="ch.qos.logback.core.ConsoleAppender">
+                      class="ch.qos.logback.core.ConsoleAppender">
                <encoder>
                   <pattern>
                         %-4relative [%thread] %-5level %logger{30} - %msg%n
--- a/source/fundamentals/logging.txt
+++ b/source/fundamentals/logging.txt
@@ -333,8 +333,8 @@ Logger Names
 ------------
 
 Your logger uses logger names to help organize different logging events. Logger
-names are strings that form a hierarchy. A logger is an ancestor of another if
-its name followed by a ``"."`` is a prefix of another logger's name. For example,
+names are strings that form a hierarchy. A logger is an ancestor of another logger if
+its name followed by a ``"."`` is a prefix of the other logger's name. For example,
 ``"grandparent"`` is an ancestor of ``"grandparent.parent"`` which is an
 ancestor of ``"grandparent.parent.child"``.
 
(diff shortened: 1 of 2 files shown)


source/fundamentals/logging.txt
```diff
@@ -0,0 +1,273 @@
...
+
+To set up a logger you must add the ``slf4j-api`` package, a logging framework, and a
+**binding** to your project's dependency list. A binding is a piece of code that
+connects the ``slf4j-api`` package with a logging framework. The two most
+popular logging frameworks used with ``slf4j-api`` are Log4j2 and Logback.
+
+Setting up a logger is optional. When you start your application the MongoDB
+Java driver looks for the ``slf4j-api`` package in your classpath. If the driver
+can't find ``slf4j-api``, the driver logs a single warning with
+``java.util.logging`` and disables all further logging.
```
terakilobyte, line 56: A few items here.

1. It is stated that a particular dependency must be added, and that it creates a binding. Then the term binding is defined.

* Is including the term binding worthwhile? It doesn't appear to be required for the task of setting up a logger?

2. It is stated that the user MUST add the `slf4j-api` package. 
 
* Is that correct? [This post](https://stackoverflow.com/a/59374626/4279756) suggests an alternative based on desired framework.

3. The two most popular logging frameworks... feels incomplete.

* Perhaps rephrasing this to "This guide uses the two most popular logging framework..."


Lastly, this all feels very informational and not directly related to the task of "Set(ting) Up a Logger".  I wonder if information here could be extracted into an overview section?

source/fundamentals/logging.txt
```diff
@@ -0,0 +1,273 @@
...
+
+Here is the error the driver outputs if it cannot find the ``slf4j-api`` package:
+
+.. code-block:: none
+
+   WARNING: SLF4J not found on the classpath.  Logging is disabled for the 'org.mongodb.driver' component
+
```
terakilobyte, line 63: nitpick: Convert to an active action. Also, is it an error or warning?

> The driver will output a single warning when no logging framework is configured:

source/fundamentals/logging.txt
```diff
@@ -0,0 +1,273 @@
...
+
+Below is an example showing how to add SLF4J dependencies to your
+project and set up your logger using either Logback or Log4j2. The examples uses Maven.
```
terakilobyte, line 66: This feels like the actual start of this particular task. Can this be switched to imperative phrasing?

> Add SLF4J dependencies to your project and use your desired framework. The following examples use Maven.

source/fundamentals/logging.txt
```diff
@@ -0,0 +1,273 @@
...
+
+      .. code-block:: xml
+
+         <!-- SLF4J API -->
+         <dependency>
+           <groupId>org.slf4j</groupId>
+           <artifactId>slf4j-api</artifactId>
+           <version>1.7.30</version>
+         </dependency>
+         <!-- This package contains an SLF4J-Logback binding -->
```
terakilobyte, line 84: The comment here doesn't match comment in the other example. Is there a reason for this?

source/fundamentals/logging.txt
```diff
@@ -0,0 +1,273 @@
...
+           <version>1.2.3</version>
+         </dependency>
+         <!-- Logback Core Functionality -->
+         <dependency>
+           <groupId>ch.qos.logback</groupId>
+           <artifactId>logback-core</artifactId>
+           <version>1.2.3</version>
+         </dependency>
+
+      Once you have installed the above dependencies, try connecting to a
```
terakilobyte, line 97: nitpick: there is no try.

In all seriousness, be prescriptive. And this seems like another step. If so, use the proper mood.

> Connect to a MongoDB instance and retrieve...

source/fundamentals/logging.txt
```diff
@@ -0,0 +1,273 @@
...
+           <version>2.14.1</version>
+         </dependency>
+         <!-- Log4j2 core functionality -->
+         <dependency>
+           <groupId>org.apache.logging.log4j</groupId>
+           <artifactId>log4j-core</artifactId>
+           <version>2.14.1</version>
+         </dependency>
+
+      Once you have installed the above dependencies, try logging an error.
```
terakilobyte, line 145: Why is this step different with this framework?

source/fundamentals/logging.txt
```diff
@@ -0,0 +1,273 @@
...
+      For more information on Log4j2, see the
+      `Log4j2 manual <https://logging.apache.org/log4j/2.x/manual/>`__.
+
+Configure Your Logger
+---------------------
+
+To configure your logger, you must configure the logging framework bound to
+SLF4J.
+
+The example below shows how to configure your logger's log level.
```
terakilobyte, line 172: What's the difference between `logger` and `logging framework` here? Do they mean different things, or could they be seen as the same thing by the user?

Also, the example shows how to configure the log level. If that's the task, should it be the heading for this section?

source/fundamentals/logging.txt
```diff
@@ -0,0 +1,273 @@
...
+
+Log levels are the hierarchy of messages output by a logger based on urgency.
+Messages related to errors are the most urgent, and messages related to
+stack traces and debugging are the least urgent. A logger's log level
+specifies a lower bound for how urgent a message must be for the logger to
+output it.
```
terakilobyte, line 178: I'm not sure "urgent" is correct? My thinking is that urgency is assigned to labels by the developer, "error" and "warning" are just labels unless they are being semantically parsed.

I'm unable to fully determine the subject of this paragraph.  It also feels like fundamental information that should be presented at the beginning of the section.

source/fundamentals/logging.txt
```diff
@@ -0,0 +1,273 @@
...
+
+Here is an example that shows how to configure your logger's log level to
+``"INFO"`` for both the Logback and Log4j2 logging frameworks.
+
+.. tabs::
+
+   .. tab:: Logback
+      :tabid: logback-binding
+
+      You specify Logback configurations in a file named ``logback.xml``. Your
```
terakilobyte, line 188: You is implied here

terakilobyte (changes requested):
some nits and suggestions

source/fundamentals/logging.txt
biniona-mongodb: Thank you for the notes. I will respond to your feedback in order.

1.  I believe including the term binding is worthwhile.  The ``slf4j-api`` requires a separate framework specific binding package that binds it to a logging framework. This means that users must install three packages to use Logback or Log4j2. 
2. I believe it is correct that the user must add the ``slf4j-api`` in order to enable logging. The [current driver documentation](https://mongodb.github.io/mongo-java-driver/4.2/driver/reference/logging/) of logging states that you must install the ``sl4j-api`` in order to enable logging. Here is [the section of the Driver Source Code](https://github.com/mongodb/mongo-java-driver/blob/20c160a8906e6d65dc89e7f55b5d85f79f79e114/bson/src/main/org/bson/diagnostics/Loggers.java#L49) that will output a no-op logger is `slf4j-api` is not installed. SLF4J allows you to plugin your desired framework (it is a facade), so it is true that the user can use their desired framework.
3. I agree your proposed phrasing improves the sentence. I didn't realize how arbitrary and incomplete my wording sounded. Going forward, I will try to identify these incomplete phrases. Fix:
4. I agree that an overview/information section could be beneficial. I saw this suggestion in some of your other comments as well. I will experiment with this and try to group foundational knowledge into its own section.

source/fundamentals/logging.txt
biniona-mongodb: Thank you for the note. I agree that phrasing this in terms of the action the driver is performing improves the sentence. Also it is a warning, good catch. Apologies for mixing up my terms. Fix:

Going forward, I will try and phrase things in terms of actions being performed rather than things that exist.

source/fundamentals/logging.txt
biniona-mongodb: Log4J2 has a default logger level of "[ERROR](https://logging.apache.org/log4j/2.x/manual/configuration.html)", where as Logback has a default logger level of "[DEBUG](https://github.com/qos-ch/logback/blob/master/logback-examples/src/main/resources/chapters/configuration/sample0.xml)" . This means that performing basic operations with the driver will give output for Logback but not Log4j2. There is a way to change the root logger level of Log4J2 [programmatically](https://logging.apache.org/log4j/log4j-2.4/faq.html#reconfig_level_from_code), but it is not supported by the public api. Configuring the log framework to change the log level through a config file is not easy, and is the subject of the next section, so I thought it made sense to use different examples (you can only use 1 log framework at a time as well). 

I tried finding a way to make the driver output an "ERROR" log without using esoteric code, but had trouble doing it (Maybe Jeff Yemin would know a way?).

source/fundamentals/logging.txt
biniona-mongodb: Good question. I tried to use "logger" to refer to the series of interconnected packages that allow the MongoDB Java Driver to output logs (`slf4j-api` + binding + logging framework), where as "logging framework" only refers to the package bound to `slf4j-api` that implements logging functionality. I agree that "logger" and "logging framework" are hard to distinguish between, but I am struggling to come up with better terms. I will continue to try and think of a clearer way to phrase this, and any suggestions are appreciated. 

I tried to use "configuring log level" as an example for showing how to configure your logger (use your logging framework's native configuration system). The concrete example tries to show how you can link logging framework documentation to your MongoDB Java Project. I can see how this is not clear as the page is written.

source/fundamentals/logging.txt
biniona-mongodb: Good question. The reason the comments are different is that the `logback-classic` package includes more than just an SLF4J-Logback binding, where as the `log4j-slf4j-impl` package only contains a logback binding. It felt inaccurate for me to write the comment "SLF4J-Logback Binding" for logback and it felt excessive to write "This package contains an SLF4J-Log4j2 binding" for Log4j2. The two frameworks are different projects, so I think it is okay for the xml comments to be different. Let me know if you disagree and you would like me to make both comments "SLF4J-<framework name>".

source/fundamentals/logging.txt
```diff
@@ -0,0 +1,273 @@
...
+In this guide, we show you how to set up and configure a logger in the
+MongoDB Java driver.
+
+.. What do any new terms mean?
+
+.. Not applicable. We are treating definition of logger as understood.
+
+.. What can you expect to see on this page?
+
+In this guide you can learn how to do the following:
```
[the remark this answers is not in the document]
biniona-mongodb: Thanks for the note. I agree this phrasing improves the sentence. I think I am afraid of very short sentences, but I know I should not be. Going forward, I will try to not add extra words to my sentences. I will try and improve my eye/ear for when words are not doing useful work.  Committing suggestion.

source/fundamentals/logging.txt
```diff
@@ -0,0 +1,273 @@
...
+- Configure the log level of your logger
+
+.. Who should read this? Where should I go if this isn't the page I was looking for?
+
+.. Not Applicable. No natural redirect location.
+
+Set Up a Logger
+---------------
+
+To set up a logger for your project, you need to add some dependencies.
```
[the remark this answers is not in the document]
biniona-mongodb: Thank you for the note. This sentence was added in our team review session to address a potentially harsh transition. I agree that it is not adding much value to users (outside of potentially helping ease into set up a logger section). I have deleted it, and am trying to use section headings to better transition into content rather than use a transition sentence.

source/fundamentals/logging.txt
```diff
@@ -0,0 +1,273 @@
...
+
+The MongoDB Java driver uses the Simple Logging Facade For Java (SLF4J).
+SLF4J allows you to specify your logging framework of choice at deployment time.
+For more information on SLF4J,
+`see the SLF4J documentation <http://www.slf4j.org/>`__.
```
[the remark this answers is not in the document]
biniona-mongodb: Thanks for the note. I have moved made "background" and "example" subsections in the "set up a logger" heading. I have moved this background content into the background subsection. https://github.com/mongodb/docs-java/pull/76/commits/8835d4ee24868817786c38926f5cd52912437b16

source/fundamentals/logging.txt
biniona-mongodb: Thanks for the note. Going forward, I will try (there is no try, but it is unlikely I will perfectly implement this feedback immediately haha) to be prescriptive in my writing. I will avoid the word "try" is step by step guides.

Fix: https://github.com/mongodb/docs-java/pull/76/commits/a9644a761d0a1f71ec9572015087f6246b69a44e

source/fundamentals/logging.txt
biniona-mongodb: Thanks for the note. I agree "you" is implied here, and having it said explicitly is not helpful for readers.

Fix: https://github.com/mongodb/docs-java/pull/76/commits/d4a6e8d0f7bd656582c3c4de4845f71867d8e4ac

source/fundamentals/logging.txt
biniona-mongodb: Thanks for the note. I agree that my phrasing is not imperative enough. I have moved this into an "example" subsection and rewritten this as:

> This example shows how to set up your logger. This example uses Maven. Click the
tab corresponding to the logging framework you would like to use in your project.

After breaking this off into its own section, the imperative mood used in your suggested change seemed to be out of place. I felt an introductory sentence clearly stating what the example did, what it uses, and what you should do was more fitting as a section header. Please let me know if you disagree with this assessment.

source/fundamentals/logging.txt
biniona-mongodb: Thank you for the note. I disagree that log level labels do not have meaning. The `Log4J2` default log levels are [placed in a hierarchy by corresponding integer values ](https://logging.apache.org/log4j/2.x/manual/customloglevels.html) giving them a very real sort order. The log4j2 source code [documents standard log levels by severity ](https://github.com/apache/logging-log4j2/blob/79d668afff389b4d9b8c1bd9af26269e28667597/log4j-1.2-api/src/main/java/org/apache/log4j/Level.java#L54). I think severity of message corresponds to urgency of responding to log. Based on this, I believe urgent is the correct term.

If you strongly disagree with this I can remove it, but I believe that log levels have semantic meaning in code.

source/fundamentals/logging.txt
```diff
@@ -0,0 +1,285 @@
...
+
+.. Who should read this? Where should I go if this isn't the page I was looking for?
+
+.. Not Applicable. No natural redirect location.
+
+Set Up a Logger
+---------------
+
+In this section, we provide background on the dependencies necessary to set up a
+logger and walk through an example logger setup.
```
terakilobyte, line 42: Is this sentence needed here or are the headings self-explanatory?

source/fundamentals/logging.txt
```diff
@@ -0,0 +1,285 @@
...
+The MongoDB Java driver uses the Simple Logging Facade For Java (SLF4J).
+SLF4J allows you to specify your logging framework of choice at deployment time.
+For more information on SLF4J,
+`see the SLF4J documentation <http://www.slf4j.org/>`__.
+
+To set up a logger you must add the ``slf4j-api`` package, a logging framework, and a
+**binding** to your project's dependency list. A binding is a piece of code that
+connects the ``slf4j-api`` package with a logging framework. The example below shows
+how to bind the ``slfj-api`` package to the two most popular logging frameworks, Log4j2 and
+Logback.
```
terakilobyte, line 56: Could this be moved closer to the example so the user doesn't have to keep this in mind while reading the next paragraph?

source/fundamentals/logging.txt
```diff
@@ -0,0 +1,285 @@
...
+
+The MongoDB Java driver uses the Simple Logging Facade For Java (SLF4J).
+SLF4J allows you to specify your logging framework of choice at deployment time.
+For more information on SLF4J,
+`see the SLF4J documentation <http://www.slf4j.org/>`__.
+
+To set up a logger you must add the ``slf4j-api`` package, a logging framework, and a
+**binding** to your project's dependency list. A binding is a piece of code that
+connects the ``slf4j-api`` package with a logging framework. The example below shows
+how to bind the ``slfj-api`` package to the two most popular logging frameworks, Log4j2 and
```
terakilobyte, line 55: typo?
```suggestion
how to bind the ``slf4j-api`` package to the two most popular logging frameworks, Log4j2 and
```

source/fundamentals/logging.txt
```diff
@@ -0,0 +1,285 @@
...
+Logback.
+
+Setting up a logger is optional. When you start your application the MongoDB
+Java driver looks for the ``slf4j-api`` package in your classpath. If the driver
+can't find ``slf4j-api``, the driver logs a single warning with
+``java.util.logging`` and disables all further logging.
+
+The driver outputs this warning if it cannot find the ``slf4j-api`` package:
```
terakilobyte, line 63: This sentence repeats what it said in the previous paragraph, and could probably be removed.

source/fundamentals/logging.txt
```diff
@@ -0,0 +1,285 @@
...
+
+.. code-block:: none
+
+   WARNING: SLF4J not found on the classpath.  Logging is disabled for the 'org.mongodb.driver' component
+
+Example
+~~~~~~~
+
+This example shows how to set up your logger. This example uses Maven. Click the
```
terakilobyte, line 72: Two short sentences beginning with "This example" feels strange to me. Can they be combined? Is stating the example uses Maven required? Is there an alternative? Do all the examples use Maven?

source/fundamentals/logging.txt
terakilobyte: is there any meaningful difference to the user though? If not, perhaps the importance you are placing on it is for you alone?

source/fundamentals/logging.txt
```diff
@@ -0,0 +1,285 @@
...
+tab corresponding to the logging framework you would like to use in your project.
+
+.. tabs::
+
+   .. tab:: Logback
+      :tabid: logback-binding
+
+      If you add these dependencies to your ``pom.xml`` file, the MongoDB Java
+      driver uses the Logback framework to output your logs.
```
terakilobyte, line 81: Isn't this implied by clicking on the tab titled `Logback`?. What value is this providing?

source/fundamentals/logging.txt
```diff
@@ -0,0 +1,285 @@
...
+      `Log4j2 manual <https://logging.apache.org/log4j/2.x/manual/>`__.
+
+Configure Your Logger
+---------------------
+
+To configure your logger, you must configure the logging framework bound to
+SLF4J.
+
+The following subsection provides a concrete example showing how to configure
+your logger.
```
terakilobyte, line 181: Do these provide any meaning or instruction more than the headings already do?

source/fundamentals/logging.txt
```diff
@@ -0,0 +1,285 @@
...
+
+Configure Your Log Level
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+In this subsection we show how to configure your logger's **log level**.
+
+A logger's log level specifies a lower bound for how urgent a message must be
+for the logger to output it.
```
terakilobyte, line 189: nitpick: this reads a bit strangely to me, maybe because of `it`.  Perhaps something like
> A logger's log level specified a lower bound for how urgent a message must be to be included in the logger's output.

source/fundamentals/logging.txt
```diff
@@ -0,0 +1,285 @@
...
+      urgent: 
+      
+      - ERROR
+      - WARN
+      - INFO
+      - DEBUG
+      - TRACE 
+
+      To configure your Logback logger to log all messages of level INFO or
+      higher, set your ``logback.xml`` file to the following.
```
terakilobyte, line 215: This is repeating the same information "This example shows..." from line 191

source/fundamentals/logging.txt
```diff
@@ -0,0 +1,285 @@
...
+      - FATAL
+      - ERROR
+      - WARN
+      - INFO
+      - DEBUG
+      - TRACE
+      - ALL 
+
+      To configure your Log4j2 logger to log all messages of level INFO or
+      higher, set your ``log4j2.xml`` file to the following.
```
terakilobyte, line 265: Again, repetitive.

terakilobyte (changes requested):
A few more things I noticed. It's looking good.

source/fundamentals/logging.txt
biniona-mongodb: Thanks for the note. I agree, it can be removed. Fix:

Going forward, I will work on creating linear, non-repeating sentences. A theme in this review has been issues around my repeating content. I will work to correct this habit.

biniona-mongodb: https://github.com/mongodb/docs-java/pull/76/commits/286ea92be8c2beeb5414ce818f315c94b86b1dfa

source/fundamentals/logging.txt
biniona-mongodb: Thanks for the note. I believe the first sentence does (it gives explicit guidance for going about configuring your logger that I do not think fits naturally into the example), but the second sentence does not. I have removed the second sentence and changed the heading of the subsection to "Example". Fix:

source/fundamentals/logging.txt
biniona-mongodb: Thanks for the note. My gut is to be explicit about what each line of code does. I have read documentation that presented steps on how to do something, but did not clearly state what each step did (assuming readers would infer what was going on). Often I made a false assumption and got off course. I am not sure whether or not that applies here, but that is my background for including this preface to the dependencies list.

I have changed the above line to "Add the following dependencies to you "pom.xml" file." I think readers will appreciate the brevity, but I fear some readers may lose some clarity. Would appreciate discussion around this later on. Fix:

source/fundamentals/logging.txt
biniona-mongodb: Thanks for the note. The Quickstart uses both Maven and Gradle. It felt excessive to make a Maven/Gradle tab for both logging frameworks (2. layers of tabs) so I decided to only make examples for Maven. I chose Maven because it was the build tool used in our Java Driver Course.

source/fundamentals/logging.txt
biniona-mongodb: Thanks for the note. Good catch, apologies for letting that slip through. Fix: https://github.com/mongodb/docs-java/pull/76/commits/b05d369d84c141f18d77e64e8c5c905239a02c03

source/fundamentals/logging.txt
biniona-mongodb: Thanks for the note. I thought this was helpful, but after starting at this for a while I can see how it is not adding significant value for readers. I changed this from stating what this section of code did, so providing a direction "set you file to this..." and then providing steps such that readers can produce a state to test their configuration. Fix:

Thanks for pointing this out. I am realizing how off base my understanding of repetition in writing is. Going forward, I will try and continue to develop my ear for when repetition is a good tool or whether something else would be better for learners.

source/fundamentals/logging.txt
biniona-mongodb: Thanks for the note. Response to equivalent comment earlier: https://github.com/mongodb/docs-java/pull/76#discussion_r621257383

source/fundamentals/logging.txt
biniona-mongodb: Thanks for the note. My ear does not pick up anything wrong with this sentence, and [the style guide approves](https://docs.mongodb.com/meta/style-guide/style/prepositions/) ending sentences in prepositions, so I am not sure how best to address this feedback. 

My writegood plugin is saying that the suggested change contains passive voice with "to be included" (who is including?). Would changing "it" to "that message" fix the problem you are hearing with the sentence?

Fix:

source/fundamentals/logging.txt
biniona-mongodb: Thanks for the note. I could see how the vertical scrolling could be annoying for readers. I have swapped the "optional" paragraph with the "set up dependencies" paragraph such that the example comes immediately after the "set up dependencies" paragraph. Fix:

source/fundamentals/logging.txt
biniona-mongodb: Thanks for the note. Assuming a user copies and pastes these dependencies into their code, I think having the comment accurately reflect the contents of the package will meaningfully benefit users. The logback classic has more than just an `SLF4J-Logback` binding, so if a user ever wants to clean up their `pom.xml` file and finds these dependencies there without context, I think "contains an `SLF4J-Logback binding` more accurately expresses that "this was added here because of `SLF4J`, but this contains things that may not be expressly related to SLF4J, becareful about removing it".

biniona-mongodb: Let me know if you disagree, it you want me to just make the change I can. I do actually feel the different comment is beneficial, but my instincts could just be incorrect, in which case I am totally open to making comments identical.

source/fundamentals/logging.txt
biniona-mongodb: I rewrote this section, so response is relative to rewrite: https://github.com/mongodb/docs-java/pull/76/commits/d7007eaa9fe51be1362c4c207773a6d823be4f27

Thanks for the note. After rereading this a few times, I think the sentence serves as a useful transition, but my instincts could be incorrect. I think "background on the dependencies" is meaningfully different than "background", and "example of a logger setup" is different than "example". That being said, if you disagree, I would appreciate a discussion of what you feel would improve the section as I am having trouble seeing the issue. 

I am trying to balance the following things: 

1. Smooth Transition into Section
2. Heavy use of heading structure
3. Avoid [stacking headings](https://docs.mongodb.com/meta/style-guide/style/titles-and-headings/)

source/fundamentals/logging.txt
```diff
@@ -0,0 +1,297 @@
...
+~~~~~~~
+
+In this subsection we show how to configure your logger's **log level**.
+
+A logger's log level specifies a lower bound for how urgent a message must be
+for the logger to output that message.
+
+This example shows how to configure your logger's log level to INFO.
+Select the tab corresponding to the logging framework you are using in your
+project.
```
terakilobyte, line 181: We might be able to improve the flow here. 

We state they have to configure the logging framework bound to sfl4j, then we jump into the log level, what it means, and then setting it.

Perhaps in the paragraph under the main heading, we could state the same information and include the info about the log level and tie them together. Then under the example heading, we could state what the example is going to do.

Lastly, is there a cleaner way than "To configure..., you must configure.."? Here's a rough jam of my idea:

```
To configure your logger, set up the logging framework bound to SFL4J by specifying the **log level** and <any other stuff they need to do>.

A logger's log level specifies the lower bound...

Example
~~~~~~~
...
```

Thoughts?

terakilobyte (changes requested):
Overall, really good. I think we can clean the last section up just a bit. Thoughts?

source/fundamentals/logging.txt
biniona-mongodb: Thanks for the note. I agree that "to configure... you must configure..." is not helpful for learners. I have changed it to "to configure.. you must use the configuration system of the logging framework...". This felt less circular. Let me know if this is not satisfactory.

I have move the log level definition into the main heading, and added this transition sentence "In the following example we show how you can use your logging framework's
configuration system to set your logger's **log level**.".

Fix:

terakilobyte (approved):
lgtm

biniona-mongodb:
Hi @rozza and @jyemin, may I please have a technical review for this PR?

source/fundamentals/logging.txt
```diff
@@ -0,0 +1,298 @@
...
+Java driver looks for the ``slf4j-api`` package in your classpath. If the driver
+can't find the ``slf4j-api`` package, the driver logs the following warning with
+``java.util.logging`` and disables all further logging:
+
+.. code-block:: none
+
+   WARNING: SLF4J not found on the classpath.  Logging is disabled for the 'org.mongodb.driver' component
+
+To set up a logger you must add the ``slf4j-api`` package, a logging framework, and a
+**binding** to your project's dependency list. A binding is a piece of code that
```
jyemin, line 62: Nit: I think the grammar of this sentence is correct, but when I first read it, I interpreted "a binding to your project's dependency list" as a single phrase, rather than "your project's dependency list" as something that all three items had to be added to.  Consider re-wording to something like:

> To set up a logger you muse add to your project's dependency list the ...

source/fundamentals/logging.txt
```diff
@@ -0,0 +1,298 @@
...
+         <!-- SLF4J API -->
+         <dependency>
+           <groupId>org.slf4j</groupId>
+           <artifactId>slf4j-api</artifactId>
+           <version>1.7.30</version>
+         </dependency>
+         <!-- This package contains an SLF4J-Logback binding -->
+         <dependency>
+           <groupId>ch.qos.logback</groupId>
+           <artifactId>logback-classic</artifactId>
```
jyemin, line 91: On http://logback.qos.ch/setup.html, it says:

> TRANSITIVITY:  Note that in addition to logback-classic.jar, the above declaration will automatically pull-in slf4j-api.jar and logback-core.jar into your project by virtue of Maven's transitivity rules.

So I think only this one dependency is required.

source/fundamentals/logging.txt
```diff
@@ -0,0 +1,298 @@
...
+Setting up a logger is optional. When you start your application the MongoDB
+Java driver looks for the ``slf4j-api`` package in your classpath. If the driver
+can't find the ``slf4j-api`` package, the driver logs the following warning with
+``java.util.logging`` and disables all further logging:
+
+.. code-block:: none
+
+   WARNING: SLF4J not found on the classpath.  Logging is disabled for the 'org.mongodb.driver' component
+
+To set up a logger you must add the ``slf4j-api`` package, a logging framework, and a
```
jyemin, line 61: See my note below.  Because logback was developed alongside SLF4J, it actually only requires a single dependency.

source/fundamentals/logging.txt
```diff
@@ -0,0 +1,298 @@
...
+**binding** to your project's dependency list. A binding is a piece of code that
+connects the ``slf4j-api`` package with a logging framework. The example below shows
+how to bind the ``slf4j-api`` package to the two most popular logging frameworks, Log4j2 and
+Logback.
+
+Example
```
jyemin, line 67: I think you should add a note saying that the versions listed below are illustrative, and users should check documentation for the current releases of these logging frameworks.  That might save us from requests to keep updating the version numbers here.

source/fundamentals/logging.txt
```diff
@@ -0,0 +1,298 @@
...
+           <artifactId>slf4j-api</artifactId>
+           <version>1.7.30</version>
+         </dependency>
+         <!-- This package contains an SLF4J-Logback binding -->
+         <dependency>
+           <groupId>ch.qos.logback</groupId>
+           <artifactId>logback-classic</artifactId>
+           <version>1.2.3</version>
+         </dependency>
+         <!-- Logback core functionality -->
```
rozza, line 94: Nit: As this is already a dependency of `logback-classic` its probably not required to be explicitly added.

source/fundamentals/logging.txt
```diff
@@ -0,0 +1,298 @@
...
+           <artifactId>slf4j-api</artifactId>
+           <version>1.7.30</version>
+         </dependency>
+         <!-- SLF4J-Log4j2 binding -->
+         <dependency>
+           <groupId>org.apache.logging.log4j</groupId>
+           <artifactId>log4j-slf4j-impl</artifactId>
+           <version>2.14.1</version>
+         </dependency>
+         <!-- Log4j2 core functionality -->
```
rozza, line 136: Nit: As this is already a dependency of `log4j-slf4j-impl` its probably not required to be explicitly added.

source/fundamentals/logging.txt
```diff
@@ -0,0 +1,298 @@
...
+      You should see output that resembles the following:
+
+      .. code-block:: none
+
+         12:35:00.438 [main] ERROR <my package path> - Logging an Error
+
+      For more information on Log4j2, see the
+      `Log4j2 manual <https://logging.apache.org/log4j/2.x/manual/>`__.
+
+Configure Your Logger
```
jyemin, line 165: I really like that we have configuration examples here, but I think we also need to document the logger names that are actually used by the driver, as in the current reference docs (http://mongodb.github.io/mongo-java-driver/4.2/driver/reference/logging/).  Perhaps we can include that information somewhere above, and then use one of the logger names the driver uses in the examples below?

For example, you could configure the "org.mongodb.driver.connection" logger to DEBUG and show the increased logging messages for connection usage.

source/fundamentals/logging.txt
```diff
@@ -0,0 +1,298 @@
...
+      following code.
+
+      .. code-block:: java
+
+         import org.slf4j.Logger;
+         import org.slf4j.LoggerFactory;
+
+         ...
+
+         Logger logger = LoggerFactory.getLogger(App.class);
```
jyemin, line 153: Replace with `getLogger("MyApp")` so the example will work regardless of the class name that they pasted this code into.

source/fundamentals/logging.txt
```diff
@@ -0,0 +1,298 @@
...
+tab corresponding to the logging framework you would like to use in your project.
+
+.. tabs::
+
+   .. tab:: Logback
+      :tabid: Logback-binding
+
+      Add the following dependencies to your ``pom.xml`` file.
+
+      .. code-block:: xml
```
rozza, line 80: Are there plans to add gradle style examples  - as per the quickstart?

eg:

```
dependencies {
   implementation 'org.slf4j:slf4j-api:1.7.30'
   implementation 'ch.qos.logback:logback-classic:1.2.3'
}
```

rozza (changes requested):
This looks great. A couple of comments and one question.

jyemin (changes requested):
Sorry for taking so long to get to this one.   I have a bunch of change requests but overall I really like the structure and details.

The main substantive comment is that the page needs to include info about all the loggers actually used by the driver.

source/fundamentals/logging.txt
biniona-mongodb: Thank you for the note. Good catch, I apologies for not documenting the logger names. I agree that it is important that the logger names be included in the docs, as there is not a way for learners to easily get that information outside the documentation. I have added a "Logger Names" section where I give some background on logger names and how they are used in the driver.

Logger Section:

source/fundamentals/logging.txt
biniona-mongodb: Thank you for the note. Great idea! I think it makes sense to add a note letting readers know that we should not be considered a source of truth in regards to version numbers on these external projects.

Fix:

Going forward, I will try and help readers understand our relationship to external projects and how they should navigate that relationship (in this case consult other projects docs) in my writing.

source/fundamentals/logging.txt
biniona-mongodb: Thanks for the note. On rereading this, I can see how it is not as clear as it could be. I took your rewording suggestion and then broke the dependencies our into an unordered list.

Fix:

source/fundamentals/logging.txt
biniona-mongodb: Thanks for the note. Fix: https://github.com/mongodb/docs-java/pull/76/commits/d46bdd65acd99fbd8b6446f51e3172c3a06fe2c9

For my own edification (and to help make sure I don't make the same mistake twice), is it correct that the reason for this is that the logger name "MyApp" will definitely not error out (as it just a string literal), where as `App.class` may potentially error out if there is no `App` class accessible from where a reader pastes this code snippet?

source/fundamentals/logging.txt
biniona-mongodb: Thank you for the note. Good catch. I will remove the `logback-classic` dependency. 

It seems like [`logback-classic`](https://mvnrepository.com/artifact/ch.qos.logback/logback-classic/1.2.3) also has `slf4j-api` as a dependency. Would you like the `slf4j-api` package removed from the dependency list? I just tested this in my local app and it seems to be working well.

source/fundamentals/logging.txt
biniona-mongodb: I did not plan to add them, but I can add them now. I was hesitant to add them as I didn't want two layers of tabs, but gradle users will probably appreciate the information. Addition:

source/fundamentals/logging.txt
biniona-mongodb: Thanks for the note, I will remove the `slfj-api` dependency. Fix:

Going forward, I will consult the Maven site to prune my dependency list before requesting an external review.

source/fundamentals/logging.txt
biniona-mongodb: Thanks for the note. I added a note about only needing to install a single package due to Maven/Gradle's dependency resolution features.

biniona-mongodb:
Hi @jyemin and @rozza, thank you for the feedback, it was very helpful. When you have time, could I please have a second round review?

source/fundamentals/logging.txt
jyemin: Yes, that's right.  It's common practice in business apps to just have a logger per class, and use the full class name as the  name of the logger by passing `<this class>.class`. But since you're not showing a full class here, safer to use a string as the logger name so that someone could copy and paste this code anywhere.

source/fundamentals/logging.txt
```diff
@@ -0,0 +1,378 @@
...
+      following code.
+
+      .. code-block:: java
+
+         import org.slf4j.Logger;
+         import org.slf4j.LoggerFactory;
+
+         ...
+
+         Logger logger = LoggerFactory.getLogger("MyApp");
```
jyemin, line 185: This is inconsistent with the Logback example, which shows actual driver usage.  I think the intention was to show this simple sort of example here, and show the MongoClient example in the next section, but I'm not sure.

jyemin (changes requested):

source/fundamentals/logging.txt
biniona-mongodb: Thanks for the note. The reason for the difference between the examples for Logback and Log4J2 is because Log4J2 has a default logger level of "[ERROR](https://logging.apache.org/log4j/2.x/manual/configuration.html)", where as Logback has a default logger level of "[DEBUG](https://github.com/qos-ch/logback/blob/master/logback-examples/src/main/resources/chapters/configuration/sample0.xml)" . This means that performing basic operations with the MongoDB Java driver will give logger output for Logback but not for Log4j2. 

There is a way to change the root logger level of Log4J2 [programmatically](https://logging.apache.org/log4j/log4j-2.4/faq.html#reconfig_level_from_code), but it is not supported by the public api. Configuring your logging framework to change the log level through a config file seemed like a separate task, and is the subject of the next section, so I didn't want to introduce it here. The most natural solution to this problem seemed to me to use different examples, as you can only use one log framework at a time. 

I had trouble figuring out how to log an error with the MongoDB Java driver using easy to understand code, so I went with this solution (also discussed in this comment https://github.com/mongodb/docs-java/pull/76#discussion_r618889732). If there is a relatively straightforward way to have the driver log an error, we could swap this or both examples to use that code, but I am not sure whether or not it would improve the experience for learners.

Please let me know if it would have been appropriate for me to have reached out to you directly in regards to the question of "is there a way to have the MongoDB Java driver easily log an error," or if waiting to have this surface in review was the right approach.

source/fundamentals/logging.txt
```diff
@@ -0,0 +1,378 @@
...
+   
+            Add the following dependency to your ``build.gradle`` file.
+   
+            .. code-block:: none
+   
+               dependencies {
+                  implementation 'ch.qos.logback:logback-classic:1.2.3'
+               }
+
+      Once you have installed the above dependency, connect to your
```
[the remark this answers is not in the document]
biniona-mongodb: Thanks for the note. Fix: https://github.com/mongodb/docs-java/pull/76/commits/4dde5414512211d3a3bbbc7051cf08c9b5899b86

Question purely for my own edification (feel free to resolve this conversation without responding if you are too busy to respond): Is there a reason why the term "included" is more appropriate here? I can see "include" as the more appropriate verb for adding items to your dependency list, but is it correct to say that the packages must be "installed" (as per the [Maven lifecycle phases](https://maven.apache.org/guides/getting-started/maven-in-five-minutes.html#maven-phases)) before the example is run?

source/fundamentals/logging.txt
```diff
@@ -0,0 +1,378 @@
...
+
+The makers of the MongoDB Java driver have defined the following logger names to
+organize different logging events in the driver. Here are the logger names
+defined in the driver and the logging events they correspond with. 
+
+* ``org.mongodb.driver`` : root logger for the MongoDB Java driver
+* ``org.mongodb.driver.cluster`` : monitoring of your MongoDB servers
+* ``org.mongodb.driver.connection`` : connections and connection pools
+* ``org.mongodb.driver.protocol`` : protocol message sent to and received from your MongoDB server
+* ``org.mongodb.driver.protocol.insert`` : insert messages and responses
```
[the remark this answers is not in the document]
biniona-mongodb: Thanks for the note, and thank you for reviewing the drivers codebase. To clarify how I understand your suggesting, I am interpreting your comment as "replace what is currently in the logger names list with the logger names I am listing in my suggestion." This removes the "org.mongodb.driver" root logger from the documentation, and removes the individual protocol loggers. Please let me know if I misinterpreted your comment, or if you would like any additional changes.

Fix:

source/fundamentals/logging.txt
```diff
@@ -0,0 +1,378 @@
...
+
+   ...
+
+   Logger logger_parent = LoggerFactory.getLogger("parent");
+   Logger logger_child = LoggerFactory.getLogger("parent.child");
+
+A logger inherits the properties of its ancestor logger and can define
+its own. You can think of this as similar to class inheritance in Java. 
+
+The makers of the MongoDB Java driver have defined the following logger names to
```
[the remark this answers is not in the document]
biniona-mongodb: Thanks for the note. Fix: 

For background around this phrasing (feel free to ignore this comment and resolve if you are not interested): when I was studying logger names, I had trouble telling where logger names were getting defined (logging framework, people using logging framework, or SLF4J). I said "the makers of the mongodb java driver" to try and put a face to which involved party made the following list. In my head that was easier to understand, but in hindsight it does deviate from the majority of our content (I agree that it would apply to everywhere else in the driver), and I agree that taking it out is beneficial.

source/fundamentals/logging.txt
```diff
@@ -0,0 +1,378 @@
...
+Background
+~~~~~~~~~~
+
+The MongoDB Java driver uses the Simple Logging Facade For Java (SLF4J).
+SLF4J allows you to specify your logging framework of choice at deployment time.
+For more information on SLF4J,
+`see the SLF4J documentation <http://www.slf4j.org/>`__.
+
+Setting up a logger is optional. When you start your application the MongoDB
+Java driver looks for the ``slf4j-api`` package in your classpath. If the driver
```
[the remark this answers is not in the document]
biniona-mongodb: Thanks for the note. I need to study my Java/Maven terminology. Great catch. I have read the linked articles, and I hope to not make the same mistake again.

To test my understanding (please feel free to not respond if you are busy), is the following statement correct: The term "package" in java is more related to project organization than dependency management (unlike in Go or Python, where package is intimately tied with code reuse). In Java, "artifact" (JAR file)  is the base unit of inter-project code reuse used by a build tool to handle project dependencies.

biniona-mongodb:
Thanks for the feedback @jyemin. I have implemented your suggested changes and am ready for the next round of review. If you would like to discuss anything synchronously (the discussion around [this comment](https://github.com/mongodb/docs-java/pull/76#discussion_r626647301) might be easier over slack), feel free to reach out at any time.

source/fundamentals/logging.txt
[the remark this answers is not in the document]
jyemin: That sounds about right.  Starting in Java 9 there was introduced the concept of `module` that maps closely to `artifact`, but it hasn't really caught on except internal to the JDK itself.

source/fundamentals/logging.txt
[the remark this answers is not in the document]
jyemin: Yes, that's correct.  We never actually log anything with a "org.mongodb.driver" logger.  It's really just serves as the root.

source/fundamentals/logging.txt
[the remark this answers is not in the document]
jyemin: Installation is part of Maven lifecycle, but it's part of actually executing your Maven pom.xml using `mvn` command.  What we're talking about here is just including the dependency as part of your Maven pom.xml and not actually executing it.

source/fundamentals/logging.txt
jyemin: There's isn't an easy way to force the driver to log an error, as in most case the driver instead just throws an exception.
Given the difference between Logback and Log4J, I don't see an easy way forward. It does seem strange though that there is only one place where we show actual driver-generated log messages, and we only do it for one of the two logging implementations.

Here's one alternative: Do it the log4J way for Logback as well, and add a complete example at the end for each logging implementation, showing how to set logging level for one of the actual loggers used the driver, and a sample message generated by the driver for that level.   It seems to me not going much further than you already have to add configuration of loggers for the driver in the configuration files for each example.

Happy to discuss it over Slack

jyemin (changes requested):
Discussing requested changes on Slack.

source/fundamentals/logging.txt
biniona-mongodb: Slack conversation my takeaways message: "Sounds good, I'll add a mention of the default log level to help readers understand the difference between the examples.
I didn't realize that support required frequent logger-name configuration, but that makes sense. As it is a common use case for clients, I agree it makes sense to include the example explicitly rather than linking to external docs. I'll add the example (hopefully by the end of the day) and ping you on slack when it is ready (let me know if you would rather I ping you on a different channel).

biniona-mongodb:
Hi @jyemin, I implemented your feedback from our slack conversation. I am ready for the next round of review.

source/fundamentals/logging.txt
```diff
@@ -0,0 +1,477 @@
...
+* ``org.mongodb.driver.operation`` : operations, including logging related to automatic retries
+* ``org.mongodb.driver.protocol`` : commands sent to and replies received from MongoDB servers
+* ``org.mongodb.driver.uri`` : connection string parsing
+* ``org.mongodb.driver.management`` : JMX (Java Management Extensions)
+
+Example - Names
+~~~~~~~~~~~~~~~
+
+This example shows how to change the log level for a specific driver logger.
+We set the root logger to OFF and the ``org.mongodb.connection`` logger to
```
jyemin, line 390: org.mongodb.connection -> org.mongodb.driver.connection

source/fundamentals/logging.txt
```diff
@@ -0,0 +1,477 @@
...
+      MongoDB instance and retrieve a document with the following code.
+
+      .. include:: /includes/fundamentals/logging.rst
+
+      You should see output that resembles the following:
+
+      .. code-block:: none
+
+         ...
+         1156 [main] DEBUG o.m.driver.protocol.command - Sending command '{"find": "<your collection>", "filter": {}, "limit": 1, "singleBatch": true, "$db": "<your db>", "lsid": {"id": {"$binary": {"base64": "<server session>", "subType": "04"}}}}' with request id 13 to database <your db> on connection <your connection> to server <your server>
```
jyemin, line 138: I'm surprised we don't also see INFO level logging in this output.

source/fundamentals/logging.txt
```diff
@@ -0,0 +1,477 @@
...
+
+      .. include:: /includes/fundamentals/logging.rst
+
+      You should see output that resembles the following.
+
+      .. code-block:: none
+
+         ...
+         1317 [cluster-ClusterId{value='<your cluster id>', description='null'}-<your connection uri>] INFO  org.mongodb.driver.cluster - Discovered replica set primary <your connection uri>
+         1568 [main] INFO  org.mongodb.driver.connection - Opened connection [connectionId{localValue:7, serverValue:<server value>}] to <your connection uri>
```
jyemin, line 285: I'm surprised we don't see DEBUG level logging here

jyemin (changes requested):
Looking good.  Please double-check that the log output matches the configured log level.  There are a few places where it seems that they don't.

source/fundamentals/logging.txt
biniona-mongodb: Thanks for the note. Fix: https://github.com/mongodb/docs-java/pull/76/commits/352229379217d8a0fd3aaae46b2c5f5f795011f4

I believe that the Logback configuration does produce the correct output (log messages with a log level of INFO or higher) but  does does it incorrectly ([by using a filter rather than changing the root loggers log level](http://logback.qos.ch/manual/filters.html#thresholdFilter)). Great catch, and apologies for missing this.

source/fundamentals/logging.txt
biniona-mongodb: Thanks for the note. Fix: https://github.com/mongodb/docs-java/pull/76/commits/23b4d91497076f64afd53a1813cf90c43f741d28

source/fundamentals/logging.txt
biniona-mongodb: Thanks for the note. I agree that having INFO logging in this output would help illustrate the point for learners. Luckily the third from last from the last line of logging output in the code snippet has a log level of INFO. Adding it now. Fix: https://github.com/mongodb/docs-java/pull/76/commits/6c92b0c431ba28c14dc12aa6b661d65bf2ff9aa9

biniona-mongodb:
Hi @jyemin, I have went through and implemented the requested changes. I am ready for the next round of review. 

I have double checked all examples and I can't find any issues.

Example 1:

- Logback : Log Level Debug (levels are in output are "info", "debug", "debug")
- Log4J2: Log Level Error (levels in output are "error")

Example 2:

- Logback : Log Level Info (levels in output are "info", "info")
- Log4J2: Log Level Info (levels in output are "info","info")

Example 3:

- Logback : Log Level Off, org.mongodb.driver.connection INFO (levels in output are "info", "info" and both have "org.mongodb.driver.connection" log name)  
- Log4J2: Log Level Off, only org.mongodb.driver.connection INFO (levels in output are "info", "info" and both have "org.mongodb.driver.connection" log name)

jyemin (approved):
Thanks for sticking with me on this one.  LGTM.

rozza (approved):

PR #83, opened 2021-05-19
DOCSP-9571 Fundamentals GridFS page

ccho-mongodb:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-9571

https://docs-mongodbcom-staging.corp.mongodb.com/5149172/java/docsworker-xlarge/DOCSP-9571-gridfs/fundamentals/gridfs/

### Self-Review Checklist


source/fundamentals/gridfs.txt
```diff
@@ -0,0 +1,387 @@
...
+referenced above, prefixed with the default bucket name ``fs``, unless
+you specify a different name. The driver also creates an index on each
+collection to ensure efficient retrieval of the files and related
+metadata. For more information on GridFS indexes, see the server manual
+page on :manual:`GridFS Indexes </core/gridfs/#gridfs-indexes>`.
+
+When storing files with GridFS, the driver splits the files into smaller
+chunks, each represented by a separate document in the ``chunks`` collection.
+It also creates a document in the ``files`` collection that contains
+a file id as an ``ObjectId``, file name, and other file metadata. You can
```
nathan-contino-mongo, line 55: Nit: since we don't specify the type of the file name or other file metadata, perhaps we can omit "as an ObjectId" here.

source/fundamentals/gridfs.txt
```diff
@@ -0,0 +1,387 @@
...
+collection in the specified bucket and uses the information to reconstruct
+the file from documents in the ``chunks`` collection. You can read the file
+into memory or output it to a stream.
+
+.. _gridfs-create-bucket:
+
+Create a GridFS Bucket
+----------------------
+
+To use GridFS, first create a bucket or reference an existing one on
```
nathan-contino-mongo, line 72:
```suggestion
To use GridFS, create a bucket or reference an existing one on
```
I'm not sure if we need "first" here since this statement basically says "to use GridFS, you need a GridFS bucket". I guess the "first" is supposed to modify both "create a bucket" and "reference an existing one", but to me it reads as if it's only modifying "create a bucket" and I was tripped up by the fact that there wasn't really another step there. Here's an idea for how you can break this up to avoid that confusion:

```
To use GridFS, you'll need a GridFS bucket. You can either create a bucket or
reference an existing one in a MongoDB database. Call the...
```

source/fundamentals/gridfs.txt
```diff
@@ -0,0 +1,387 @@
...
+a MongoDB database. Call the ``GridFSBuckets.create()`` helper method with
+a database instance as the parameter to instantiate a ``GridFSBucket``.
+You can use the ``GridFSBucket`` instance to call read and write operations on
+the files in your bucket.
```
nathan-contino-mongo, line 76: Perhaps I just didn't read this section closely enough, but I was a bit confused at how ``GridFSBuckets.create(database)`` actually works. I guess it's sort of like databases in MongoDB, which you normally just lazily create when you insert documents into them? Or does it actually explicitly instantiate a GridFS bucket, but grab a reference to the existing bucket if it already exists? Maybe we should call out that behavior more explicitly.

source/fundamentals/gridfs.txt
```diff
@@ -0,0 +1,387 @@
...
+
+This section shows you how to upload a file to a GridFS bucket using an input
+stream. The code example below shows how you can use a ``FileInputStream`` to
+read data from a file in your filesystem and upload it to GridFS by performing
+the following operations:
+
+- Read from the filesystem using a ``FileInputStream``.
+- Set the chunk size using ``GridFSUploadOptions``.
+- Set a custom metadata field called ``type`` to the value "zip archive".
+- Upload the file, specifying the file name as "myProject.zip".
```
nathan-contino-mongo, line 135: In the code snippet, I see the file path "/path/to/project.zip". Should this line instead read "project.zip"? If not, can we add a note or something explaining why the name changed? Unless I missed something.

source/fundamentals/gridfs.txt
```diff
@@ -0,0 +1,387 @@
...
+following code snippet:
+
+.. literalinclude:: /includes/fundamentals/code-snippets/GridFSOperations.java
+   :language: java
+   :dedent:
+   :start-after: start uploadOptions
+   :end-before: end uploadOptions
+
+See the :java-docs:`GridFSUploadOptions <apidocs/mongodb-driver-core/com/mongodb/client/gridfs/model/GridFSUploadOptions.html>`
+API documentation for more information.
```
nathan-contino-mongo, line 122: I wonder if we should try to give any guidance as to which one of these is most performance, robust, debuggable, extensible, etc. -- if I'm a user reading this page, it's helpful to see both options, but I would probably like a recommendation of which I should use as a sane default as well in case it looks like both would work for my use case.

source/fundamentals/gridfs.txt
```diff
@@ -0,0 +1,387 @@
...
+-------------------------
+
+In this section, you can learn how to retrieve file metadata stored in the
+``files`` collection of the GridFS bucket. The metadata includes information
+about the file it refers to including:
+
+- Id of the file
+- Name of the file
+- Length/size of the file
+- Upload date and time
```
nathan-contino-mongo, line 188: nit: since these aren't full sentence bullet points and "Id" looks *very* strange to me with a capital I... would it make sense to use lowercase letters at the beginning of these instead?

source/fundamentals/gridfs.txt
```diff
@@ -0,0 +1,387 @@
...
+:java-docs:`GridFSUploadStream <apidocs/mongodb-driver-sync/com/mongodb/client/gridfs/GridFSUploadStream.html>`.
+
+.. _gridfs-retrieve-file-info:
+
+Retrieve File Information
+-------------------------
+
+In this section, you can learn how to retrieve file metadata stored in the
+``files`` collection of the GridFS bucket. The metadata includes information
+about the file it refers to including:
```
nathan-contino-mongo, line 183:
```suggestion
about the file it refers to, including:
```

source/fundamentals/gridfs.txt
```diff
@@ -0,0 +1,387 @@
...
+of a file in an instance of ``GridFSDownloadOptions``:
+
+.. code-block:: java
+
+   GridFSDownloadOptions downloadOptions = new GridFSDownloadOptions().revision(1);
+
+For more information on the enumeration of revisions, see the API
+documentation for :java-docs:`GridFSDownloadOptions <apidocs/mongodb-driver-core/com/mongodb/client/gridfs/model/GridFSDownloadOptions.html>`.
+
+Download a File to an Output Stream
```
nathan-contino-mongo, line 262: This might get a little too wordy, but I think it would be easier to understand the difference between this section and the next section if the titles referred to the use case for input/output streams instead of just the type of stream. For instance:

"Write GridFS Data to a File Using an Output Stream"
and
"Write GridFS Data to Application Memory Using an Input Stream"

source/fundamentals/gridfs.txt
```diff
@@ -0,0 +1,387 @@
...
+:java-docs:`rename() <apidocs/mongodb-driver-sync/com/mongodb/client/gridfs/GridFSBucket.html#rename(org.bson.types.ObjectId,java.lang.String)>`
+API documentation.
+
+.. _gridfs-delete-files:
+
+Delete Files
+------------
+
+You can remove a file from your GridFS bucket by calling the ``delete()``
+method. You must specify the file by the value of its file id rather than
```
nathan-contino-mongo, line 337: nit: in this section, you say "specify the file by the value of its file id", but in "Rename Files" you say "specify the file to rename by its file id" (omitting "the value of"). This doesn't hurt readability, but it did make me double-take to make sure the instructions were actually referring to the same thing, so perhaps we should use the same wording for each of those.

source/fundamentals/gridfs.txt
```diff
@@ -0,0 +1,387 @@
...
+its file name.
+
+.. note::
+
+   The ``delete()`` method only supports deleting one file at a time. To
```
nathan-contino-mongo, line 342: Similar as above: rename uses the wording "only supports updating the name of a single file", maybe we should use the same wording here "only supports  deleting a single file" (or vice versa).

nathan-contino-mongo (approved):
LGTM with some very small scattered nitpicks. Awesome job with a pretty huge piece of writing.

source/fundamentals/gridfs.txt
ccho-mongodb: Good idea.
I wasn't certain whether it was an important detail since we don't call this out for other lazily-created entities like Databases and Collections. While it might be beyond the level of behind-the-scenes type information we usually include, I'll add it since that info probably isn't available anywhere else.

source/fundamentals/gridfs.txt
ccho-mongodb: Are you asking whether Input or Output streams are better for performance/robustness/etc.? I think those are just two different ways to move data, and performance/robustness/etc. depends more on the Java built-in `io` package implementation.

source/fundamentals/gridfs.txt
ccho-mongodb: I have seen "/path/to" prefixed to describe the filesystem path in various MongoDB documentation without additional explanation. If this doesn't seem apparent, I can remove it.

source/fundamentals/gridfs.txt
ccho-mongodb: Our style guide mentions these should be capitalized: https://docs.mongodb.com/meta/style-guide/style/lists/#all-list-items-are-fragments

source/fundamentals/gridfs.txt
nathan-contino-mongo: Fair enough -- if there really isn't much difference in usability or anything, seems fair enough to leave this as is with no explicit recommended default.

source/fundamentals/gridfs.txt
ccho-mongodb: Updating the rename() note.

source/fundamentals/gridfs.txt
ccho-mongodb: Oh ok, sorry if I misread. I'll clarify source and destination files don't have the same name and do not need to. I wouldn't want the reader to mistakenly think the filenames had to be the same.

nathan-contino-mongo: ahh it's actually my misunderstanding -- I didn't realize that "specifying the file name as" applied to the file as saved in GridFS (which I suppose should be obvious, since the step starts with "upload the file"). Personally I lay the blame at the feet of whoever decided that documents stored in GridFS are just called files instead of gFiles or something.

source/fundamentals/gridfs.txt
ccho-mongodb: I thought about this comment for a while. I wonder if my assumption that working with input and output streams are understood when working with reading/writing files might not be suitable for new developers. I'm going to pass on this for now and try to think of a way to introduce streams more clearly.

source/fundamentals/gridfs.txt
nathan-contino-mongo: seeing "Id" with a capital I is so jarring to me that I would lean towards adding the article, but use your judgment here -- it's a very subjective call.

ccho-mongodb:
Hi @rozza and @jyemin, this PR on GridFS is ready for your review when you get the chance. Thanks!

jyemin:
@rozza can you take this one?

source/fundamentals/gridfs.txt
```diff
@@ -0,0 +1,395 @@
...
+defined in the GridFS specification:
+
+- The ``chunks`` collection stores the binary file chunks.
+- The ``files`` collection stores the file metadata.
+
+When you create a new GridFS bucket, the driver creates the collections
+referenced above, prefixed with the default bucket name ``fs``, unless
+you specify a different name. The driver also creates an index on each
+collection to ensure efficient retrieval of the files and related
+metadata. For more information on GridFS indexes, see the server manual
```
rozza, line 49: Note: Indexes will only be added if the files collection doesn't exist or is empty.

See: https://github.com/mongodb/specifications/blob/master/source/gridfs/gridfs-spec.rst#before-write-operations

source/fundamentals/gridfs.txt
```diff
@@ -0,0 +1,395 @@
...
+
+.. literalinclude:: /includes/fundamentals/code-snippets/GridFSOperations.java
+   :language: java
+   :dedent:
+   :start-after: start uploadFromOutputStream
+   :end-before: end uploadFromOutputStream
+
+This code example prints the file id of the uploaded file after it is
+successfully saved in GridFS.
+
```
rozza, line 179: Suggestion add note about failure modes when uploading to GridFS and users having to clean up any orphaned chunks.

> Operation Failure
> If any of the above operations fail against the server, drivers MUST raise an error. If some inserts succeeded before the failed operation, these become orphaned chunks. Drivers MUST NOT attempt to clean up these orphaned chunks. The rationale is that whatever failure caused the orphan chunks will most likely also prevent cleaning up the orphaned chunks, and any attempts to clean up the orphaned chunks will simply cause long delays before reporting the original failure to the application.

See: https://github.com/mongodb/specifications/blob/master/source/gridfs/gridfs-spec.rst#file-upload

rozza (changes requested):
Its nearly there, I've added some comments / suggestions

source/includes/fundamentals/code-snippets/GridFSOperations.java
```diff
@@ -0,0 +1,162 @@
...
+                .metadata(new Document("myField", "myValue"));
+        // end uploadOptions
+    }
+
+    private static void uploadFromInputStream(GridFSBucket gridFSBucket) throws Exception {
+        // start uploadFromInputStream
+        String filePath = "/path/to/project.zip";
+        try (InputStream streamToUploadFrom = new FileInputStream(filePath) ) {
+            GridFSUploadOptions options = new GridFSUploadOptions()
+                    .chunkSizeBytes(1024)
```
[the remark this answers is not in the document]
ccho-mongodb: Thanks. I had read this, but noticed [GridFSTour.java](https://github.com/mongodb/mongo-java-driver/blob/master/driver-sync/src/examples/gridfs/GridFSTour.java#L78) had specified 1k, so wasn't sure if it mattered with WT. Will update to 1MB similar to my first UploadOptions example.

source/fundamentals/gridfs.txt
ccho-mongodb: I'll add a paragraph on orphaned chunks. Would you recommend nesting a try block within the try-with-resources around the "write" method so that you can clean up using the GridFSUploadStream's abort() instance method in the catch block? Or that that they manually clean up orphaned chunks by directly querying and deleting them from the chunks collection?

ccho-mongodb:
@rozza addressed your feedback. Had one question on handling failure https://github.com/mongodb/docs-java/pull/83#discussion_r637088570

source/includes/fundamentals/code-snippets/GridFSOperations.java
```diff
@@ -0,0 +1,168 @@
...
+
+        GridFSUploadOptions options = new GridFSUploadOptions()
+                .chunkSizeBytes(1048576)
+                .metadata(new Document("type", "zip archive"));
+
+        try (GridFSUploadStream uploadStream = gridFSBucket.openUploadStream("myProject.zip", options)) {
+            try {
+                uploadStream.write(data);
+                System.out.println("The file id of the uploaded file is: " + uploadStream.getObjectId().toHexString());
+            } catch (MongoGridFSException e) {
```
rozza, line 75: I think this won't work in the way you'd expect. 

The exception would come from  an insertOne and wouldn't be wrapped in a MongoGridFSException.  The MongoGridFSException would only happen if  `uploadStream.close()` had been called and then the abort call wouldn't work.

rozza: Catching Exception or MongoException maybe better. But if the insert failed there is a chance calling abort() will also fail.

rozza (changes requested):
I'm not sure the best course of action for cleaning up any orphaned chunks.  See the comments in the code sample.

It may be worth just adding it as a note that users are responsible for cleaning up any failed uploads. What do you think?

source/includes/fundamentals/code-snippets/GridFSOperations.java
ccho-mongodb: Will remove per our discussion on Slack.

rozza (approved):

PR #87, opened 2021-05-25
DOCSP-15781: Versioned API

ccho-mongodb:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-15781

https://docs-mongodbcom-staging.corp.mongodb.com/4d2fe1e/java/docsworker-xlarge/DOCSP-15781-versioned-api/fundamentals/versioned-api/

### Self-Review Checklist

### Notes for the Reviewer:

The API links assume that the `:java-docs:` target points to the 4.3 version. Currently, some of the links do not work because the `:java-docs:` target still points to the 4.2 version.

### DIFF
--- a/source/fundamentals/versioned-api.txt
+++ b/source/fundamentals/versioned-api.txt
@@ -14,8 +14,7 @@ Versioned API
 
    The Versioned API feature requires MongoDB Server 5.0 or later. You
    should only use the Versioned API feature if you are connecting to
-   MongoDB servers that are upgraded to a version that supports the
-   Versioned API feature.
+   MongoDB servers that are upgraded to a version that supports this feature.
 
 This guide shows you how to specify the **Versioned API** when connecting to
 a MongoDB instance or replica set. You can use the Versioned API feature to
@@ -34,8 +33,8 @@ covered by the Versioned API.
 The following sections describe how you can enable Versioned API for
 your MongoDB client and each of the options that you can specify.
 
-Specify the API Version
------------------------
+Enable Versioned API
+--------------------
 
 To enable the Versioned API, you must specify an API version in the settings
 that you use to create your MongoDB client. Once you instantiate a
@@ -45,13 +44,14 @@ run commands not covered by the Versioned API or with a different version of
 it, instantiate a separate client.
 
 The example below shows how you can instantiate a ``MongoClient`` that
-specifies version "1" of the Versioned API by performing the following
-operations:
+specifies version "1" of the Versioned API and connects to a server by 
+performing the following operations:
 
 - Construct a ``ServerApi`` instance using the ``ServerApi.Builder``
   helper class.
 - Construct a ``MongoClientSettings`` instance using the
   ``MongoClientSettings.Builder`` class.
+- Specify a server to connect to using a ``ServerAddress`` instance.
 - Instantiate a ``MongoClient`` using the ``MongoClients.create()`` method
   and pass your ``MongoClientSettings`` instance as a parameter.
 
@@ -63,11 +63,11 @@ operations:
 
 .. warning::
 
-   Make sure that you are connecting to a version of MongoDB server that
-   supports Versioned API. Otherwise, your application may throw an exception
-   when attempting to perform any operation that sends a command to
-   MongoDB. For example, a query to a server could return an exception
-   message that resembles the following:
+   Make sure that when you specify an API version that you connect to a 
+   version of MongoDB server that supports Versioned API. Otherwise, your
+   application may raise an exception when attempting to perform any operation 
+   that sends a command to MongoDB. For example, a query to a server could 
+   return an exception message that resembles the following:
 
    .. code-block:: none
 
@@ -78,6 +78,7 @@ section, see the following API documentation:
 
 - :java-docs:`ServerApi <apidocs/mongodb-driver-core/com/mongodb/ServerApi.html>`
 - :java-docs:`ServerApi.Builder <apidocs/mongodb-driver-core/com/mongodb/ServerApi.Builder.html>`
+- :java-docs:`ServerAddress <apidocs/mongodb-driver-core/com/mongodb/ServerAddress.html>`
 - :java-docs:`MongoClientSettings <apidocs/mongodb-driver-core/com/mongodb/MongoClientSettings.html>`
 - :java-docs:`MongoClientSettings.Builder <apidocs/mongodb-driver-core/com/mongodb/MongoClientSettings.Builder.html>`
 - :java-docs:`MongoClients.create() <apidocs/mongodb-driver-sync/com/mongodb/client/MongoClients.html#create(com.mongodb.MongoClientSettings)>`
--- a/source/fundamentals/versioned-api.txt
+++ b/source/fundamentals/versioned-api.txt
@@ -12,26 +12,29 @@ Versioned API
 
 .. note::
 
-   The Versioned API feature requires MongoDB Server 5.0 or later. You
-   should only use the Versioned API feature if you are connecting to
-   MongoDB servers that are upgraded to a version that supports this feature.
+   The Versioned API feature requires MongoDB Server 5.0 or later.
+
+   You should only use the Versioned API feature if all the MongoDB
+   servers you are connecting to are upgraded to a version that supports
+   this feature.
 
 This guide shows you how to specify the **Versioned API** when connecting to
 a MongoDB instance or replica set. You can use the Versioned API feature to
-force the server to run operations with consistent behavior as specified by
-an **API version**. An API version defines expected behavior of operations
-such as the information expected in the request and the response.
+force the server to run operations with behavior compatible with the
+specified **API version**. An API version defines the expected behavior of the
+operations it covers and the format of server responses. If you change to
+a different API version, the operations are not guaranteed to be
+compatible and the server responses are not guaranteed to be similar.
 
-When you use the Versioned API feature, you can update your MongoDB driver or
-server without worrying about backward compatibility issues of the commands
-covered by the Versioned API.
+When you use the Versioned API feature with an official MongoDB driver, you
+can update your driver or server without worrying about backward compatibility
+issues of the commands covered by the Versioned API.
 
-..
-   TODO: See the server manual page on `Versioned API <https://docs.mongodb.com/v5.0/reference/versioned-api/>`__
-   for more information including the set the commands covered.
+See the server manual page on `Versioned API <https://docs.mongodb.com/v5.0/reference/versioned-api/>`__
+for more information including a list of commands it covers.
 
 The following sections describe how you can enable Versioned API for
-your MongoDB client and each of the options that you can specify.
+your MongoDB client and the options that you can specify.
 
 Enable Versioned API on a MongoDB Client
 ----------------------------------------
@@ -44,11 +47,13 @@ run commands not covered by the Versioned API or with a different version of
 it, instantiate a separate client.
 
 The example below shows how you can instantiate a ``MongoClient`` that
-specifies version "1" of the Versioned API and connects to a server by
-performing the following operations:
+sets the Versioned API version and connects to a server by performing the
+following operations:
 
 - Construct a ``ServerApi`` instance using the ``ServerApi.Builder``
   helper class.
+- Specify a Versioned API version using a constant from the
+  ``ServerApiVersion`` class.
 - Construct a ``MongoClientSettings`` instance using the
   ``MongoClientSettings.Builder`` class.
 - Specify a server to connect to using a ``ServerAddress`` instance.
@@ -63,11 +68,12 @@ performing the following operations:
 
 .. warning::
 
-   Make sure that when you specify an API version that you connect to a
-   version of MongoDB server that supports Versioned API. Otherwise, your
-   application may raise an exception when attempting to perform any operation
-   that sends a command to MongoDB. For example, a query to a server could
-   return an exception message that resembles the following:
+   If you specify an API version and connect to a MongoDB server that does
+   not support Versioned API, your application may raise an exception when
+   executing a command on your MongoDB server. If you use
+   a ``MongoClient`` that specifies the API version to query a server that
+   does not support it, your query could fail with the following exception
+   message:
 
    .. code-block:: none
       :copyable: false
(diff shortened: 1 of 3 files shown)


source/fundamentals/versioned-api.txt
```diff
@@ -0,0 +1,123 @@
...
+To enable the Versioned API, you must specify an API version in the settings
+that you use to create your MongoDB client. Once you instantiate a
+``MongoClient`` instance with a specified Versioned API, all commands you
+run with that client use that version of the Versioned API. If you need to
+run commands not covered by the Versioned API or with a different version of
+it, instantiate a separate client.
+
+The example below shows how you can instantiate a ``MongoClient`` that
+specifies version "1" of the Versioned API and connects to a server by
+performing the following operations:
```
kyuan-mongodb, line 48: Nit: When I didn't see 1 on its own in the snippet, I was a bit confused.
```suggestion
specifies version "V1" of the Versioned API and connect to a server by:
```

source/fundamentals/versioned-api.txt
```diff
@@ -0,0 +1,123 @@
...
+
+- Construct a ``ServerApi`` instance using the ``ServerApi.Builder``
```
kyuan-mongodb, line 50: Tying to the previous comments suggestion, I think the first word of each bullet should be a verb.

Construct -> Constructing

source/fundamentals/versioned-api.txt
```diff
@@ -0,0 +1,123 @@
...
+   :language: java
+   :dedent:
+
+.. warning::
+
+   Make sure that when you specify an API version that you connect to a
+   version of MongoDB server that supports Versioned API. Otherwise, your
+   application may raise an exception when attempting to perform any operation
+   that sends a command to MongoDB. For example, a query to a server could
+   return an exception message that resembles the following:
```
kyuan-mongodb, line 70: I had to read this section a few times to understand this warning. I think it can be simplified to something along the lines of:
```suggestion
   If you specify an API version to a MongoDB server that does not support 
   Versioned API, your application may raise an exception when attempting 
   to send a command to MongoDB. For example, a query to a server could
   return an exception message that resembles the following:
```

source/fundamentals/versioned-api.txt
```diff
@@ -0,0 +1,123 @@
...
+- :java-docs:`ServerAddress <apidocs/mongodb-driver-core/com/mongodb/ServerAddress.html>`
+- :java-docs:`MongoClientSettings <apidocs/mongodb-driver-core/com/mongodb/MongoClientSettings.html>`
+- :java-docs:`MongoClientSettings.Builder <apidocs/mongodb-driver-core/com/mongodb/MongoClientSettings.Builder.html>`
+- :java-docs:`MongoClients.create() <apidocs/mongodb-driver-sync/com/mongodb/client/MongoClients.html#create(com.mongodb.MongoClientSettings)>`
+- :java-docs:`MongoClient <apidocs/mongodb-driver-sync/com/mongodb/client/MongoClient.html>`
+
+Additional Options
+------------------
+
+You can set additional options to enable or disable behavior related to
```
kyuan-mongodb, line 91: Nit: There didn't seem to be options prior to this, so saying additional options may make the reader think there was.
```suggestion
You can set options to enable or disable behavior related to
```

source/fundamentals/versioned-api.txt
```diff
@@ -0,0 +1,123 @@
...
+       | For more information, see the :java-docs:`strict() <apidocs/mongodb-driver-core/com/mongodb/ServerApi.Builder.html#strict(boolean)>` API documentation.
+
+   * -  DeprecationErrors
+     - | **Optional**. When set, if you call a command that is deprecated in the declared API version, the driver raises an exception.
+       |
+       | Default: **false**
+       | For more information, see the :java-docs:`deprecationErrors() <apidocs/mongodb-driver-core/com/mongodb/ServerApi.Builder.html#>` API documentation.
+
+
+The following code example shows how you can set the two options on an
```
kyuan-mongodb, line 115:
```suggestion
The following example shows how you can set the two options on an
```

kyuan-mongodb (changes requested):
Overall, a great page since I was able to learn about Versioned API. Just some comments and nits around simplifying and clarifying.

source/fundamentals/versioned-api.txt
ccho-mongodb: Thanks for pointing this out. I'll update the section title too. I think I had considered the "version" to be an option and "strict" and "deprecationErrors" to be additional. Since "version" is required, I can see how referring to it as an option could be confusing.

source/fundamentals/versioned-api.txt
ccho-mongodb: Not sure what the previous comments suggestion is (I currently see "Nit: When I didn't see 1 on its own"), "construct" is a verb which means "to build". Construct also has a noun form which means "a concept or idea", but that doesn't fit in the context nor sentence structure. The form you recommended, "constructing" is a gerund which actually makes it function as a noun. The part of speech that I'm using across the bullets are verbs in this list.

source/fundamentals/versioned-api.txt
ccho-mongodb: Good catch. The actual value is "1" whereas the name of the class variable is "V1". I'll add some text to clarify.

source/fundamentals/versioned-api.txt
```diff
@@ -0,0 +1,123 @@
...
+   MongoDB servers that are upgraded to a version that supports this feature.
+
+This guide shows you how to specify the **Versioned API** when connecting to
+a MongoDB instance or replica set. You can use the Versioned API feature to
+force the server to run operations with consistent behavior as specified by
+an **API version**. An API version defines expected behavior of operations
+such as the information expected in the request and the response.
+
+When you use the Versioned API feature, you can update your MongoDB driver or
+server without worrying about backward compatibility issues of the commands
```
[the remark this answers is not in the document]
ccho-mongodb: This one is tricky. It's not describing a MongoDB instance, but specifically the server binary you installed which is why it's listed together with the driver ("you can update... driver or server"). I don't really know whether "update your MongoDB instance" adds any confusion, but am not sure it's helpful.

source/fundamentals/versioned-api.txt
ccho-mongodb: Thanks, I agree it's confusing and needs to be reworded. Will fix to something similar to your suggestion.

source/fundamentals/versioned-api.txt
```diff
@@ -0,0 +1,123 @@
...
+   :class: singlecol
+
+.. note::
+
+   The Versioned API feature requires MongoDB Server 5.0 or later. You
+   should only use the Versioned API feature if you are connecting to
+   MongoDB servers that are upgraded to a version that supports this feature.
+
+This guide shows you how to specify the **Versioned API** when connecting to
+a MongoDB instance or replica set. You can use the Versioned API feature to
```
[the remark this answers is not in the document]
ccho-mongodb: I think we use replica set throughout our docs with the assumption that there's some set of terminology that we expect the reader to know. I don't know whether knowing this particular term is a valid assumption, but I think there are probably some terms we need to consider explaining or omitting explanation of.

E.g. From the Node landing page https://docs.mongodb.com/drivers/node/current/ "Connection Guide: connect to a MongoDB instance or replica set". From a quick search, I see it's also in a few documents in our Java docs.

For that reason, I think it's ok to leave there. Feel free to docket this for discussion when we determine what edits we need to make for https://jira.mongodb.org/browse/DOCSP-15827

source/fundamentals/versioned-api.txt
```diff
@@ -0,0 +1,123 @@
...
+force the server to run operations with consistent behavior as specified by
+an **API version**. An API version defines expected behavior of operations
+such as the information expected in the request and the response.
```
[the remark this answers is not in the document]
ccho-mongodb: I would say "parameters" and format. Parameters is in quotes because it's not just the method signature (user parameters), but also how that data gets shared by the driver/server (Presentation and Application Layers).
I'm not sure how to improve upon how I've generalized this, but am open to suggestions.

source/fundamentals/versioned-api.txt
```diff
@@ -0,0 +1,123 @@
...
+.. note::
+
+   The Versioned API feature requires MongoDB Server 5.0 or later. You
+   should only use the Versioned API feature if you are connecting to
+   MongoDB servers that are upgraded to a version that supports this feature.
+
+This guide shows you how to specify the **Versioned API** when connecting to
+a MongoDB instance or replica set. You can use the Versioned API feature to
+force the server to run operations with consistent behavior as specified by
+an **API version**. An API version defines expected behavior of operations
```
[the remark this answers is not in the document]
ccho-mongodb: Good catch, I agree that the current wording is confusing. I think the suggestion might not be the change I'm looking for though since it emphasizes behavior specified by the API version instead of the operations for that version of the API.
My sentence might also be inaccurate from reading the C# documentation on Versioned API [here](https://mongodb.github.io/mongo-csharp-driver/2.13/reference/driver/versioned_api/):

> During communication with a server, clients with a declared API version will force the server to behave in a manner compatible with the API version. eclaring an API version on a client can be used to ensure consistent responses from a server, providing long term API stability for an application.

If that documentation is correct, I should emphasize compatible behavior for operations rather than consistent, and consistent responses from the server.

source/fundamentals/versioned-api.txt
kyuan-mongodb: I meant if you changed line 47-48 to end with: "and connect to a server by:"

When I see lines end this way, I expect each bullet to start with an action (ending in "ing").

source/fundamentals/versioned-api.txt
```diff
@@ -0,0 +1,123 @@
...
+   :local:
+   :backlinks: none
+   :depth: 1
+   :class: singlecol
+
+.. note::
+
+   The Versioned API feature requires MongoDB Server 5.0 or later. You
+   should only use the Versioned API feature if you are connecting to
+   MongoDB servers that are upgraded to a version that supports this feature.
```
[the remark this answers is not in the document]
kyuan-mongodb: I think putting it in 2 different paragraphs help

source/fundamentals/versioned-api.txt
ccho-mongodb: I wasn't able to figure out if there was something wrong with its current form. I think the style guide has a note about avoiding fragments for the introductory text. It doesn't seem like there's an example, but the note says:

> Avoid using fragments to introduce lists. Fragments can be harder to understand than sentences.

The version you suggested "and connect to a server by" is a fragment since it's not a complete sentence by itself. In the suggested version, the fragment depends on the list item fragment to complete the sentence.

source/fundamentals/versioned-api.txt
```diff
@@ -0,0 +1,129 @@
...
+for more information including a list of commands it covers.
+
+The following sections describe how you can enable Versioned API for
+your MongoDB client and the options that you can specify.
+
+Enable Versioned API on a MongoDB Client
+----------------------------------------
+
+To enable the Versioned API, you must specify an API version in the settings
+that you use to create your MongoDB client. Once you instantiate a
```
kyuan-mongodb, line 42:
```suggestion
of your MongoDB client. Once you instantiate a
```

kyuan-mongodb, line 42: Nit: I prefer this wording, but if it makes the sentence inaccurate, leave it as is.
```suggestion
of your MongoDB client. Once you instantiate a
```

source/fundamentals/versioned-api.txt
```diff
@@ -0,0 +1,129 @@
...
+.. literalinclude:: /includes/fundamentals/code-snippets/VersionedApiExample.java
+   :start-after: start serverAPIVersion
+   :end-before: end serverAPIVersion
+   :language: java
+   :dedent:
+
+.. warning::
+
+   If you specify an API version and connect to a MongoDB server that does
+   not support Versioned API, your application may raise an exception when
```
kyuan-mongodb, line 71: Nit: Saying "may raise" makes me wonder why it might or might not be raised.
```suggestion
   not support Versioned API, your application raises an exception when
```

kyuan-mongodb (approved):
LGTM with two nits

source/fundamentals/versioned-api.txt
ccho-mongodb: This is based on my observation rather than any rule in the Versioned API spec, so I think it would be better not to make it a definitive statement.

ccho-mongodb:
Hi @jyemin and @rozza , this page on Versioned API is ready for review. Could you take a look when you get the chance? Thanks!

source/fundamentals/versioned-api.txt
```diff
@@ -0,0 +1,128 @@
...
+your MongoDB client and the options that you can specify.
+
+Enable Versioned API on a MongoDB Client
+----------------------------------------
+
+To enable the Versioned API, you must specify an API version in the settings
+of your MongoDB client. Once you instantiate a ``MongoClient`` instance with
+a specified Versioned API, all commands you run with that client use that
+version of the Versioned API. If you need to run commands not covered by the
+Versioned API or with a different version of it, instantiate a separate client.
```
jyemin, line 45: > If you need to run commands not covered by theVersioned API 

You can just leave apiStrict off for this case

source/includes/fundamentals/code-snippets/VersionedApiExample.java
```diff
@@ -0,0 +1,41 @@
...
+        // end apiVersionOptions
+    }
+
+    public static void main(String[] args) {
+        // start serverAPIVersion
+        ServerApi serverApi = ServerApi.builder()
+                .version(ServerApiVersion.V1)
+                .build();
+
+        ServerAddress serverAddress = new ServerAddress("localhost", 27017);
```
jyemin, line 30: I think it's more common to use a connection string and use MongoClientSetting.Builder.applyConnectionString

source/fundamentals/versioned-api.txt
```diff
@@ -0,0 +1,128 @@
...
+   not support Versioned API, your application may raise an exception when
+   executing a command on your MongoDB server. If you use
+   a ``MongoClient`` that specifies the API version to query a server that
+   does not support it, your query could fail with the following exception
+   message:
+
+   .. code-block:: none
+      :copyable: false
+
+      com.mongodb.MongoQueryException: Query failed with error code 9 and error message 'Failed to parse: {...} Unrecognized field 'apiVersion'.' on server
```
jyemin, line 79: I'm surprised this is the exception you get.  What operation did you try that threw this exception?

jyemin: Maybe try a few different operations, both reads and writes, and see what you get.

jyemin (changes requested):

rozza:
I'm out for a number of days this week so removing myself from this review.

source/fundamentals/versioned-api.txt
ccho-mongodb: Just want to make sure we're referring to the same thing. If you're using a server version older than 5.0 (doesn't support Versioned API) and run commands on a client, you'll receive an error that complains about the unrecognized field apiVersion.

For `find()`:
```
com.mongodb.MongoQueryException: Query failed with error code 9 and error message 'Failed to parse: {...} Unrecognized field 'apiVersion'.' on server
```

For insertOne()
```
com.mongodb.MongoCommandException: Command failed with error 40415 (Location40415): 'BSON field 'insert.apiVersion' is an unknown field.' on server localhost:27017. The full response is {"ok": 0.0, "errmsg": "BSON field 'insert.apiVersion' is an unknown field.", "code": 40415, "codeName": "Location40415"}
```
For aggregation:
```
com.mongodb.MongoCommandException: Command failed with error 9 (FailedToParse): 'unrecognized field 'apiVersion'' on server localhost:27017. The full response is {"ok": 0.0, "errmsg": "unrecognized field 'apiVersion'", "code": 9, "codeName": "FailedToParse"}
```

I'll update the message so it only mentions unrecognized field apiVersion unless there's some other error that you expected.

ccho-mongodb:
Hi @jyemin , I think I've addressed all your comments. Let me know what you think!

source/fundamentals/versioned-api.txt
jyemin: I'm still a bit confused by this.  There are two paragraphs under the tip, both of which are correct.  But the one not under the tip repeats parts of both tip paragraphs, and offers slightly different advice.

I think this paragraph could just be removed?

jyemin (changes requested):

source/fundamentals/versioned-api.txt
ccho-mongodb: Sorry, forgot to remove this after moving it to the tip section.

PR #88, opened 2021-05-26
Docsp 9573 monitoring

biniona-mongodb:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-9573

https://docs-mongodbcom-staging.corp.mongodb.com/77ca00d/java/docsworker-xlarge/DOCSP-9573-Monitoring/fundamentals/monitoring/

### Self-Review Checklist


source/fundamentals/monitoring.txt
```diff
@@ -0,0 +1,406 @@
...
+Cluster Events
+~~~~~~~~~~~~~~
+
+A cluster event in the driver corresponds to a change in the state of the
+MongoDB instance or cluster you have connected the driver to. Cluster events are
+also called server discovery and monitoring (SDAM) events. 
+ 
+The driver defines nine cluster events. The driver divides these nine events
+between three separate listener interfaces which each listen for three of the
+nine events:
```
kyuan-mongodb, line 165:
```suggestion
The driver defines nine cluster events and divides them into three listener interfaces. Each interface listens for three of the nine events:
```

source/fundamentals/monitoring.txt
```diff
@@ -0,0 +1,406 @@
...
+
+- ``ClusterListener``: :spec:`topology </server-discovery-and-monitoring/server-discovery-and-monitoring.rst#topology>`
```
kyuan-mongodb, line 167: Suggestion: As is, it seems somewhat abrupt. I didn't realize you were describing what each listener listens for.
Applies to all instances.
```suggestion
- ``ClusterListener``: listens for :spec:`topology </server-discovery-and-monitoring/server-discovery-and-monitoring.rst#topology>`
```

source/includes/fundamentals/code-snippets/Monitoring.java
```diff
@@ -0,0 +1,169 @@
...
+                MongoClientSettings.builder()
+                        .applyConnectionString(URI)
+                        .applyToClusterSettings(builder ->
+                                builder.addClusterListener(clusterListener))
+                        .build();
+        MongoClient mongoClient = MongoClients.create(settings);
+        MongoDatabase database = mongoClient.getDatabase(DATABASE);
+        MongoCollection<Document> collection = database.getCollection(COLLECTION);
+        // Run a command to trigger a ClusterDescriptionChangedEvent event
+        collection.find().first();
```
kyuan-mongodb, line 61: Question: Your output says it's also able to write to the server. But my understanding of this event is that it only prints if an operation was performed. So the message about being able to read makes sense to me, but where does the write message come from?

source/fundamentals/monitoring.txt
```diff
@@ -0,0 +1,406 @@
...
+  related events
+- ``ServerListener``: events related to ``mongod`` or ``mongos`` processes
+- ``ServerMonitorListener``: heartbeat related events
+
+For a detailed description of all nine cluster events in the driver, see the 
+:spec:`MongoDB SDAM monitoring specification </server-discovery-and-monitoring/server-discovery-and-monitoring-monitoring.rst#events>`.
+
```
kyuan-mongodb, line 174: Issue: Since you include a sentence about what the user needs to do to create that event in the other two event sections, I think you should include one here as well.

source/fundamentals/monitoring.txt
```diff
@@ -0,0 +1,406 @@
...
+.. code-block:: none
+   :copyable: false
+
+   Let me get you the connection with id 21...
+   Let me get you the connection with id 21...
+
+It looks like our find operation requested two connections from the pool.
+If you are curious why the driver uses two connections in one find operation, 
+`you can find the answer in this line of the MongoDB driver source code <https://github.com/mongodb/mongo-java-driver/blob/a82ce71d0aa7da9c6455f1b52ae24ed3cd4c1f62/driver-sync/src/main/com/mongodb/client/internal/FindIterableImpl.java#L203>`__.
```
kyuan-mongodb, line 266:
```suggestion 
Our find operation requested two connections from the pool.
For information on why the driver uses two connections in a find operation, see
`this line of the MongoDB driver source code <https://github.com/mongodb/mongo-java-driver/blob/a82ce71d0aa7da9c6455f1b52ae24ed3cd4c1f62/driver-sync/src/main/com/mongodb/client/internal/FindIterableImpl.java#L203>`__.
```

source/fundamentals/monitoring.txt
```diff
@@ -0,0 +1,406 @@
...
+
+- :java-docs:`ConnectionPoolListener <apidocs/mongodb-driver-core/com/mongodb/event/ConnectionPoolListener.html>`
+- :java-docs:`MongoClientSettings <apidocs/mongodb-driver-core/com/mongodb/MongoClientSettings.html>`
+- :java-docs:`MongoClient <apidocs/mongodb-driver-sync/com/mongodb/client/MongoClient.html>`
+- :java-docs:`ConnectionCheckedOutEvent <apidocs/mongodb-driver-core/com/mongodb/event/ConnectionCheckedOutEvent.html>`
+- :java-docs:`ConnectionCheckOutFailedEvent <apidocs/mongodb-driver-core/com/mongodb/event/ConnectionCheckOutFailedEvent.html>`
+
+.. _monitoring-jmx:
+
+Monitor Connection Pools with JMX
```
kyuan-mongodb, line 279: Question: Should this heading say: "Monitor Connection Pool Events with JMX"?

source/fundamentals/monitoring.txt
```diff
@@ -0,0 +1,406 @@
...
+---------------------------------
+
+You can monitor connection pool events using **Java Management Extensions (JMX)**. 
+JMX provides tools to monitor applications and devices.
+
+For more information on JMX, see 
+`the official Oracle JMX documentation <https://docs.oracle.com/javase/tutorial/jmx/index.html>`__. 
+
+JMX Support in the MongoDB Driver
```
kyuan-mongodb, line 288: Suggestion: I think the MongoDB Driver is implied here.
```suggestion
JMX Support
```

source/fundamentals/monitoring.txt
```diff
@@ -0,0 +1,406 @@
...
+- :java-docs:`ConnectionCheckedOutEvent <apidocs/mongodb-driver-core/com/mongodb/event/ConnectionCheckedOutEvent.html>`
+- :java-docs:`ConnectionCheckOutFailedEvent <apidocs/mongodb-driver-core/com/mongodb/event/ConnectionCheckOutFailedEvent.html>`
+
+.. _monitoring-jmx:
+
+Monitor Connection Pools with JMX
+---------------------------------
+
+You can monitor connection pool events using **Java Management Extensions (JMX)**. 
+JMX provides tools to monitor applications and devices.
```
kyuan-mongodb, line 283: Issue: This sentence leaves me wondering what tools? And in what type of situation is it useful?

source/fundamentals/monitoring.txt
```diff
@@ -0,0 +1,406 @@
...
+
+- Creates MXBean instances for each ``mongod`` or ``mongos`` process the driver
+  maintains a connection pool with.
+- Registers these MXBean instances with the platform MBean server. 
+
+MXBeans registered on the platform MBean server have the following properties:
+
+.. list-table::
+   :header-rows: 1
+   :stub-columns: 1
```
kyuan-mongodb, line 304: Issue: the columns have unsatisfactory widths

source/fundamentals/monitoring.txt
```diff
@@ -0,0 +1,406 @@
...
+
+   * - Property
+     - Description
```
kyuan-mongodb, line 307: Question: Do any of the properties have default values? If so, I think it's worth listing as the last line in the description column,

source/fundamentals/monitoring.txt
```diff
@@ -0,0 +1,406 @@
...
+
+JMX Support in the MongoDB Driver
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+To enable JMX connection pool monitoring, add an instance of the
+``JMXConnectionPoolListener`` class to your ``MongoClient``.
+
+The ``JMXConnectionPoolListener`` class performs the following actions: 
+
+- Creates MXBean instances for each ``mongod`` or ``mongos`` process the driver
```
kyuan-mongodb, line 296: Question: Should `MXBean` and `MBean` be monospaced? 
Applies to all instances.

source/fundamentals/monitoring.txt
```diff
@@ -0,0 +1,406 @@
...
+  maintains a connection pool with.
+- Registers these MXBean instances with the platform MBean server.
```
kyuan-mongodb, line 298:
```suggestion
- Registers the MXBean instances with the platform MBean server. 
```

source/fundamentals/monitoring.txt
```diff
@@ -0,0 +1,406 @@
...
+To enable JMX connection pool monitoring, add an instance of the
+``JMXConnectionPoolListener`` class to your ``MongoClient``.
+
+The ``JMXConnectionPoolListener`` class performs the following actions: 
+
+- Creates MXBean instances for each ``mongod`` or ``mongos`` process the driver
+  maintains a connection pool with.
+- Registers these MXBean instances with the platform MBean server. 
+
+MXBeans registered on the platform MBean server have the following properties:
```
kyuan-mongodb, line 300: Issue: Before showing the properties of MXBean, I think it would be helpful to briefly explain what it is/does.

source/fundamentals/monitoring.txt
```diff
@@ -0,0 +1,406 @@
...
+
+.. code-block:: shell
+   
+   jconsole
+
+.. note:: Consult the Official JConsole Documentation
+
+   The descriptions of the JConsole GUI on this page are illustrative
+   rather than a source of truth. For guaranteed up to date information, consult 
+   `the official JConsole documentation <https://www.oracle.com/technical-resources/articles/java/jconsole.html>`__.
```
kyuan-mongodb, line 379: Issue: Let's show confidence in our documentation. Instead, you can omit this or put this after your steps (at line 395) to say:

```suggestion
For more information on starting up and using JConsole, visit `the official JConsole documentation <https://www.oracle.com/technical-resources/articles/java/jconsole.html>`__. 
```

source/fundamentals/monitoring.txt
```diff
@@ -0,0 +1,406 @@
...
+
+- `Oracle Platform MBean Server Reference documentation <https://docs.oracle.com/en/java/javase/16/management/overview-java-se-monitoring-and-management.html#GUID-F7B9AB8A-F5A8-472A-AEC6-93B5B7FBE7CE>`__
+- `Oracle MXBean documentation <https://docs.oracle.com/javase/tutorial/jmx/mbeans/mxbeans.html>`__
+
+JMX and JConsole Example
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+This example shows how you can monitor the driver's connection pools using JMX
+and **JConsole**. JConsole is a JMX compliant GUI monitoring tool that comes with
+the Java Platform.
```
kyuan-mongodb, line 347: Question: Is there a reason to include JConsole here? If so, I think there needs to be more information around why it's useful and when to use it before jumping into an example.

kyuan-mongodb (changes requested):
Great start so far! Most comments revolve around clarification and consistency.

source/fundamentals/monitoring.txt
```diff
@@ -0,0 +1,406 @@
...
+
+.. _monitoring-monitor-events:
+
+Monitor Events
+--------------
+
+To monitor an **event**, you must register a **listener** on your ``MongoClient``
+instance. 
+
+An event is any action that happens in code. The driver includes functionality
```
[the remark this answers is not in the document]
biniona-mongodb: 1: By action I mean anything that happens in a program (variable is referenced, network request is made, resource is allocated). An event is anything that happens, so "action" is very broad. 

2: I am referring to the code and not to the database.

source/fundamentals/monitoring.txt
```diff
@@ -0,0 +1,406 @@
...
+Command Events
+~~~~~~~~~~~~~~
+
+A command event is an event related to a MongoDB database command. Here are some
+examples of database commands that produce command events:
+
+- ``find``
+- ``insert``
+- ``delete``
+- ``count``
```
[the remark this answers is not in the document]
biniona-mongodb: I think it is correct as written. See the [commands reference page](https://docs.mongodb.com/manual/reference/command/) (commands are written in same format as here). 

From the Glossary:

[Database Command](https://docs.mongodb.com/manual/reference/glossary/#std-term-database-command): A MongoDB operation, other than an insert, update, remove, or query. For a list of database commands, see Database Commands. To use database commands, see Issue Commands.

[Operator](https://docs.mongodb.com/manual/reference/glossary/#std-term-operator): A keyword beginning with a $ used to express an update, complex query, or data transformation. For example, $gt is the query language's "greater than" operator. For available operators, see Operators.

Based off of these definitions, operators (keywords used in update or query specifications) receive a `$`, and methods receive `()`.

source/fundamentals/monitoring.txt
biniona-mongodb: I agree that "our find operation" is incorrect. I disagree with the other two edits. This is meant to be a fun aside for readers exploring a curious behavior of the driver, not necessary for understanding the feature being described. For that reason, this was written with a different tone, "it looks like... if you are curious".

```suggestion
It looks like the above find operation requested two connections from the pool.
If you are curious why the driver uses two connections in one find operation, 
`you can find the answer in this line of the MongoDB driver source code <https://github.com/mongodb/mongo-java-driver/blob/a82ce71d0aa7da9c6455f1b52ae24ed3cd4c1f62/driver-sync/src/main/com/mongodb/client/internal/FindIterableImpl.java#L203>`__.
```

source/fundamentals/monitoring.txt
biniona-mongodb: As JConsole is a third party tool that may change without us knowing, I believe it is appropriate to provide a disclaimer that readers should expect discrepancies in our guide and their experience.

In a previous review documenting 3rd party technologies, [a disclaimer similar to this was requested by Jeff Yemin](https://github.com/mongodb/docs-java/pull/76#discussion_r623846300)

source/fundamentals/monitoring.txt
biniona-mongodb: I do not believe they should. These are referring to a software design concept rather than a class. I believe they should receive the same formatting as POJO (no monospace).

source/fundamentals/monitoring.txt
biniona-mongodb: Thanks for the note. I agree that this is unnecessary. Thank you for pointing this out.

source/fundamentals/monitoring.txt
```diff
@@ -0,0 +1,406 @@
...
+--------
+
+.. What is on this page?
+
+This guide explains how you can set up and configure **monitoring** in the
+MongoDB Java driver.
+
+.. What do any new terms mean?
+
+Monitoring is the process of getting information about the activities a piece of
```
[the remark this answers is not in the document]
biniona-mongodb: I agree that "piece of software" is too wordy. I am changing it to "program".

source/fundamentals/monitoring.txt
```diff
@@ -0,0 +1,406 @@
...
+- :java-docs:`CommandStartedEvent <apidocs/mongodb-driver-core/com/mongodb/event/CommandStartedEvent.html>`
+- :java-docs:`CommandSucceededEvent <apidocs/mongodb-driver-core/com/mongodb/event/CommandSucceededEvent.html>`
+- :java-docs:`CommandFailedEvent <apidocs/mongodb-driver-core/com/mongodb/event/CommandFailedEvent.html>`
+
+Cluster Events
+~~~~~~~~~~~~~~
+
+A cluster event in the driver corresponds to a change in the state of the
+MongoDB instance or cluster you have connected the driver to. Cluster events are
+also called server discovery and monitoring (SDAM) events.
```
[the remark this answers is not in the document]
biniona-mongodb: I disagree with this edit. I don't believe it would benefit readers to see the names as equivalent (cluster name is more useful in the Java driver, so I believe it should be given preference). The second name is introduced later only because the it is the term used in the specification, and I don't want readers to be confused when they are linked to the specification later in the paragraph.

source/fundamentals/monitoring.txt
```diff
@@ -0,0 +1,406 @@
...
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+Overview
+--------
+
+.. What is on this page?
+
+This guide explains how you can set up and configure **monitoring** in the
```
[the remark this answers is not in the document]
biniona-mongodb: I believe this intro style is also used in our documentation. I prefer the intro as written, so I will not implement the suggested edit.

See the following pages for examples of this intro style:

[POJO](https://docs-mongodbcom-staging.corp.mongodb.com/e1acf1b/java/docsworker-xlarge/DOCSP-9573-Monitoring/fundamentals/data-formats/document-data-format-pojo/)
[BSON](https://docs-mongodbcom-staging.corp.mongodb.com/e1acf1b/java/docsworker-xlarge/DOCSP-9573-Monitoring/fundamentals/data-formats/document-data-format-bson/)

source/includes/fundamentals/code-snippets/Monitoring.java
```diff
@@ -0,0 +1,169 @@
...
+        // start monitor-command-example
+        MongoClientSettings settings =
+                MongoClientSettings.builder()
+                        .applyConnectionString(URI)
+                        .addCommandListener(new CommandTimer())
+                        .build();
+        MongoClient mongoClient = MongoClients.create(settings);
+        // Run some commands to test the timer
+        MongoDatabase database = mongoClient.getDatabase(DATABASE);
+        MongoCollection<Document> collection = database.getCollection(COLLECTION);
```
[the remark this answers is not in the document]
biniona-mongodb: As the crux of this example involves instantiating a MongoClient instance in a specific way, I believe it makes sense to include these lines of code. If we did not include it, the example would not be runnable. If we broke this up into 2 code examples, we would need a comment illustrating that we instantiated a MongoCollection object and a prose sentence connecting the two examples. At present, this seems like the most compact readable way to represent this information.

source/fundamentals/monitoring.txt
biniona-mongodb: Thanks for the note. I believe it should, good catch.

source/fundamentals/monitoring.txt
biniona-mongodb: Thanks for the note, I agree these columns are too narrow. Fix: https://github.com/mongodb/docs-java/pull/88/commits/fa78aebb61993631209080b0147246e5cc78d347

source/fundamentals/monitoring.txt
biniona-mongodb: I believe these are set through getter/setter methods on the connection pool. They would be default connection pool values, which I do not believe should be documented here.

source/fundamentals/monitoring.txt
biniona-mongodb: I am not sure how this edit improves the piece. This seems to duplicate "interface" rather than "the driver" in breaking this across multiple sentences. 

Am I missing the crux of this edit in some way?

source/fundamentals/monitoring.txt
biniona-mongodb: Thanks for the note. I agree this transition is abrupt. Fix:

Going forward, I will work on smoothing my transitions. I believe this is becoming a theme in reviews (choppy sentences and abrupt transitions).

source/fundamentals/monitoring.txt
biniona-mongodb: I believe some discussion of JConsole is necessary on this page. JMX requires a JMX compliant monitoring tool, the easiest to use JMX compliant tool [(and the tool referenced by Oracle in it's JMX monitoring guide ](https://docs.oracle.com/en/java/javase/16/management/monitoring-and-management-using-jmx-technology.html#GUID-E17343B0-FED2-4050-88D7-E4793E704ED5)) is JConsole. For this reason, I believe JConsole must be mentioned.

source/fundamentals/monitoring.txt
biniona-mongodb: I feel it is appropriate to provide an unsatisfactory definition of JMX here as it is not the intention of this guide to explain JMX to our readers. The goal of this definition is to give readers enough context to consult Oracle documentation on JMX.

source/fundamentals/monitoring.txt
```diff
@@ -0,0 +1,406 @@
...
+software performs. You can use the information you get from monitoring in your
+application or an application performance management library. 
+
+Monitoring the MongoDB Java driver lets you understand the
+driver's resource usage and performance, and can help you make informed
+decisions when designing and debugging your application.    
+
+.. What can you expect to see on this page?
+
+On this page you will learn how to perform these tasks:
```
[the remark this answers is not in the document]
biniona-mongodb: I don't think the "guide"/"page" discrepancy is a problem. Our sidebar TOC refers to all of our guides/pages as "pages" (says "On This Page" for as far as I can tell all of our pages), but we often refer to them as guides in the intro. A "guide" is necessarily a "page" within our docs.

I do not agree with the second edit. I believe the proposed rephrasing is less task oriented than the current phrasing on the page (I believe readers are interested in "tasks" they wish to perform rather than "sections" they may read).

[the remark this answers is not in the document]
biniona-mongodb: This edit suggests it may be beneficial to discuss how we would like to refer to our pages. Is "guide" the only appropriate term? Are "page" and "guide" both acceptable?

source/fundamentals/monitoring.txt
```diff
@@ -0,0 +1,406 @@
...
+listener classes extend the ``EventListener`` interface. A listener's
+API defines the events it can respond to.
+
+Each method of a listener class represents a response to a certain event. Each
+method receives one argument: an object representing the event the method
+responds to. All event objects are a subclass of the ``EventObject`` class.  
+
+The MongoDB Java driver organizes the events it defines into three categories:
+
+- Command Events
```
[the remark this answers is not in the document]
biniona-mongodb: I agree that this could be helpful for some readers, but I believe that turning these items into links would[ not provide sufficient context](https://docs.mongodb.com/meta/style-guide/style/links-and-cross-references/#link-and-cross-reference-text).

> Begin a cross-reference sentence with the purpose or benefit of the cross-reference (such as more information or examples). Such context helps users decide whether to follow the reference.

I do not wish to rephrase this sentence to be about the page itself (I want this sentence to be focused on the driver rather than the page), so I would like to keep this as it is currently written (not linking).

source/fundamentals/monitoring.txt
biniona-mongodb: I believe "these" is correct here. 

Definition of [these](https://www.google.com/search?q=these&oq=these&aqs=chrome..69i57j69i60l5j69i65j69i64.502j0j7&sourceid=chrome&ie=UTF-8): "referring to a specific thing or situation just mentioned."

Is there a reason why "the" is more appropriate here? Am I missing the crux of this suggestion?

source/includes/fundamentals/code-snippets/Monitoring.java
biniona-mongodb: The write message comes from the `hasWritableServer()` method of the [`ClusterDescription`](https://mongodb.github.io/mongo-java-driver/4.2/apidocs/mongodb-driver-core/com/mongodb/connection/ClusterDescription.html#hasWritableServer()) object.

The [Cluster Description](https://mongodb.github.io/mongo-java-driver/4.2/apidocs/mongodb-driver-core/com/mongodb/connection/ClusterDescription.html#hasWritableServer()) has general information about the cluster the driver is connecting with. Is there something about this paragraph that is unclear or confusing?

source/fundamentals/monitoring.txt
```diff
@@ -0,0 +1,406 @@
...
+- :java-docs:`ServerListener <apidocs/mongodb-driver-core/com/mongodb/event/ServerListener.html>`
+- :java-docs:`ServerMonitorListener <apidocs/mongodb-driver-core/com/mongodb/event/ServerMonitorListener.html>`
+- :java-docs:`MongoClientSettings <apidocs/mongodb-driver-core/com/mongodb/MongoClientSettings.html>`
+- :java-docs:`MongoClient <apidocs/mongodb-driver-sync/com/mongodb/client/MongoClient.html>`
+- :java-docs:`ClusterDescriptionChangedEvent <apidocs/mongodb-driver-core/com/mongodb/event/ClusterDescriptionChangedEvent.html>`
+
+Connection Pool Events
+~~~~~~~~~~~~~~~~~~~~~~
+
+Connection pool events are events related to a **connection pool** held by the driver.
```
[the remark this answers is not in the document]
biniona-mongodb: Thanks for the note. I have updated all event subsections to have a parallel intro sentence structure. Fix: https://github.com/mongodb/docs-java/pull/88/commits/3e00609af4b22c26daa2affc1596ecdf416f3536

Going forward, I will continue to try and refine my ear for parallel structure between sections.

source/fundamentals/monitoring.txt
```diff
@@ -0,0 +1,406 @@
...
+- :java-docs:`MongoClientSettings <apidocs/mongodb-driver-core/com/mongodb/MongoClientSettings.html>`
+- :java-docs:`MongoClient <apidocs/mongodb-driver-sync/com/mongodb/client/MongoClient.html>`
+- :java-docs:`CommandStartedEvent <apidocs/mongodb-driver-core/com/mongodb/event/CommandStartedEvent.html>`
+- :java-docs:`CommandSucceededEvent <apidocs/mongodb-driver-core/com/mongodb/event/CommandSucceededEvent.html>`
+- :java-docs:`CommandFailedEvent <apidocs/mongodb-driver-core/com/mongodb/event/CommandFailedEvent.html>`
+
+Cluster Events
+~~~~~~~~~~~~~~
+
+A cluster event in the driver corresponds to a change in the state of the
```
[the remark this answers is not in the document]
biniona-mongodb: I believe this comment responds to this issue: https://github.com/mongodb/docs-java/pull/88#discussion_r643284081

source/fundamentals/monitoring.txt
biniona-mongodb: Thanks for the note. I agree the section could benefit from addition. Fix: https://github.com/mongodb/docs-java/pull/88/commits/cda95419a82eac050d8c18b114c719d79df64b43

Going forward, I will try and better sense parallel structure issues.

source/fundamentals/monitoring.txt
```diff
@@ -0,0 +1,411 @@
...
+The driver defines nine cluster events. The driver divides these nine events
+between three separate listener interfaces which each listen for three of the
+nine events. Here are the three interfaces and the events they listen for:
+
+- ``ClusterListener``: :spec:`topology </server-discovery-and-monitoring/server-discovery-and-monitoring.rst#topology>`
+  related events
+- ``ServerListener``: events related to ``mongod`` or ``mongos`` processes
+- ``ServerMonitorListener``: heartbeat related events
+
+To monitor one of the three types of cluster events, write a class that
```
kyuan-mongodb, line 172: Nit: I don't think mentioning "one of the three" is necessary since you just listed them
```suggestion
To monitor any of the cluster events, write a class that
```

source/fundamentals/monitoring.txt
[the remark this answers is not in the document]
kyuan-mongodb: Since it's only 3 items and none are being defined here, can it be in a sentence? This would save vertical space.

source/includes/fundamentals/code-snippets/Monitoring.java
```diff
@@ -0,0 +1,169 @@
...
+
+    private void monitorCommandEvent() {
+        // start monitor-command-example
+        MongoClientSettings settings =
+                MongoClientSettings.builder()
+                        .applyConnectionString(URI)
+                        .addCommandListener(new CommandTimer())
+                        .build();
+        MongoClient mongoClient = MongoClients.create(settings);
+        // Run some commands to test the timer
```
kyuan-mongodb, line 39: Suggestion: I noticed with your other snippets, you place the comment for the tests being run right when it starts. For that reason, I think Line 39 should go right above your first test case.

source/fundamentals/monitoring.txt
```diff
@@ -0,0 +1,411 @@
...
+implements one of the three interfaces above and register an instance of that
+class with your ``MongoClient`` instance.
+
+For a detailed description of all nine cluster events in the driver, see the
```
kyuan-mongodb, line 176: Nit: Didn't think you needed to mention nine here since you did above.
```suggestion
For a detailed description for each cluster event in the driver, see the 
```

kyuan-mongodb (changes requested):
Almost there! Just a few nits on simplifying/clarifying some wording.

source/fundamentals/monitoring.txt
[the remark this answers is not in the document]
biniona-mongodb: The style guide wants lists of three or more items broken out of paragraphs into lists.

[Section of Style Guide](https://docs.mongodb.com/meta/style-guide/writing/write-clear-concise-sentences-paragraphs/#create-short-paragraphs)

>  When listing three or more items, use a bullet list instead of embedding the items in a paragraph.

I also personally prefer this style, as I think it is easier on the eyes.

Based off of this section of the style guide, I think it is correct as written.

source/fundamentals/monitoring.txt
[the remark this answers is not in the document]
biniona-mongodb: I'll change "in code" to "in a running program". This seems I think this is similar to the edit made here https://github.com/mongodb/docs-java/pull/88#discussion_r643139809

source/fundamentals/monitoring.txt
biniona-mongodb: The reason this phrasing was chosen is that a Java class can implement multiple interfaces. The way a reader has to attach an event to their MongoClient it wouldn't make sense to do this, so I wanted to be explicit that you should only implement classes that monitor one type at a time.

source/fundamentals/monitoring.txt
```diff
@@ -0,0 +1,411 @@
...
+Command Events
+~~~~~~~~~~~~~~
+
+A command event is an event related to a MongoDB database command. Here are some
+examples of database commands that produce command events:
+
+- ``find``
+- ``insert``
+- ``delete``
+- ``count``
```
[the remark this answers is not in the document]
biniona-mongodb: I believe this comment addresses this suggestion: https://github.com/mongodb/docs-java/pull/88#discussion_r643971348

source/fundamentals/monitoring.txt
biniona-mongodb: Sounds good. 

Reasons for keeping it as is:

I don't think this should be numbered, as in code this all happens as one function call, but I think it should be "these" as it is referencing the items created in the previous bullet.

source/includes/fundamentals/code-snippets/Monitoring.java
biniona-mongodb: Thanks for the note. I agree this comment should be moved after creating the MongoCollection.

Fix: https://github.com/mongodb/docs-java/pull/88/commits/64b0850a3a4318719a9f6d9e6749baa281ab2153

source/fundamentals/monitoring.txt
```diff
@@ -0,0 +1,411 @@
...
+.. What is on this page?
+
+This guide explains how you can set up and configure **monitoring** in the
+MongoDB Java driver.
+
+.. What do any new terms mean?
+
+Monitoring is the process of getting information about the activities a
+program performs. You can use the information you get from monitoring in your
+application or an application performance management library.
```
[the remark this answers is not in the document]
biniona-mongodb: I don't think this is an incomplete sentence.

Equivalent sentence: "You can use a screwdriver in your house."

- Subject: you
- Verb: use

I don't think this paragraph should touch on how monitoring information can be used (monitoring is mainly about getting information from running code to use in your code). 

I think describing how monitoring you can use monitoring in the driver is appropriate.

source/fundamentals/monitoring.txt
biniona-mongodb: I agree "of all nine" can be removed here.

```suggestion
For a detailed description of each cluster event in the driver, see the 
```

source/fundamentals/monitoring.txt
biniona-mongodb: I agree I think this makes sense. I have expanded the disclaimer to also include JMX and moved it to the top of the example.

Fix:

source/fundamentals/monitoring.txt
biniona-mongodb: I think these are different cases.

Kafka is documenting a feature which can be configured.

This is documenting a feature which exposes values to a monitoring framework. This feature in and of itself does not configure connection pools, it only passes its values. I agree that a discussion of connection pools should document default connection pool values, but I think this (a discussion of a feature exposing connection pool values to JMX) should not discuss default connection pool values as it is not immediately tied to the feature.

source/fundamentals/monitoring.txt
```diff
@@ -0,0 +1,409 @@
...
+
+.. What is on this page?
+
+This guide explains how you can set up and configure **monitoring** in the
+MongoDB Java driver.
+
+.. What do any new terms mean?
+
+Monitoring is the process of getting information about the activities a running 
+program performs for use in an application or an application performance
```
kyuan-mongodb, line 24:
```suggestion
program performs to use in an application or an application performance
```

source/fundamentals/monitoring.txt
```diff
@@ -0,0 +1,409 @@
...
+that every time the driver executes a command on your MongoDB instance
+you see a message letting you know how long it took to execute the command.
+
+To make a timer, do the following:
+
+#. Make a class with timer functionality that implements the ``CommandListener`` interface.
+#. Add an instance of the new class that implements ``CommandListener`` to a ``MongoClientSettings`` object.
+#. Configure a ``MongoClient`` instance with the ``MongoClientSettings`` object.
+
+The following code defines the ``CommandTimer`` class which implements the
```
kyuan-mongodb, line 113:
```suggestion
The following code defines the ``CommandTimer`` class, which implements the
```
Also applies to the other 2 event examples below.

source/fundamentals/monitoring.txt
```diff
@@ -0,0 +1,409 @@
...
+
+#. Select the Java process running the above example code.
+#. Press :guilabel:`Insecure Connection` in the warning dialog box.
+#. Click on the :guilabel:`MBeans` tab.
+#. Inspect your connection pool events under the ``"org.mongodb.driver"`` domain.
+
+When you no longer want to inspect your connection pools in JConsole, do the
+following: 
+
+- Exit JConsole by closing the JConsole window
```
kyuan-mongodb, line 394: Nit: Since you number the other sections for steps, you should here too.

kyuan-mongodb (approved):
LGTM with two suggestions and a nit!

source/fundamentals/monitoring.txt
biniona-mongodb: As these actions are not ordered (they can be performed in any order), I do not believe these should be numbered.

source/fundamentals/monitoring.txt
biniona-mongodb: I believe the suggested edit should not be performed. I think "which implements the ____ interface" is a restrictive clause. I believe you only [add a comma before which for nonrestrictive clauses. ](https://www.grammarly.com/blog/comma-before-which/?&utm_source=google&utm_medium=cpc&utm_campaign=10273012991&utm_targetid=dsa-1233402314764&gclid=Cj0KCQjw2NyFBhDoARIsAMtHtZ7H64TbNm032HTN7BNzgoBZlmVoB8hKBXbACGNcXt6xhoHi3iiAg2saAp43EALw_wcB&gclsrc=aw.ds)

source/fundamentals/monitoring.txt
biniona-mongodb: I made a sentence that I think is equivalent to test this edit:

"Dish washing is the process of washing dishes for use in meals."
"Dish washing is the process of washing dishes to use in meals."

I looked around online and I think the main difference is that  "use in meals" becomes a noun with "for" and "use" becomes a verb with "to". 

In my head "for" seems like a slightly better fit, but I think both are correct. Because I can't see a reason to choose either one, I will go with the one I prefer and stick with "for".

biniona-mongodb:
Hi @jyemin and @rozza, could I please have a review for this page?

source/fundamentals/logging.txt
```diff
@@ -33,7 +33,10 @@ You will learn how to:
 
 .. Who should read this? Where should I go if this isn't the page I was looking for?
 
-.. Not Applicable. No natural redirect location.
+This guide shows how to record events the driver performs.
```
jyemin, line 36: TODO: I think this needs to be re-worded.

source/fundamentals/monitoring.txt
```diff
@@ -0,0 +1,409 @@
...
+An event is any action that happens in a running program. The driver includes functionality
+for listening to a subset of the events that occur when the driver is running.
+
+A listener is a class that performs some action when certain events occur. All
+listener classes extend the ``EventListener`` interface. A listener's
+API defines the events it can respond to.
+
+Each method of a listener class represents a response to a certain event. Each
+method receives one argument: an object representing the event the method
+responds to. All event objects are a subclass of the ``EventObject`` class.
```
jyemin, line 59: > All event objects are a subclass of the ``EventObject`` class

Not true in practice.  The driver's event objects are not.

source/fundamentals/monitoring.txt
```diff
@@ -0,0 +1,409 @@
...
+--------------
+
+To monitor an **event**, you must register a **listener** on your ``MongoClient``
+instance. 
+
+An event is any action that happens in a running program. The driver includes functionality
+for listening to a subset of the events that occur when the driver is running.
+
+A listener is a class that performs some action when certain events occur. All
+listener classes extend the ``EventListener`` interface. A listener's
```
jyemin, line 54: > All listener classes extend the ``EventListener`` interface

Not true in practice. The driver's event listener interfaces are not.

source/fundamentals/monitoring.txt
```diff
@@ -0,0 +1,409 @@
...
+To monitor command events, write a class that implements the
+``CommandListener`` interface and register an instance of that class with your
+``MongoClient`` instance.
+
+For more information on MongoDB database commands, see the
+:manual:`MongoDB manual entry on database commands </reference/command/>`.
+
+.. note:: Redacted Output
+
+   As a security measure, the driver is unable to monitor some commands. This
```
jyemin, line 94: > unable

Consider a different word.  The driver could monitor all commands.  We just chose not to.  Also, it's not that we don't monitor them, we just redact the command documents.  The events are still generated.

source/fundamentals/monitoring.txt
```diff
@@ -0,0 +1,409 @@
...
+For a full list of the events you can monitor,
+:java-docs:`see the event package of the MongoDB Java Driver <apidocs/mongodb-driver-core/com/mongodb/event/package-summary.html>`.
+
+For more information on the classes and interfaces mentioned in this section,
+see the following API documentation:
+
+- `EventListener <https://docs.oracle.com/en/java/javase/16/docs/api/java.base/java/util/EventListener.html>`__
+- `EventObject <https://docs.oracle.com/en/java/javase/16/docs/api/java.base/java/util/EventObject.html>`__
+
+Command Events
```
jyemin, line 78: Add a note indicating that the driver does not publish events for commands called internally by the driver, e.g. for monitoring the cluster or for handshake on connection establishment, which includes `hello` and auth-related commands.

source/fundamentals/monitoring.txt
```diff
@@ -0,0 +1,409 @@
...
+   As a security measure, the driver is unable to monitor some commands. This
+   protects the sensitive information contained in the output of these
+   commands. For a full list of commands the driver can't monitor,
+   see the 
+   :spec:`MongoDB command monitoring specification </command-monitoring/command-monitoring.rst#security>`.
+
+Example
+^^^^^^^
+
+This example shows how to make a timer for database commands. The timer makes it so
```
jyemin, line 103: Unfortunately this not a good example, because both CommandSucceededEvent and CommandFailedEvent include an elapsedTime property!

jyemin: Also, if this is all you want to do, better to enable DEBUG logging, which already logs elapsed time.

source/includes/fundamentals/code-snippets/Monitoring.java
```diff
@@ -0,0 +1,169 @@
...
+            }
+        } else {
+            if (!event.getNewDescription().hasWritableServer()) {
+                isWritable = false;
+                System.out.println("Unable to write to server");
+            }
+        }
+
+        if (!isReadable) {
+            if (event.getNewDescription().hasReadableServer(ReadPreference.primary())) {
```
jyemin, line 140: Perhaps more interesting with `ReadPreference.secondary()`.  Otherwise, it's the same as `isWritable`.

source/fundamentals/monitoring.txt
```diff
@@ -0,0 +1,409 @@
...
+
+The above code snippet should produce output that looks like this:
+
+.. code-block:: none
+   :copyable: false
+
+   Let me get you the connection with id 21...
+   Let me get you the connection with id 21...
+
+It looks like the above find operation requested two connections from the pool.
```
jyemin, line 263: That's not actually the reason.   Here's the stack trace for the second check-out:

```
connectionCheckedOut:159, ConnectionPoolLibrarian (com.mycompany.app)
get:123, DefaultConnectionPool (com.mongodb.internal.connection)
get:100, DefaultConnectionPool (com.mongodb.internal.connection)
getConnection:96, DefaultServer (com.mongodb.internal.connection)
endClosedSessions:156, ServerSessionPool (com.mongodb.internal.session)
close:103, ServerSessionPool (com.mongodb.internal.session)
close:110, MongoClientDelegate (com.mongodb.client.internal)
close:139, MongoClientImpl (com.mongodb.client.internal)
monitorConnectionPoolEvent:81, Monitoring (com.mycompany.app)
main:28, Monitoring (com.mycompany.app)
```

So it's happening because of a requirement that the driver informs the server about sessions that it has created and will no longer be using, since the client is being closed.

You can avoid this by not calling MongoClient.close()

source/includes/fundamentals/code-snippets/JMXMonitoring.java
```diff
@@ -0,0 +1,102 @@
...
+        // start jmx-example
+        JMXConnectionPoolListener connectionPoolListener = new JMXConnectionPoolListener();
+        MongoClientSettings settings =
+                MongoClientSettings.builder()
+                        .applyConnectionString(URI)
+                        .applyToConnectionPoolSettings(builder -> builder.addConnectionPoolListener(connectionPoolListener))
+                        .build();
+        MongoClient mongoClient = MongoClients.create(settings);
+        int port = 9999;
+        javax.management.MBeanServer server = ManagementFactory.getPlatformMBeanServer();
```
jyemin, line 70: I don't think you need any of this funky code.  The JVM does this for you.  

Just put a `Thread.sleep(Long.MAX_VALUE)` after creating MongoClient, so the program doesn't exit, and then run jconsole.  You should still be able to connect and see all the MBeans.

jyemin (changes requested):

source/fundamentals/monitoring.txt
biniona-mongodb: Thanks for catching this. Apologies for getting this incorrect. Here is the fix: https://github.com/mongodb/docs-java/pull/88/commits/c7ed0891dc61a14408a0085f0125838bc058ac9d

Going forward, I will try check these sorts of statements by consulting stack traces before including them in pages. Thanks for sharing this technique.

source/fundamentals/monitoring.txt
biniona-mongodb: [Facepalm](https://en.wikipedia.org/wiki/Facepalm)

Thanks for pointing this out. I agree that as there are much more straightforward ways to achieve this, this example should be changed. I have changed this example to be a command counter (keeps track of how many times each command has been successfully executed in a map and prints the map). Fix: https://github.com/mongodb/docs-java/pull/88/commits/c0fecb8671ddbe9318a92b1c759220198d4bfb04

source/fundamentals/monitoring.txt
biniona-mongodb: Thanks for the note. I agree "unable" is the incorrect word. I have rephrased this note to change language around "unable" and "can't monitor" to be more focused on redacting contents. Fix: https://github.com/mongodb/docs-java/pull/88/commits/4cfc007205c993ad583f9d72594b84cbde7f7d4f

source/fundamentals/monitoring.txt
biniona-mongodb: Thanks for the suggestion. Here is the note: https://github.com/mongodb/docs-java/pull/88/commits/7bd930288fa4d60fbd8e9520f5c414b5602e658e

Let me know if you would like this note to go into more detail. I tried digging into the driver code to get an understanding of the specific types of commands the driver would not produce events for, but had trouble coming up with anything useful to readers. Apologies for not being able to fill out this content on my own, and please let me know if this is insufficient.

source/includes/fundamentals/code-snippets/JMXMonitoring.java
biniona-mongodb: Thanks for the suggestion. Fix: https://github.com/mongodb/docs-java/pull/88/commits/20e629ece2705d227875b2895c8d40de52800e7d

This next paragraph is purely for my own edification (please feel free to ignore):

It seems like the platform Mbean server is automatically exposed by the JVM for local connections when the ``com.sun.management.jmxremote`` property is set. Does the driver set this property? I looked into this for a little bit and couldn't find an answer.

source/includes/fundamentals/code-snippets/Monitoring.java
biniona-mongodb: Thanks for the note. I agree this 'is readable' section of the class is redundant. I agree that this would be more interesting with `ReadPreference.secondary()` as was demonstrated in the previous Java docs, however by using `ReadPreference.secondary()` I fear that the example will be harder for readers to understand (not a straightforward explanation for what it is doing, 'writable primary and readable secondary' is less intuitive than 'can read' or 'can write'.

To reduce redundancy in the example, I have changed the example class to 'IsWritable'. This has the benefit of shortening the example code. 

Please let me know if you agree with this fix, or if you would prefer the 'ReadPreference.Secondary()` style example be included.

Fix: https://github.com/mongodb/docs-java/pull/88/commits/a21f0d21e7083ae2b4bbc0a4c8f526a20d2b8256

source/fundamentals/monitoring.txt
biniona-mongodb: Thanks for the note. Apologies, I didn't realize that this convention was not always followed in practice. I am removing content related to this from the page.

Purely for my own edification (feel free to ignore):

Why are the driver's events and listeners not subclasses of `EventListener` and `EventObject`? Just curious why this decision was made (no need to answer this if you are busy).

source/fundamentals/monitoring.txt
biniona-mongodb: Response is here: https://github.com/mongodb/docs-java/pull/88#discussion_r646715047

source/fundamentals/logging.txt
biniona-mongodb: Thanks for the note. I agree "record events the driver performs" is incorrect (I don't think you can perform an event). I rephrased this to "record events in the driver".

Fix: https://github.com/mongodb/docs-java/pull/88/commits/f74439b83566b5873b5358bb1dca34928cc13530

jyemin: Sorry, that TODO was for me to come back to in order to suggest new wording.  But I like what you came up with!

source/fundamentals/monitoring.txt
jyemin: I think those types go back to an early time, when listeners were used primarily in GUI applications.  They are just not useful in this context.

source/fundamentals/monitoring.txt
jyemin: This looks just fine.  Thanks.

source/fundamentals/monitoring.txt
jyemin: This is a better example.  Only suggestion is to make the `commandSucceeded` method` `synchronized`, since as is the listener will suffer from race conditions in a multi-threaded application.

source/includes/fundamentals/code-snippets/Monitoring.java
```diff
@@ -0,0 +1,151 @@
...
+class CommandCounter implements CommandListener {
+
+    private Map<String, Integer> commands = new HashMap<String, Integer>();
+
+    @Override
+    public void commandSucceeded(final CommandSucceededEvent event) {
+        String commandName = event.getCommandName();
+        int count = commands.containsKey(commandName) ? commands.get(commandName) : 0;
+        commands.put(commandName, count + 1);
+        System.out.println(commands.toString());
```
jyemin, line 102: Use `commands.toJson()` instead (this will change the output to JSON, so you'll have to change that as well.

source/includes/fundamentals/code-snippets/Monitoring.java
```diff
@@ -0,0 +1,151 @@
...
+
+}
+
+// start command-listener-impl
+class CommandCounter implements CommandListener {
+
+    private Map<String, Integer> commands = new HashMap<String, Integer>();
+
+    @Override
+    public void commandSucceeded(final CommandSucceededEvent event) {
```
jyemin, line 98: `public synchronized void`

source/includes/fundamentals/code-snippets/Monitoring.java
```diff
@@ -0,0 +1,151 @@
...
+                event.getCommandName(),
+                event.getRequestId()));
+    }
+}
+// end command-listener-impl
+
+// start cluster-listener-impl
+class IsWriteable implements ClusterListener {
+
+    private boolean isWritable;
```
jyemin, line 117: `private volatile boolean`.

Otherwise there is a race condition.

source/fundamentals/monitoring.txt
jyemin: It's ok.  You're learning!

jyemin (changes requested):
Looking good.  Just a few small changes left.

source/includes/fundamentals/code-snippets/Monitoring.java
biniona-mongodb: Thanks for the note. Didn't realize that Java `HashMap` is not thread safe (and was not considering this class being used in a multithreaded app). Fix: https://github.com/mongodb/docs-java/pull/88/commits/d45b8ba06cb1687ce13a742cc722cefa54a81f99

Going forward, I will try to think about structuring examples for multithreaded environments.

source/includes/fundamentals/code-snippets/Monitoring.java
biniona-mongodb: Thanks for the note. Fix: https://github.com/mongodb/docs-java/pull/88/commits/841ecea2470d2b560f92be1793d4e88080183164

Question for my own edification:

I am not an expert of usage of `volatile`, but does the usage of `isWritable` as an invariant in the conditional disqualify it from usage of `volatile` (and thus require `synchronized` in the method signature)? 

[Old Java Concurrency Volatile Discussion:](http://web.archive.org/web/20210221170926/https://www.ibm.com/developerworks/java/library/j-jtp06197/)

Conditions for usage of volatile:

>  Writes to the variable do not depend on its current value.
The variable does not participate in invariants with other variables.

In researching volatile, a theme I am running into is "don't use a volatile variable if it's previous state will determine something about its future state". In this specific code snippet, I am having trouble coming up with a scenario that would actually create problems (if `isWritable` is written to by another thread in between reading in conditional and writing new boolean value I don't think this will create problems in the code), but based off my current understanding of conventions/patterns around `volatile` usage I am not sure which is more appropriate in this scenario.

jyemin: Thanks for asking, as it forced me to look more deeply at this, and it turns out I am wrong about it.  I thought that the driver guaranteed there are no concurrent calls to this method, but that's not actually true: the driver calls this method without holding any exclusive lock.  So I think it is more appropriate to make the method synchronized instead of making the field volatile.

biniona-mongodb: Ok, sounds good. Thanks for looking into this and helping me better understand usage of `volatile` and `synchronized`. Fix: https://github.com/mongodb/docs-java/pull/88/commits/7643f1a15acc72947e41a62f6a4328ccb009ebf2

source/includes/fundamentals/code-snippets/Monitoring.java
jyemin: Sorry, I mis-read this as `System.out.println(command.toString()`

biniona-mongodb: Ok, sounds good, I'll revert this back to the style it was when the initial comment in this thread was posted:

 `System.out.println(commands.toString());`

jyemin: So you should revert this change.

biniona-mongodb: Reverting Changes: https://github.com/mongodb/docs-java/pull/88/commits/72d0bb1157dc8c7b3d16a5c287e6e132251c6a0e

jyemin (approved):
LGTM!

PR #100, opened 2021-06-11
DOCSP-16466: TLS warning and OCSP stapling

ccho-mongodb:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-16466

Note at the bottom of this section: https://docs-mongodbcom-staging.corp.mongodb.com/java/docsworker-xlarge/DOCSP-16466-ocsp-stapling/quick-start/#query-your-mongodb-cluster-from-your-application

https://docs-mongodbcom-staging.corp.mongodb.com/java/docsworker-xlarge/DOCSP-16466-ocsp-stapling/fundamentals/connection/tls/#client-driven-ocsp

### Self-Review Checklist

- [x] Target the [Jasmin persona](https://drive.google.com/file/d/14FbBOLCVxwSP6M9BK4Nz1Ir9tzxT8_02/view)
- [x] Target the [Lucas persona](https://drive.google.com/file/d/1J2vqJxo7ldv7OP_obA9Q-avf0o_ju4Lk/view)


source/fundamentals/connection/tls.txt
```diff
@@ -211,3 +211,82 @@ object to the builder in the ``applyToSslSettings()`` lambda:
...
+   :widths: 35 65
+
+   * - Property
+     - Value
+
+   * - ``com.sun.net.ssl.checkRevocation``
+     - Set this property to ``true`` to enable revocation checking.
+
+   * - ``ocsp.enable``
+     - Set this property to ``true`` to enable client-driven OCSP.
```
terakilobyte, line 256: Question: Is there any default here, and if there is, would it be worth mentioning? This question applies to the other table in OCSP Stapling as well.

source/fundamentals/connection/tls.txt
```diff
@@ -211,3 +211,82 @@ object to the builder in the ``applyToSslSettings()`` lambda:
...
+
+   If the OCSP responder is unavailable, the TLS support provided by the
+   JDK reports a "hard fail". This differs from the "soft fail" behavior of
+   the MongoDB Shell and certain drivers.
+
+OCSP Stapling
+~~~~~~~~~~~~~
+
+OCSP Stapling is a type of OCSP in which the server must obtain the signed
+certificate from the certificate authority (CA) and including it in a
```
terakilobyte, line 268:
```suggestion
certificate from the certificate authority (CA) and include it in a
```

source/fundamentals/connection/tls.txt
```diff
@@ -211,3 +211,82 @@ object to the builder in the ``applyToSslSettings()`` lambda:
...
+.. warning::
+
+   If the OCSP responder is unavailable, the TLS support provided by the
+   JDK reports a "hard fail". This differs from the "soft fail" behavior of
+   the MongoDB Shell and certain drivers.
+
+OCSP Stapling
+~~~~~~~~~~~~~
+
+OCSP Stapling is a type of OCSP in which the server must obtain the signed
```
terakilobyte, line 267:
```suggestion
OCSP stapling is a type of OCSP in which the server must obtain the signed
```
suggestion for consistency. The [RFC](https://datatracker.ietf.org/doc/html/rfc6961#section-1) and the  wiki article on it uses lowercase 's'

source/includes/quick-start/force-tls.rst
```diff
@@ -0,0 +1,27 @@
...
+
+      jdk.tls.client.protocols=TLSv1.2
+
+   If you run your application from the command line or your IDE, you can
+   specify this property as a JVM argument using the following syntax:
+
+   .. code-block:: properties
+
+      -Djdk.tls.client.protocols=TLSv1.2
+
```
terakilobyte, line 27: This is great. Should it also be added to the FAQ?

source/fundamentals/connection/tls.txt
```diff
@@ -211,3 +211,82 @@ object to the builder in the ``applyToSslSettings()`` lambda:
                 })
         .build();
    MongoClient client = MongoClients.create(settings);
+
+Online Certificate Status Protocol (OCSP)
+-----------------------------------------
+
+OCSP is a standard used to check whether X.509 certificates are valid or
+revoked during the TLS handshake. The driver supports the following variations
+of OCSP:
```
terakilobyte, line 220: suggestion: I wonder if the removal of the conjunction could simplify this?

```suggestion
OCSP is a standard used to check the validity of X.509 certificates during the TLS handshake. The driver supports the following variations
of OCSP:
```

terakilobyte (approved):
LGTM. A couple of nonblocking suggestions and questions.

source/fundamentals/connection/tls.txt
ccho-mongodb: No, there's no default value. The variable names are from the built-in J2SE handling of OCSP:
https://docs.oracle.com/javase/8/docs/technotes/guides/security/jsse/ocsp.html

source/fundamentals/connection/tls.txt
ccho-mongodb: Good catch!

source/includes/quick-start/force-tls.rst
ccho-mongodb: Created a ticket: https://jira.mongodb.org/browse/DOCSP-17059

source/fundamentals/connection/tls.txt
ccho-mongodb: I think my usage of valid might not be specific enough. I was hoping that the juxtaposition of "valid" and "revoked" would signal opposites. OCSP as far as I know is only to check revocation status of a certificate with information that a revocation server can access. I don't know if revocation servers have any other responsibilities other than checking whether a certificate has been revoked prior to its expiry (whether the certificate is on the Certificate Revocation List), i.e. it doesn't explicitly validate anything about the certificate.
I found some information about the potential statuses that the revocation server can return, so I'll update.

ccho-mongodb:
Hi @jyemin and @rozza , in this PR, I added info on OCSP and OCSP stapling. Could one of you take a look when you get the chance?

jyemin (changes requested):
I just ran into this myself while trying to connection to a Serverless cluster with 11.0.5.  I think the bug is reported at https://bugs.openjdk.java.net/browse/JDK-8236039 and fixed in 11.0.7 patch release.  If that's the case can we instead recommend an upgrade to 11.0.7 or later?

ccho-mongodb:
> I just ran into this myself while trying to connection to a Serverless cluster with 11.0.5. I think the bug is reported at https://bugs.openjdk.java.net/browse/JDK-8236039 and fixed in 11.0.7 patch release. If that's the case can we instead recommend an upgrade to 11.0.7 or later?

Sure, that works. From the openjdk ticket, it looks like this was also an issue with some releases of 13, 14, and 15 so I will need to add those as well.

source/fundamentals/connection/tls.txt
```diff
@@ -211,3 +211,87 @@ object to the builder in the ``applyToSslSettings()`` lambda:
...
+     - Set this property to ``true`` to enable revocation checking.
+
+   * - ``ocsp.enable``
+     - Set this property to ``true`` to enable client-driven OCSP.
+
+.. warning::
+
+   If the OCSP responder is unavailable, the TLS support provided by the
+   JDK reports a "hard fail". This differs from the "soft fail" behavior of
+   the MongoDB Shell and certain drivers.
```
jyemin, line 267: certain => some other

source/fundamentals/connection/tls.txt
```diff
@@ -211,3 +211,87 @@ object to the builder in the ``applyToSslSettings()`` lambda:
...
+
+OCSP Stapling
+~~~~~~~~~~~~~
+
+OCSP stapling is a type of OCSP in which the server must obtain the signed
```
jyemin, line 272: Not sure about `type of OCSP` phrasing.  Is it really a type, or just a mechanism?

jyemin (approved):
A couple of phrasing suggestions, which you can feel free to ignore.

LGTM

source/fundamentals/connection/tls.txt
ccho-mongodb: I meant "type" as in "variation". I guess it's probably ok to assume the reader can draw the connection with OCSP and describe it instead as "a mechanism in which..."

PR #139, opened 2021-07-19
Docsp 17220 client specific key trust store

biniona-mongodb:
## Pull Request Info
Client specific trust and key stores

### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-17220

https://docs-mongodbcom-staging.corp.mongodb.com/java/docsworker-xlarge/DOCSP-17220-Client-Specific-Key-Trust-Store/fundamentals/connection/tls/#configure-a-client-specific-trust-store-and-key-store

### Self-Review Checklist

### DIFF
--- a/source/fundamentals/connection/tls.txt
+++ b/source/fundamentals/connection/tls.txt
@@ -172,8 +172,8 @@ store:
       sslContext.init(<Your Key Managers>, <Your Trust Managers>, <A Secure Random Number Generator>);
       MongoClientSettings settings = MongoClientSettings.builder()
                .applyToSslSettings(builder ->
-                     builder.enabled(true).
-                              context(sslContext)
+                     builder.enabled(true)
+                              .context(sslContext)
                               .build())
                .build();
       MongoClient client = MongoClients.create(settings);
--- a/source/fundamentals/connection/tls.txt
+++ b/source/fundamentals/connection/tls.txt
@@ -155,13 +155,13 @@ please refer to the `JSSE Reference Guide
 
 .. _tls-disable-hostname-verification:
 
-Configure a Client Specific Trust Store and Key Store
+Configure a Client-Specific Trust Store and Key Store
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-You can configure a client specific trust store and key store using the
+You can configure a client-specific trust store and key store using the
 ``init()`` method of the ``SSLContext`` class.
 
-The following code snippet configures a client specific trust store and key
+The following code snippet configures a client-specific trust store and key
 store:
 
 .. code-block:: java
--- a/source/fundamentals/connection/tls.txt
+++ b/source/fundamentals/connection/tls.txt
@@ -182,8 +182,8 @@ store:
       System.err.println(algorithm + " is not a valid message digest algorithm");
    }
 
-For more information on the ``SSLContext`` class, see the official Oracle API
-documentation for `SSL Context <https://docs.oracle.com/en/java/javase/16/docs/api/java.base/javax/net/ssl/SSLContext.html>`__. 
+For more information on the ``SSLContext`` class, see the API
+documentation for `SSL Context <https://cr.openjdk.java.net/~iris/se/16/spec/fr/java-se-16-fr-spec/api/java.base/javax/net/ssl/SSLContext.html>`__. 
 
 Disable Hostname Verification
 -----------------------------
--- a/source/fundamentals/connection/tls.txt
+++ b/source/fundamentals/connection/tls.txt
@@ -171,15 +171,15 @@ store:
       final SSLContext sslContext = SSLContext.getInstance(algorithm);
       sslContext.init(<Your Key Managers>, <Your Trust Managers>, <A Secure Random Number Generator>);
       MongoClientSettings settings = MongoClientSettings.builder()
-               .applyToSslSettings(builder ->
-                     builder.enabled(true)
-                              .context(sslContext)
-                              .build())
+               .applyToSslSettings(builder -> {
+                     builder.enabled(true);
+                     builder.context(sslContext);
+               })
                .build();
       MongoClient client = MongoClients.create(settings);
    }
    catch (NoSuchAlgorithmException e) {
-      System.err.println("I'm sorry, but " + algorithm + " is not a valid message digest algorithm");
+      System.err.println(algorithm + " is not a valid message digest algorithm");
    }
 
 For more information on the ``SSLContext`` class, see the official Oracle API
--- a/source/fundamentals/connection/tls.txt
+++ b/source/fundamentals/connection/tls.txt
@@ -78,7 +78,7 @@ interacts. You can configure access to these certificates in your application in
 two ways:
 
 - Through the JVM Trust Store and JVM Key Store.
-- Through a client-specific Trust Store or Keystore
+- Through a client-specific Trust Store and Key Store
 
 .. _tls-configure-jvm-truststore:
 
@@ -158,7 +158,31 @@ please refer to the `JSSE Reference Guide
 Configure a Client Specific Trust Store and Key Store
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+You can also configure a trust store and key store specific to a client using an
+instance of the ``SSLContext`` class.
 
+The following code snippet configures a client specific trust store and key
+store:
+
+.. code-block:: java
+
+   try {
+     final SSLContext sslContext = SSLContext.getInstance("TLS");
+     sslContext.init(<Your Key Managers>, <Your Trust Managers>, <A Secure Random Nummber Generator>);
+     MongoClientSettings settings = MongoClientSettings.builder()
+           .applyToSslSettings(builder ->
+                    builder.enabled(true).
+                             context(sslContext)
+                             .build())
+           .build();
+     MongoClient client = MongoClients.create(settings);
+   }
+   catch (NoSuchAlgorithmException e) {
+     System.err.println("I'm sorry, but TLS is not a valid message digest algorithm");
+   }
+
+For more information on the ``SSLContext`` class, see the official Oracle API
+documentation for `SSL Context <https://docs.oracle.com/en/java/javase/16/docs/api/java.base/javax/net/ssl/SSLContext.html>`__. 
 
 Disable Hostname Verification
 -----------------------------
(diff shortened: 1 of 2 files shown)


source/fundamentals/connection/tls.txt
```diff
@@ -74,8 +74,11 @@ Configure Certificates
 Java applications that initiate TLS/SSL requests require access to
 cryptographic certificates that prove identity for the application
 itself as well as other applications with which the application
-interacts. To configure access to these certificates in your application,
-you should use the JVM Trust Store and the JVM Key Store.
+interacts. You can configure access to these certificates in your application with
+two mechanisms:
+
+- The JVM Trust Store and JVM Key Store
```
kyuan-mongodb, line 80: Question: Are these two co-dependent? If so, should each one have its own bullet?

source/fundamentals/connection/tls.txt
```diff
@@ -74,8 +74,11 @@ Configure Certificates
 Java applications that initiate TLS/SSL requests require access to
 cryptographic certificates that prove identity for the application
 itself as well as other applications with which the application
-interacts. To configure access to these certificates in your application,
-you should use the JVM Trust Store and the JVM Key Store.
+interacts. You can configure access to these certificates in your application with
+two mechanisms:
```
kyuan-mongodb, line 78: Nit: you specified an exact amount of mechanisms -- I think we've been told to not to do this as it can be hard to maintain down the line.

```suggestion
the following mechanisms:
```

source/fundamentals/connection/tls.txt
```diff
@@ -152,6 +155,36 @@ please refer to the `JSSE Reference Guide
 
 .. _tls-disable-hostname-verification:
 
+Configure a Client-Specific Trust Store and Key Store
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+You can configure a client-specific trust store and key store using the
+``init()`` method of the ``SSLContext`` class.
```
kyuan-mongodb, line 162: Question: The old documentation also mentions that you also "set the JVM default ones", should that message be conveyed here as well?

source/fundamentals/connection/tls.txt
```diff
@@ -152,6 +155,36 @@ please refer to the `JSSE Reference Guide
...
+   String algorithm = "TLS";
+   try {
+      final SSLContext sslContext = SSLContext.getInstance(algorithm);
+      sslContext.init(<Your Key Managers>, <Your Trust Managers>, <A Secure Random Number Generator>);
+      MongoClientSettings settings = MongoClientSettings.builder()
+               .applyToSslSettings(builder ->
+                     builder.enabled(true)
+                              .context(sslContext)
+                              .build())
+               .build();
```
kyuan-mongodb, line 178: Question: Did you mean to have `.build()` twice?

source/fundamentals/connection/tls.txt
```diff
@@ -152,6 +155,36 @@ please refer to the `JSSE Reference Guide
...
+      MongoClient client = MongoClients.create(settings);
+   }
+   catch (NoSuchAlgorithmException e) {
+      System.err.println("I'm sorry, but " + algorithm + " is not a valid message digest algorithm");
```
kyuan-mongodb, line 182: Nit: I thought it was a bit awkward seeing code have personality.

```suggestion
      System.err.println(algorithm + " is not a valid message digest algorithm");
```

kyuan-mongodb (changes requested):
Looks good overall, I have some questions and nits. Also, I think you should link to each section since you're bulleting them out.

source/fundamentals/connection/tls.txt
biniona-mongodb: Great catch!

source/fundamentals/connection/tls.txt
biniona-mongodb: Great question. That was intentional, but I think it might be incorrect now that I am [double checking the API documentation](https://mongodb.github.io/mongo-java-driver/4.2/apidocs/mongodb-driver-core/com/mongodb/MongoClientSettings.Builder.html#applyToSslSettings(com.mongodb.Block)). Fix:

source/fundamentals/connection/tls.txt
biniona-mongodb: I agree with this edit. This is awkward. Fix: https://github.com/mongodb/docs-java/pull/139/commits/b9860821c896c9999ff47eecf7ab8389175f8070

Was looking at examples of how people like to handle this error, and didn't look at the sample I was drawing from critically enough.

source/fundamentals/connection/tls.txt
biniona-mongodb: Interestingly, it seems like the answer is [yes and no](https://stackoverflow.com/questions/54897842/java-sslcontext-init-with-null-key-manager) (according to this Stack Overflow post which may or may not be good information), depends on your JDK. I think "are JVM Trust Store and JVM Keystore Co-dependent?" is worth asking in technical review, as TLS certificate management is something I don't know well, and I don't think it is worth looking into too much for this PR specifically (should definitely learn it well soon however).

I think this breakdown is alright (this mirrors the breakdown on the old java doc as well), where specifying Key/Trust with `SSLContext.init` is one section and JVM Key and JVM Trust are their own sections which are discussed together in an intro paragraph.

source/fundamentals/connection/tls.txt
biniona-mongodb: I think it is an either or situation. You either specify the JVM defaults or a client specific one. 

> One may either configure trust stores and key stores specific to the client via javax.net.ssl.SSLContext.init(KeyManager[] km, TrustManager[] tm, SecureRandom random), or set the JVM default ones.

kyuan-mongodb (approved):
LGTM!

biniona-mongodb:
Hi @jyemin, could I get a technical review of this content?

Also, is there a use case where a user would want to configure a Client Specific Trust Store but use a JVM Key Store or vice versa? Is this supported? Is this edge case worth discussing in the Docs? Apologies, I don't have much of a sense of how customers use this feature.

source/fundamentals/connection/tls.txt
```diff
@@ -152,6 +155,36 @@ please refer to the `JSSE Reference Guide
...
+                     builder.context(sslContext);
+               })
+               .build();
+      MongoClient client = MongoClients.create(settings);
+   }
+   catch (NoSuchAlgorithmException e) {
+      System.err.println(algorithm + " is not a valid message digest algorithm");
+   }
+
+For more information on the ``SSLContext`` class, see the official Oracle API
```
stIncMale, line 185: The above example contains lots of code that does not convey much helpful information. Following are the things we need to tell the user:

1. you can configure driver-specific key store and trust-store (was already done verbally before the code sample);
2. you can use `SSLContext.init` to do this (was already done verbally before the code sample; all we need is to specify the method name and the link to its documentation as it was done in the old driver docs: [javax.net.ssl.SSLContext.init(KeyManager[] km, TrustManager[] tm, SecureRandom random](https://docs.oracle.com/en/java/javase/16/docs/api/java.base/javax/net/ssl/SSLContext.html#init(javax.net.ssl.KeyManager%5B%5D,javax.net.ssl.TrustManager%5B%5D,java.security.SecureRandom)));
3. you can then tell the driver to use your custom `SSLContext` as described in ["Customize TLS/SSL Configuration with an SSLContext"](https://docs.mongodb.com/drivers/java/sync/current/fundamentals/connection/tls/#customize-tls-ssl-configuration-with-an-sslcontext).

I.e., the whole part can be documented in three sentences of which two are already in this PR. I propose removing the code sample and replacing it with a sentence as described in item 3.

stIncMale (changes requested):

source/fundamentals/connection/tls.txt
```diff
@@ -152,6 +155,36 @@ please refer to the `JSSE Reference Guide
...
+documentation for `SSL Context <https://docs.oracle.com/en/java/javase/16/docs/api/java.base/javax/net/ssl/SSLContext.html>`__.
```
stIncMale, line 186: > the official Oracle API documentation

`javax.net.ssl.SSLContext` is a part of the Java SE, and this specific URI simply points to a page that contains a part of the Java SE API documentation published by Oracle, i.e., it is not specific to Oracle JDK, which means there is no need to emphasize "Oracle API", as it is not Oracle-specific API.

source/fundamentals/connection/tls.txt
biniona-mongodb: Thank you for the note. Re-reading this section with your comments in mind, I can see how the code snippet does not provide much helpful information for readers. I have removed the code snippet, and linked to the "Customize TLS/SSL Configuration..." section as suggested. Here is the fix: https://github.com/mongodb/docs-java/pull/139/commits/c2803079ee4485e005c4627f6b3f3da68cfa05f9

source/fundamentals/connection/tls.txt
biniona-mongodb: Thank you for the note. I have updated the API docs to the non-vendor specific link provided above and removed the reference to Oracle. Here is the fix: https://github.com/mongodb/docs-java/pull/139/commits/c2803079ee4485e005c4627f6b3f3da68cfa05f9

Apologies for my poor understanding of the Java ecosystem. For my own edification, do you think it is a good practice to link to OpenJDK API docs over Oracle JDK API Docs for Java SE libraries? Sorry if this is a poorly worded or silly question.

stIncMale: No worries. Here are the ideas in a nutshell:

- if we are linking to a piece of documentation that is part of the Java SE, we should not imply that it is vendor-specific by saying "Oracle documentation" even if the linked Java SE documentation is published by Oracle or a different vendor;
- If we are linking to a piece of documentation that is not part of the Java SE and is vendor-specific, then we should point that out.

As for the [API docs published at docs.oracle.com](https://docs.oracle.com/en/java/javase/16/docs/api/index.html), one can differentiate between the Java SE API and Oracle JDK API by looking at the "Java SE" and "JDK"/"Other modules" tabs. Everything in modules listed under the "Java SE" tab is part of the Java SE, anything listed under "JDK"/"Other modules" is not.

> do you think it is a good practice in general to link to OpenJDK API docs over Oracle JDK API Docs for Java SE libraries

Since Java driver docs have been referring to docs published at docs.oracle.com, I think we should continue doing so. I provided the official Java SE API link (it is non-vendor-specific) just to support my words that "`javax.net.ssl.SSLContext` is a part of the Java SE`, not to suggest that we should start linking to a different place. Sorry if that caused confusion.

biniona-mongodb: Sorry for misinterpreting your comment, and thanks for all of the helpful background information! I will make sure to check what is part of the Java SE vs. a vendor specific release and use the correct language going forward. Fix: https://github.com/mongodb/docs-java/pull/139/commits/a6cbdf0922cfd5cd8e9664435988e653ac70ed0f

stIncMale (approved):

PR #162, opened 2021-09-22
DOCSP-11685 mongo client settings page

kyuan-mongodb:
## Pull Request Info

Added the MongoClientSettings page.

This page covers the different settings you can specify within your MongoClientSettings.

### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-11685

### Snooty build log:
**Paste your workerpool job link here**

https://docs-mongodbcom-staging.corp.mongodb.com/java/docsworker-xlarge/DOCSP-11685-MongoClientSettings/fundamentals/connection/mongoclientsettings/

### Self-Review Checklist

- [x] Are the staging and workerpool job links in the PR description updated?

### DIFF
--- a/source/fundamentals/connection/mongoclientsettings.txt
+++ b/source/fundamentals/connection/mongoclientsettings.txt
@@ -1,0 +1,181 @@
+====================
+MongoClient Settings
+====================
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 1
+   :class: singlecol
+
+
+Overview
+--------
+
+In this guide, you can learn how to use the different settings to
+control the behavior of your ``MongoClient``.
+
+You can control the behavior of your ``MongoClient`` by passing in a
+`MongoClientSettings <{+api+}/apidocs/mongodb-driver-core/com/mongodb/MongoClientSettings.html>`__
+object to the `MongoClients.create() <{+api+}/apidocs/mongodb-driver-sync/com/mongodb/client/MongoClients.html#create(com.mongodb.MongoClientSettings)>`__
+method.
+
+In your ``MongoClientSettings``, you can specify the following settings:
+
+- :ref:`Cluster Settings <mcs-cluster-settings>`
+- :ref:`Socket Settings <mcs-socket-settings>`
+- :ref:`Connection Pool Settings <mcs-connectionpool-settings>`
+- :ref:`Server Settings <mcs-server-settings>`
+- :ref:`SSL Settings <mcs-ssl-settings>`
+
+To specify any of the preceding settings, use the ``MongoClientSettings.builder()``
+method. This builder makes your ``MongoClientSettings`` immutable and to support
+a simpler construction through chaining. You must use their ``build()``
+method to create the setting to apply to ``MongoClient``.
+
+Example
+~~~~~~~
+
+The following example specifies the ``MongoClient`` to use the settings
+embedded in the ``ConnectionString``:
+
+.. literalinclude:: /includes/fundamentals/code-snippets/mcs.java
+   :start-after: begin connectionString
+   :end-before: end connectionString
+   :language: java
+
+See the `API Documentation <{+api+}/apidocs/mongodb-driver-core/com/mongodb/MongoClientSettings.Builder.html>`__
+for a full list of methods you can use to set your ``MongoClient`` settings.
+
+The following sections describe the builder and method to use with an
+example on how to construct your ``MongoClient`` for specific settings
+to apply.
+
+.. _mcs-cluster-settings:
+
+Cluster Settings
+----------------
+
+Pass a ``ClusterSettings.Builder`` object to the `applyToClusterSettings() <{+api+}/apidocs/mongodb-driver-core/com/mongodb/MongoClientSettings.Builder.html#applyToClusterSettings(com.mongodb.Block)>`__
+to specify settings on how to connect to a MongoDB cluster. 
+
+Example
+~~~~~~~
+
+The following example specifies the following settings for the cluster:
+
+- The selection to take at most five milliseconds
+- To operate in a ``LOAD_BALANCED`` mode
+- To use ``host1`` as the host on port ``27017``
+
+.. literalinclude:: /includes/fundamentals/code-snippets/mcs.java
+   :start-after: begin clusterSettings
+   :end-before: end clusterSettings
+   :language: java
+
+See the `API Documentation
+<{+api+}/apidocs/mongodb-driver-core/com/mongodb/connection/ClusterSettings.Builder.html>`__
+for a full list of methods you can use to set your cluster settings.
+
+.. _mcs-socket-settings:
+
+Socket Settings
+---------------
+
+Pass a ``SocketSettings.Builder`` object to the `applyToSocketSettings() <{+api+}/apidocs/mongodb-driver-core/com/mongodb/MongoClientSettings.Builder.html#applyToSocketSettings(com.mongodb.Block)>`__
+to specify settings on how to connect to a MongoDB server.
+
+Example
+~~~~~~~
+
+The following example specifies the following settings for the socket:
+
+- The time to connect to take at most 1000 milliseconds
+- The time to read from the socket to take at most 10 seconds
+
+.. literalinclude:: /includes/fundamentals/code-snippets/mcs.java
+   :start-after: begin SocketSettings
+   :end-before: end SocketSettings
+   :language: java
+
+See the `API Documentation
+<{+api+}/apidocs/mongodb-driver-core/com/mongodb/connection/SocketSettings.Builder.html>`__
+for a full list of methods you can use to set your socket settings.
+
+.. _mcs-connectionpool-settings:
+
+Connection Pool Settings
+------------------------
+
+Pass a ``ConnectionPoolSettings.Builder`` object to the `applyToConnectionPoolSettings() <{+api+}/apidocs/mongodb-driver-core/com/mongodb/MongoClientSettings.Builder.html#applyToConnectionPoolSettings(com.mongodb.Block)>`__
+to specify settings related to a pool of connections to a MongoDB server.
+
+Example
+~~~~~~~
+
+The following example specifies the following settings for the
+connection pool:
+
+- The time for a thread to wait for an available connection to take at most five seconds
+- There should be at least one connection
+- The time a pooled connection to live for to be at most 20 milliseconds
+
+.. literalinclude:: /includes/fundamentals/code-snippets/mcs.java
+   :start-after: begin ConnectionPoolSettings
+   :end-before: end ConnectionPoolSettings
+   :language: java
+
+See the `API Documentation
+<{+api+}/apidocs/mongodb-driver-core/com/mongodb/connection/ConnectionPoolSettings.Builder.html>`__
+for a full list of methods you can use to set your connection pool settings.
+
+.. _mcs-server-settings:
+
+Server Settings
+---------------
+
+Pass a ``ServerSettings.Builder`` object to the `applyToServerSettings() <{+api+}/apidocs/mongodb-driver-core/com/mongodb/MongoClientSettings.Builder.html#applyToServerSettings(com.mongodb.Block)>`__
+to specify settings on how to monitor each server.
+
+Example
+~~~~~~~
+
+The following example specifies the following settings for the
+server:
+
+- The heartbeat frequency to be at least 10 milliseconds
+- The cluster monitor to attempt reaching a server every three seconds
+
+.. literalinclude:: /includes/fundamentals/code-snippets/mcs.java
+   :start-after: begin ServerSettings
+   :end-before: end ServerSettings
+   :language: java
+
+See the `API Documentation
+<{+api+}/apidocs/mongodb-driver-core/com/mongodb/connection/ServerSettings.Builder.html>`__
+for a full list of methods you can use to set your server settings.
+
+.. _mcs-ssl-settings:
+
+SSL Settings
+------------
+
+Pass a ``SslSettings.Builder`` object to the `applyToSslSettings() <{+api+}/apidocs/mongodb-driver-core/com/mongodb/MongoClientSettings.Builder.html#applyToSslSettings(com.mongodb.Block)>`__
+to specify settings on how to connect to MongoDB though SSL.
+
+Example
+~~~~~~~
+
+The following example specifies the following settings for the SSL:
+
+- To enable SSL
+- To not allow invalid host names
+
+.. literalinclude:: /includes/fundamentals/code-snippets/mcs.java
+   :start-after: begin SslSettings
+   :end-before: end SslSettings
+   :language: java
+
+See the `API Documentation
+<{+api+}/apidocs/mongodb-driver-core/com/mongodb/connection/SslSettings.Builder.html>`__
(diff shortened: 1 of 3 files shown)


source/fundamentals/connection/mongoclientsettings.txt
```diff
@@ -0,0 +1,197 @@
...
+- :ref:`Server Settings <mcs-server-settings>`
+- :ref:`SSL Settings <mcs-ssl-settings>`
+
+.. _mcs-cluster-settings:
+
+Cluster Settings
+----------------
+
+Use the `applyToClusterSettings() <{+api+}/apidocs/mongodb-driver-core/com/mongodb/MongoClientSettings.Builder.html#applyToClusterSettings(com.mongodb.Block)>`__
+method to specify settings on how to connect to a MongoDB cluster.
```
terakilobyte, line 64: issue: These settings aren't solely responsible for modifying how a driver connects to a cluster.

source/fundamentals/connection/mongoclientsettings.txt
```diff
@@ -0,0 +1,197 @@
...
+method to specify settings on how to connect to a MongoDB cluster. 
+
+Example
+~~~~~~~
+
+This example specifies the following settings for connecting to a
+MongoDB cluster:
+
+- To use the settings from your ``ConnectionString``
+- To connect to ``MULTIPLE`` servers in your cluster
```
terakilobyte, line 73: question: Why would this be necessary? Is it required if I'm specifying a replicaset or srv uri in the connection string?

source/fundamentals/connection/mongoclientsettings.txt
```diff
@@ -0,0 +1,197 @@
...
+method to specify settings related to a pool of connections to a MongoDB server.
+
+Example
+~~~~~~~
+
+This example specifies the following settings for connecting to a
+MongoDB connection pool:
+
+- The thread to wait at most ``5 SECONDS`` for an available connection
+- To have at most one connection associated with the pool
```
terakilobyte, line 128: issue: This wouldn't be a pool, and I am strongly opposed to including this value. A user could copy this and then would experience very poor performance (if not an outright seized application).

Better would be to discuss the _implications_ of these options. Why do we support connection pooling? What are some drawbacks with higher, and lower, values?

source/fundamentals/connection/mongoclientsettings.txt
```diff
@@ -0,0 +1,197 @@
...
+- The pooled connection to live for at most ``100 MILLISECONDS``
```
terakilobyte, line 129: issue: What would the performance implications be of creating and destroying a connection 10 times a second be?

source/fundamentals/connection/mongoclientsettings.txt
```diff
@@ -0,0 +1,197 @@
...
+Use the `applyToServerSettings() <{+api+}/apidocs/mongodb-driver-core/com/mongodb/MongoClientSettings.Builder.html#applyToServerSettings(com.mongodb.Block)>`__
+method to specify settings on how to monitor each server.
+
+Example
+~~~~~~~
+
+This example specifies the following settings for connecting to a
+MongoDB server:
+
+- The heartbeat frequency to be at least ``10 MILLISECONDS``
```
terakilobyte, line 155: question: What would happen here if a primary is elected and its average ping is 12 ms?

terakilobyte (changes requested):
Overall I think this lacks empathy for the developer. Our goal here is to help explain what these options _actually_ do from the user's perspective. Transcribing API documentation isn't valuable and would require us to modify this page any time the API docs changed that was reflected here.

Keep in mind that the examples we show need to be plausible in real scenarios. Avoid specifying values that would likely cause negative effects as this detracts from our authority and may leave a bad taste in the reader's mouth when they wonder "Why would I ever set a value to *that*?"

source/fundamentals/connection/mongoclientsettings.txt
```diff
@@ -0,0 +1,201 @@
...
+
+To create a ``MongoClientSettings`` object, use the
+``MongoClientSettings.builder()`` method and chain methods to specify
+your desired settings. After chaining all the settings you want, use the
+``build()`` method to create the ``MongoClientSettings`` object.
+
+.. tip:: Builder Pattern
+
+   The builder pattern makes the settings immutable and supports 
+   a simpler construction through chaining.
```
terakilobyte, line 31: question: This feels out of place reading the page now. Would it be safe to remove it?

source/fundamentals/connection/mongoclientsettings.txt
```diff
@@ -0,0 +1,201 @@
...
+
+Example
+~~~~~~~
+
+This example demonstrates specifying a ``ConnectionString`` in a
+``MongoClientSettings``:
```
terakilobyte, line 37: suggestion: I think this can be simplified without impacting the meaning

```suggestion
This example demonstrates specifying a ``ConnectionString``:
```

source/fundamentals/connection/mongoclientsettings.txt
```diff
@@ -0,0 +1,201 @@
...
+method to modify the driver's behavior when interacting with your
+MongoDB cluster.
+
+Example
+~~~~~~~
+
+This example specifies the following driver behavior in a MongoDB cluster:
+
+- To connect directly to a server, regardless of the type of cluster it is a part of
+- The primary node election to take at most ``35 SECONDS``
```
terakilobyte, line 77: question: Since direct connection means "connect to this server whether it is the primary or not", does it make sense to also specify a selection timeout? I think it's safe to remove the timeout and maybe mention this is analogous to directConnection and link to that.

source/fundamentals/connection/mongoclientsettings.txt
```diff
@@ -0,0 +1,201 @@
...
+For a list of methods you can use to specify your socket settings, see
+the `API Documentation <{+api+}/apidocs/mongodb-driver-core/com/mongodb/connection/SocketSettings.Builder.html>`__.
+
+.. _mcs-connectionpool-settings:
+
+Connection Pool Settings
+------------------------
+
+Chain the `applyToConnectionPoolSettings() <{+api+}/apidocs/mongodb-driver-core/com/mongodb/MongoClientSettings.Builder.html#applyToConnectionPoolSettings(com.mongodb.Block)>`__
+method to modify the driver's behavior when interacting with a pool of
```
terakilobyte, line 122: suggestion: I think this can be simplified

```suggestion
method to modify the way the driver manages its' connection pool.
```

source/fundamentals/connection/mongoclientsettings.txt
```diff
@@ -0,0 +1,201 @@
...
+   :language: java
+   :emphasize-lines: 3-5
+   :dedent:
+
+For a list of methods you can use to specify your server settings, see
+the `API Documentation <{+api+}/apidocs/mongodb-driver-core/com/mongodb/connection/ServerSettings.Builder.html>`__.
+
+.. _mcs-ssl-settings:
+
+SSL Settings
```
terakilobyte, line 173: suggestion: Also state TLS

```suggestion
TLS/SSL Settings
```

source/fundamentals/connection/mongoclientsettings.txt
```diff
@@ -0,0 +1,201 @@
...
+------------
+
+Chain the `applyToSslSettings() <{+api+}/apidocs/mongodb-driver-core/com/mongodb/MongoClientSettings.Builder.html#applyToSslSettings(com.mongodb.Block)>`__
+method to modify the driver's behavior when using SSL to secure a
```
terakilobyte, line 177:
```suggestion
method to modify the driver's behavior when using TLS/SSL to secure a
```

terakilobyte: throughout

terakilobyte (commented):
A few suggestions were made and some questions. Almost there.

terakilobyte (approved):
lgtm

kyuan-mongodb:
Hi @jyemin and @rozza , This PR is ready for review.

source/fundamentals/connection/mongoclientsettings.txt
```diff
@@ -0,0 +1,198 @@
...
+connection between your application and MongoDB.
+
+Example
+~~~~~~~
+
+This example specifies the following driver behavior for connecting to
+MongoDB with TLS/SSL:
+
+- To ensure your Atlas connection string works
+- To not allow a mismatched ``hostname`` in the server and certificate
```
jyemin, line 184: This might mislead users into thinking that the default behavior is to allow this mismatch, whereas the default is actually to now allow it.  Especially since this is a security-sensitive area, we should be very clear.  On the other hand, I don't want the example to be one that weakens security.

Perhaps use SslContext configuration as the example?

jyemin, line 184: This might mislead users into thinking that the default behavior is to allow this mismatch, whereas the default is actually to now allow it.  Especially since this is a security-sensitive area, we should be very clear.  On the other hand, I don't want the example to be one that weakens security.

Perhaps use SSLContext configuration as the example?

source/fundamentals/connection/mongoclientsettings.txt
```diff
@@ -0,0 +1,198 @@
...
+.. literalinclude:: /includes/fundamentals/code-snippets/mcs.java
+   :start-after: begin SslSettings
+   :end-before: end SslSettings
+   :language: java
+   :emphasize-lines: 3-5
+   :dedent:
+
+.. tip::
+
+   If you connect to localhost, you must disable TLS/SSL because there is no certificate.
```
jyemin, line 195: This is also misleading, since the default is for TLS to be disabled.

source/fundamentals/connection/mongoclientsettings.txt
```diff
@@ -0,0 +1,198 @@
...
+.. literalinclude:: /includes/fundamentals/code-snippets/mcs.java
+   :start-after: begin ConnectionString
+   :end-before: end ConnectionString
+   :language: java
+   :emphasize-lines: 3
+   :dedent:
+
+.. tip::
+
+   If you omit the ``applyConnectionString()`` method, the driver
```
jyemin, line 42: We should specify somewhere that order matters: last one wins, i.e. that if you call applyConnectionString after the other apply method, the connection string will take precedence.  If you call it before, whatever you call after will take precedence.

jyemin (changes requested):
Looking good. Remind me: is the plan to stop here, or to document more of the settings in a subsequent patch?

source/fundamentals/connection/mongoclientsettings.txt
```diff
@@ -0,0 +1,367 @@
...
+   * - ``maintenanceInitialDelay()``
+     - Sets the time to wait before running the first maintenance job.
+
+   * - ``maxConnectionIdleTime()``
+     - Sets the amount of time a connection can be idle before it's closed.
+
+   * - ``maxConnectionLifeTime()``
+     - Sets the maximum time a pooled connection can live for.
+
+   * - ``maxSize()``
```
jyemin, line 245: Let's take this opportunity to clarify a point of confusion for many users by saying something like:

The maxSize setting applies to each server in the cluster to which the driver is connected.  So if the driver is connected to three mongos servers, there can be at most maxSize connections to _each_ mongos server.

source/fundamentals/connection/mongoclientsettings.txt
```diff
@@ -0,0 +1,367 @@
...
+     - Sets the maximum amount of connections associated with a connection pool.
+
+   * - ``maxWaitTime()``
+     - Sets the maximum time to wait for an available connection.
+
+   * - ``minSize()``
```
jyemin, line 251: Same as my comment for maxSize

source/fundamentals/connection/mongoclientsettings.txt
```diff
@@ -0,0 +1,367 @@
...
+     - Sets the minimum amount of connections associated with a connection pool.
+
+Example
+~~~~~~~
+
+This example specifies the following driver behavior in a pool of
+``Connection`` types:
+
+- The thread to wait at most ``3 MINUTES`` for an available connection
```
jyemin, line 260: 3 minutes is an eternity.  let's change to 10 seconds

source/fundamentals/connection/mongoclientsettings.txt
```diff
@@ -0,0 +1,367 @@
...
+     - Uses the settings from a ``ConnectionString`` object.
+
+   * - ``applySettings()``
+     - Uses the server settings specified in a ``ServerSettings`` object.
+
+   * - ``heartbeatFrequency()``
+     - Sets the interval for a cluster monitor to attempt reaching a server.
+
+   * - ``minHeartbeatFrequency()``
+     - Sets the interval for server monitoring checks.
```
jyemin, line 305: The description in the javadoc is more accurate, though not very informative either

jyemin: Perhaps "minimum interval"?

source/fundamentals/connection/mongoclientsettings.txt
```diff
@@ -0,0 +1,367 @@
...
+
+Example
+~~~~~~~
+
+This example specifies the following driver behavior in a MongoDB server:
+
+- To interval between server monitoring checks to be at least ``700 MILLISECONDS``
```
jyemin, line 312: "To interval" isn't right, but not sure what you meant to say

source/fundamentals/connection/mongoclientsettings.txt
```diff
@@ -0,0 +1,367 @@
...
+     - Description
+
+   * - ``applyConnectionString​()``
+     - Uses the settings from a ``ConnectionString`` object.
+
+   * - ``applySettings()``
+     - Uses the TLS/SSL settings specified in a ``SslSettings`` object.
+
+   * - ``context()``
+     -  Sets the secure socket protocol implementation to use when you enable TLS/SSL.
```
jyemin, line 348: It's not really the implementation, it's the context.  Just use what's in the Javadoc:

> Sets the SSLContext for use when TLS/SSL is enabled.

source/fundamentals/connection/mongoclientsettings.txt
```diff
@@ -0,0 +1,367 @@
...
+
+   * - ``enabled()``
+     - Whether to ensure your Atlas connection string works.
```
jyemin, line 351: Where'd you get this from?  It should be just:

> Sets whether TLS/SSL is enabled.

and perhaps add:

> (required for all Atlas clusters)

source/fundamentals/connection/mongoclientsettings.txt
```diff
@@ -0,0 +1,367 @@
...
+
+   * - ``invalidHostNameAllowed()``
+     - Whether to allow a mismatch between the server’s hostname and the hostname specified by the TLS certificate.
+
+Example
+~~~~~~~
+
+This example specifies the driver to ensure your Atlas connection string
```
jyemin, line 359: "specifies the driver" seems ungrammatical.  Can you rework this sentence (and perhaps similar ones above)?

Also, I don't think we should make this Atlas-specific.  Ops Manager also enables TLS, for example, as do many customers who deploy MongoDB themselves.

jyemin (changes requested):

PR #194, opened 2022-01-06
Docsp 19628 slim

terakilobyte:
## Pull Request Info

Exposes `maxConnecting`. This has been split from the original PR since that content needs to be seen by more eyes since it is prescriptive in connection tuning.

It also adds in a checker with github actions.

### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-19628

### Snooty build log:
https://workerpool-boxgs.mongodbstitch.com/pages/job.html?collName=queue&jobId=61d71ae241accb36335bad85

https://docs-mongodborg-staging.corp.mongodb.com/java/docsworker-xlarge/DOCSP-19628-slim/

### Self-Review Checklist

- [ ] Is this free of any warnings or errors in the RST?
- [ ] Does it render on staging correctly?
- [ ] Are the staging and workerpool job links in the PR description updated?

### DIFF
--- a/.github/workflows/checker.yml
+++ b/.github/workflows/checker.yml
@@ -1,4 +1,4 @@
-name: Checker
+name: Link Checker
 on: [push, pull_request]
 jobs:
   build:
--- a/.github/workflows/checker.yml
+++ b/.github/workflows/checker.yml
@@ -1,0 +1,22 @@
+name: Checker
+on: [push, pull_request]
+jobs:
+  build:
+    name: Build
+    runs-on: ubuntu-latest
+    steps:
+      - name: Set up Go 1.17
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.17
+
+      - name: Check out source code
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 2
+
+      - name: Install checker tool
+        run: go install github.com/terakilobyte/checker@latest
+
+      - name: Run checker tool
+        run: git diff --name-only HEAD^..HEAD | tr "\n" "," | xargs checker --changes
--- a/source/fundamentals/connection/connection-options.txt
+++ b/source/fundamentals/connection/connection-options.txt
@@ -208,8 +208,8 @@ parameters of the connection URI to specify the behavior of the client.
 
    * - **maxConnecting**
      - integer
-     - Specifies the number of connections the driver concurrently
-       establishes before making them available in the connection pool.
+     - Specifies the number of connections the driver will driver concurrently
+       establish before making them available in the connection pool.
 
 For a complete list of options, see the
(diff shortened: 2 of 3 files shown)


.github/workflows/checker.yml
```diff
@@ -0,0 +1,22 @@
...
+      - name: Check out source code
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 2
+
+      - name: Install checker tool
+        run: go install github.com/terakilobyte/checker@latest
+
+      - name: Run checker tool
+        run: git diff --name-only HEAD^..HEAD | tr "\n" "," | xargs checker --changes
```
ccho-mongodb, line 22: Aside from the issue of other directives or their targets changing external to the repo, I wonder if this misses any links that might be change by a modification to snooty.toml source constants. Does it take too much time to run if you leave out this optimization?

.github/workflows/checker.yml
```diff
@@ -0,0 +1,22 @@
+name: Checker
```
ccho-mongodb, line 1: How about link checker?

.github/workflows/checker.yml
terakilobyte: On large repos it can take a long time. For example, `docs` has ~9.2K links to check. The other issue is, and the next major improvement, would be to introduce more sophisticated throttling like [muffet](https://github.com/raviqqe/muffet)

.github/workflows/checker.yml
terakilobyte: Sure, the name isn't important to me!

.github/workflows/checker.yml
ccho-mongodb: Agreed it was ok in an offline discussion.
One additional point: if you set this up to run on every commit to every repo, you'll potentially get a lot of duplicate checking. While I don't think there's a way to conditionally trigger github actions, maybe you exit unless the latest commit message contains something like "check links". Open to other solutions too.

source/fundamentals/connection/connection-options.txt
```diff
@@ -206,6 +206,11 @@ parameters of the connection URI to specify the behavior of the client.
      - boolean
      - Specifies that the driver must connect to the host directly.
 
+   * - **maxConnecting**
+     - integer
+     - Specifies the number of connections the driver concurrently
+       establishes before making them available in the connection pool.
```
ccho-mongodb, line 212: Suggestion:

> Specifies the number of concurrent connections the driver establishes before making them available in the connection pool.

The adverb "concurrently" should describe the connections not how the driver establishes them.

ccho-mongodb (approved):
LGTM, one suggestion for clarity

PR #208, opened 2022-02-16
DOCSP-20939: what's new updates for 4.5

ccho-mongodb:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-20939

### Snooty build log:
https://workerpool-boxgs.mongodbstitch.com/pages/job.html?collName=queue&jobId=620d7c189c4c7f6bfb1099de

https://docs-mongodbcom-staging.corp.mongodb.com/java/docsworker-xlarge/DOCSP-20939-whats-new-4.5/whats-new/#what-s-new-in-4.5

### Self-Review Checklist

- [x] Are the staging and workerpool job links in the PR description updated?

### DIFF
--- a/source/whats-new.txt
+++ b/source/whats-new.txt
@@ -34,11 +34,13 @@ New features of the 4.5 Java driver release include:
 - Added support for overriding listener settings in ``ClusterSettings``,
   ``ServerSettings`` and ``ConnectionPoolSettings`` builders. 
 - All events received by ``ClusterListener``, ``ServerListener``, and
-  ``ServerMonitorListener`` are now ordered when they are not shared across 
-  different instances of ``MongoClient``. 
-- Added ``EnumCodec``, ``EnumCodecProvider``, and
-  ``EnumPropertyCodecProvider`` classes to separate codec support for enum 
-  types from the ``PojoCodec`` class.
+  ``ServerMonitorListener`` are now **totally ordered** using
+  a happens-before relationship when the listeners are not shared by
+  different MongoClient instances. To learn more about the happens-before
+  ordering, see `Happens-before Order <https://docs.oracle.com/javase/specs/jls/se17/html/jls-17.html#jls-17.4.5>`__
+  in the Java Language Specification.
+- Added ``EnumCodec`` and ``EnumCodecProvider`` classes to separate codec 
+  support for enum types from the ``PojoCodec`` class.
 - Resolved performance issues that impacted versions 4.4 and 4.3 of the
   driver. Performance in this version should be similar to performance in 4.2.
 
--- a/source/whats-new.txt
+++ b/source/whats-new.txt
@@ -12,11 +12,27 @@ What's New
 
 Learn what's new in:
 
+* :ref:`Version 4.5 <version-4.5>`
 * :ref:`Version 4.4 <version-4.4>`
 * :ref:`Version 4.3 <version-4.3>`
 * :ref:`Version 4.2 <version-4.2>`
 * :ref:`Version 4.1 <version-4.1>`
 * :ref:`Version 4.0 <version-4.0>`
+
+.. _version-4.5:
+
+What's New in 4.5
+-----------------
+
+New features of the 4.5 Java driver release include:
+
+- Added support for specifying custom service names of SRV resource
+  records. See the ``srvServiceName`` setting in the :ref:`Connection
+  Options <connection-options>` guide for more information.
+- Added support to customize UUID representation to in a registry using
+  the ``CodecRegistries.withUuidRepresentation()`` method.
+- Added support for overriding listener settings in ``ClusterSettings``,
+  ``ServerSettings`` and ``ConnectionPoolSettings`` builders.
 
 .. _version-4.4:
 
--- a/source/whats-new.txt
+++ b/source/whats-new.txt
@@ -34,7 +34,7 @@ New features of the 4.5 Java driver release include:
 - Added support for overriding listener settings in ``ClusterSettings``,
   ``ServerSettings`` and ``ConnectionPoolSettings`` builders. 
 - All events received by ``ClusterListener``, ``ServerListener``, and
-  ``ServerMonitorListener`` are now **totally ordered** using
+  ``ServerMonitorListener`` are now totally ordered using
   a happens-before relationship when the listeners are not shared by
   different MongoClient instances. To learn more about the happens-before
   ordering, see `Happens-before Order <https://docs.oracle.com/javase/specs/jls/se17/html/jls-17.html#jls-17.4.5>`__
--- a/source/whats-new.txt
+++ b/source/whats-new.txt
@@ -29,10 +29,18 @@ New features of the 4.5 Java driver release include:
 - Added support for specifying custom service names of SRV resource
   records. See the ``srvServiceName`` setting in the :ref:`Connection
   Options <connection-options>` guide for more information.
-- Added support to customize UUID representation to in a registry using
+- Added support to customize UUID representation in a codec registry using
   the ``CodecRegistries.withUuidRepresentation()`` method.
 - Added support for overriding listener settings in ``ClusterSettings``,
-  ``ServerSettings`` and ``ConnectionPoolSettings`` builders.
+  ``ServerSettings`` and ``ConnectionPoolSettings`` builders. 
+- All events received by ``ClusterListener``, ``ServerListener``, and
+  ``ServerMonitorListener`` are now ordered when they are not shared across 
+  different instances of ``MongoClient``. 
+- Added ``EnumCodec``, ``EnumCodecProvider``, and
+  ``EnumPropertyCodecProvider`` classes to separate codec support for enum 
+  types from the ``PojoCodec`` class.
+- Resolved performance issues that impacted versions 4.4 and 4.3 of the
+  driver. Performance in this version should be similar to performance in 4.2.
 
 .. _version-4.4:
 


source/whats-new.txt
```diff
@@ -12,12 +12,28 @@ What's New
...
+
+New features of the 4.5 Java driver release include:
+
+- Added support for specifying custom service names of SRV resource
+  records. See the ``srvServiceName`` setting in the :ref:`Connection
+  Options <connection-options>` guide for more information.
+- Added support to customize UUID representation to in a registry using
+  the ``CodecRegistries.withUuidRepresentation()`` method.
+- Added support for overriding listener settings in ``ClusterSettings``,
+  ``ServerSettings`` and ``ConnectionPoolSettings`` builders.
```
stIncMale, line 35: I think we should also mention the following changes:

- https://jira.mongodb.org/browse/JAVA-4449. As a result of this change, all events received by `ClusterListener`, `ServerListener`, `ServerMonitorListener` are totally ordered provided that the listeners are not shared by different `MongoClient`s.
- https://jira.mongodb.org/browse/JAVA-4440
- https://jira.mongodb.org/browse/JAVA-4452. Improved the performance characteristics of 4.5 when compared with 4.4 and 4.3. Now they are back to be on par with 4.2.

stIncMale (changes requested):

ccho-mongodb:
@stIncMale thanks for the feedback. I've made the requested changes -- let me know if they match with what you're looking for. Thanks!

source/whats-new.txt
```diff
@@ -29,10 +29,18 @@ New features of the 4.5 Java driver release include:
...
   Options <connection-options>` guide for more information.
-- Added support to customize UUID representation to in a registry using
+- Added support to customize UUID representation in a codec registry using
   the ``CodecRegistries.withUuidRepresentation()`` method.
 - Added support for overriding listener settings in ``ClusterSettings``,
-  ``ServerSettings`` and ``ConnectionPoolSettings`` builders.
+  ``ServerSettings`` and ``ConnectionPoolSettings`` builders. 
+- All events received by ``ClusterListener``, ``ServerListener``, and
+  ``ServerMonitorListener`` are now ordered when they are not shared across 
+  different instances of ``MongoClient``.
```
stIncMale, line 38: ["totally ordered"](https://mathworld.wolfram.com/TotallyOrderedSet.html) is a term that cannot be replaced with "ordered".

I also propose replacing "they" with "listeners" as otherwise, a reader may interpret "they" as referring to "events".

source/whats-new.txt
```diff
@@ -29,10 +29,18 @@ New features of the 4.5 Java driver release include:
...
   the ``CodecRegistries.withUuidRepresentation()`` method.
 - Added support for overriding listener settings in ``ClusterSettings``,
-  ``ServerSettings`` and ``ConnectionPoolSettings`` builders.
+  ``ServerSettings`` and ``ConnectionPoolSettings`` builders. 
+- All events received by ``ClusterListener``, ``ServerListener``, and
+  ``ServerMonitorListener`` are now ordered when they are not shared across 
+  different instances of ``MongoClient``. 
+- Added ``EnumCodec``, ``EnumCodecProvider``, and
+  ``EnumPropertyCodecProvider`` classes to separate codec support for enum 
+  types from the ``PojoCodec`` class.
```
stIncMale, line 41: Only `EnumCodec` and `EnumCodecProvider` were added.

stIncMale (changes requested):

source/whats-new.txt
ccho-mongodb: Thanks for pointing that out. I'll need to explain the term since we can't assume the audience is familiar with it.

source/whats-new.txt
ccho-mongodb: Thanks, thought I had seen all the classes I listed in the related PR.

stIncMale: `EnumPropertyCodecProvider` was modified, not added. And its modifications are purely internal and not perceivable by users.

source/whats-new.txt
stIncMale: I don't think we need to explain the term: those to whom the order relations between the events matter, likely know the term. Just specifying the term as a hyperlink to the definition is more than enough.

stIncMale: Just to elaborate on "those to whom the order relations between the events matter, likely know the term". The full guarantee as specified in the API documentation is more specific: _"All events received by `ClusterListener`, `ServerListener`, `ServerMonitorListener` are totally ordered (and the event order implies the happens-before order), provided that the listeners are not shared by different `MongoClients`."_ It mentions the happens-before order. 100% of those who know what this is, know that total order is. If we go down the road explaining total order, we then need to also explain the happens-before order, which will require reciting a significant chunk of the Java memory model, i.e., will be completely impractical.

ccho-mongodb: Before I commit a revision of this, how does this look:

All events received by ``ClusterListener``, ``ServerListener``, and
``ServerMonitorListener`` are now **totally ordered** when the listeners
are not shared across different instances of ``MongoClient``. Events are
totally ordered when you can compare the values of any two of them.

stIncMale: > Events are totally ordered when you can compare the values of any two of them.

"Values" (I don't know what it is, the reader will also not know) of events are unrelated to the order mentioned here. The order mentioned here is just some abstract total order, which one cannot predict in advance. But it is related to the following two partial orders in a way that is useful for users:

- it is a superset of the causal order of these events
- it is tied to the happens-before order (which is defined by the Java memory model) by the words _"the event order implies the happens-before order"_.

> up to the docs writers to ensure it's widely communicated to the audiences we target this documentation towards (Jasmin and Lucas).

That's fair. I previously linked a page with the definition, and I don't see any way of shortening it.

stIncMale: > I don't see any way of shortening it

I guess, we could say "are now totally ordered, that is, for any pair of events, one is ordered before the other"

ccho-mongodb: > I guess, we could say "are now totally ordered, that is, for any pair of events one is ordered before the other"

Would this be close enough to your messaging in the suggestion above? It helps keep the number of clauses down, avoids breaks, and is therefore assumed to be more understandable:

"...are now totally ordered when listeners are not shared across different instances of MongoClient. Events are totally ordered when one precedes another when comparing any two events."

Please let me know if that works ^


Responses to address your other comments:

What I read about total order ([how it's enforced by the Comparable interface and relates to natural ordering](https://docs.oracle.com/javase/7/docs/api/java/lang/Comparable.html)) and as a [partial order](https://en.wikipedia.org/wiki/Total_order) are in a different context of the total order you are referring to. Sorry for the misunderstanding.

The issue with the link was that it's a mathematical definition and doesn't provide context to how it applies to the event listeners. A couple alternatives:

- You mentioned that total ordering includes a Java memory model [happened before](https://en.wikipedia.org/wiki/Happened-before) relation in ordering and causal ordering. If it's a superset of causal order and follows the happens-before order, would it be accurate to describe it as covering both criteria for happened-before and causal ordering, or is that inaccurate?

- Maybe including an example for that What's New item would help clarify what you want to communicate about total order?

stIncMale: > "...are now totally ordered when listeners are not shared across different instances of MongoClient. Events are totally ordered when one precedes another when comparing any two events."
> Please let me know if that works ^

Using the word "comparing" here is confusing despite being technically correct exactly because it alludes to the `Comparable` interface which is not actually related to the order we are talking about.

> What I read about total order ([how it's enforced by the Comparable interface and relates to natural ordering](https://docs.oracle.com/javase/7/docs/api/java/lang/Comparable.html)) and as a [partial order](https://en.wikipedia.org/wiki/Total_order) are in a different context of the total order you are referring to.

Right, one may define any number of order relations over a set, and those orders may have different meaning despite all being total orders, for example.

> * If it's a superset of causal order and follows the happens-before order, would it be accurate to describe it as covering both criteria for happened-before and causal ordering, or is that inaccurate?
> * Maybe including an example for that What's New item would help clarify what you want to communicate about total order?

I am not sure whether you are referring to an example of the text I would propose, or a code example explaining the change to a reader.

If it's the former, then I already expressed the text I would propose: _"All events received by `ClusterListener`, `ServerListener`, `ServerMonitorListener` are totally ordered (and the event order implies the happens-before order), provided that the listeners are not shared by different `MongoClient`s."_ Note that the happens-before order implies the causal order, which is why causality is not mentioned. Again, those who know what the happens-before order is, know that it implies causality as a result of [this section in the Java language specification](https://docs.oracle.com/javase/specs/jls/se17/html/jls-17.html#jls-17.4.8) (this is the single JMM section which formal semantics I cannot comprehend for the life of me, but a consequence of this statement is that happens-before implies causality).

If it's the latter, then a code example is inapplicable, but we can give an example of a consequence. Our [API documentation](https://github.com/mongodb/mongo-java-driver/blob/master/driver-core/src/main/com/mongodb/event/ClusterListener.java#L22-L27) states the following consequence of the aforementioned guarantee: _"This means that even if you have a single class implementing all of `ClusterListener`, `ServerListener`, `ServerMonitorListener`, it does not have to be thread-safe."_ Another example is that `ClusterOpeningEvent` is delivered before (in the [happens-before order](https://docs.oracle.com/javase/specs/jls/se17/html/jls-17.html#jls-17.4.5)) any `ServerHeartbeatStartedEvent`, provided that they originate from the same `MongoClient`.

I recon that the amount of resources that we have spent on trying to fit this single what's new item into the documentation guidelines is disproportionately higher than the value of such fitting.

stIncMale: > not sure what it means to imply an ordering

The phrase "the order X implies the order Y" means that for any two events e1, e2 such that e1 is ordered before e2 in X, e1 is also ordered before e2 in Y.

> see Happens-before Order in the Java 17 documentation

The name of the document is "The Java Language Specification". Specifying a version suggests that this version contains something that is of importance and is not present in other version, but it's not the case.

stIncMale: > maybe we can create a follow-up PR so I can release the 4.5 docs now to make them available to the public, and make any fixes later

Chris, my understanding is that merging doc PRs is an option at any point: I review and reply only because I was requested either directly, or because request was directed to someone from the Java driver team. If you say at any point "thanks, I'll take it from here, no more help is needed", I will be happy to stop.

stIncMale (approved):

source/whats-new.txt
ccho-mongodb: > Chris, my understanding is that merging doc PRs is an option at any point: I review and reply only because I was requested either directly, or because request was directed to the Java driver team. If you say at any point "thanks, I'll take it from here, no more help is needed", I will be happy to stop.

Our team follows a process very similar to that of the [Realm Docs team](https://github.com/mongodb/docs-realm/blob/master/REVIEWING.md) (organizationally, we are part of the same docs team), so we look forward to reviewers like you to provide technical guidance and corrections. For these cases, we seek technical review approval before declaring something complete as we don't want to misguide our audience. We're working to better communicate these reviewing guidelines with our technical reviewers.

When something is inaccurate, it's helpful for writers if the technical reviewer imparts all the knowledge necessary for the writer to convey the knowledge to the reader, according to specific style guidelines and presentation that cater to the appropriate audiences. It's tough for writers to impart knowledge without having a solid grasp, both in presentation of docs as well as direct interactions with our readers.

Thanks to the various sources of information you provided, I do understand the logical rules of the ordering in "happens-before order" and how it applies to ordering across all the listeners. It might be helpful to expedite the conversation initially by collaborating to figure out what information gaps exist that need to be filled for the writer to produce something both parties will be satisfied with.

stIncMale: Thank you for the link to the reviewing guidelines.

> It's tough for writers to impart knowledge without having a solid grasp

Of course. It's impossible for technical writers to know everything that is collectively known by engineers working on different features of different products written in different programming languages. That's why when I anticipate that there may not be enough knowledge, I leave detailed descriptions or even propose wordings that could be used in the docs. But then in the process of trying to fit those into the documentation guidelines, which is to a significant degree means "to dumb down", the meaning is often distorted. That, in turn, causes back and forth communications that consume the amount of resources that is not proportionate to the potential benefits, whatever they are deemed to be.

> it's helpful for writers if the technical reviewer imparts all the knowledge necessary for the writer to convey the knowledge to the reader

I am trying to explain things and provide as much information as reasonable. I think that some people may even consider that to be too much, which is one of the reasons I don't dump everything I may say on a topic (another reason is that doing so is just impractical). But this takes lots of time, which brings us back to the inefficiency of the process.

Now, why did I write this comment? It's just unrequested feedback, which I will try to avoid repeating in the future.

PR #216, opened 2022-03-21
(DOCSP-20371) Subdomain Consolidation Updates

biniona-mongodb:
## Pull Request Info

Update docs set for subdomain consolidation.

[PR Reviewing Guidelines](https://github.com/mongodb/docs-java/blob/master/REVIEWING.md)

### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-20371

### Snooty build log:

https://workerpool-boxgs.mongodbstitch.com/pages/job.html?collName=queue&jobId=6239de5c0ab54725f138f21e

### Docs staging link (requires sign-in on MongoDB Corp SSO):

https://docs-mongodborg-staging.corp.mongodb.com/java/docsworker-xlarge/DOCSP-20371-Subdomain-Consolidation/

### Self-Review Checklist

- [x] Is this free of any warnings or errors in the RST?
- [ ] Did you run a spell-check?
- [ ] Did you run a grammar-check?
- [x] Does it render on staging correctly?
- [x] Are all the links working?
- [x] Are the staging and workerpool job links in the PR description updated?

### If your page documents a concept, does it meet the following criteria?

- [ ] Target the [Jasmin persona](https://drive.google.com/file/d/14FbBOLCVxwSP6M9BK4Nz1Ir9tzxT8_02/view)
- [ ] Target the [Lucas persona](https://drive.google.com/file/d/1J2vqJxo7ldv7OP_obA9Q-avf0o_ju4Lk/view)

PR #217, opened 2022-03-21
Update redirects base for subdomain consolidation

schmalliso:


schmalliso:
@terakilobyte Good to go for you!

biniona-mongodb:
Hi @schmalliso, just for context I recently merged in #216. I definitely could be incorrect, but it looks like this PR (#217) may make similar updates. Should #217 be closed, or is this adding something not addressed by #216?

terakilobyte (approved):

schmalliso:
This is in fact adding something #216 didn't address in that it moves the `docs/` to the prefix which ensures that the redirect created by the `raw: blabla` row actually ends up in the right folder in S3. Thanks @biniona-mongodb for your updates, though, they're :chefskiss:!

schmalliso:
@terakilobyte I've convinced github that the 17 attempts made at merge conflict resolution were indeed things that happened and now we're happy here.

terakilobyte (approved):

PR #227, opened 2022-04-25
DOCSP-22100 4.6 updates

ccho-mongodb:
# Pull Request Info

API documentation links do not work because GH docs are not yet published. Due to a circular reference between GH docs and MongoDB docs, they will be published around the same time to minimize time in which links are broken.

JIRA tickets (release notes and source constants update):

 https://jira.mongodb.org/browse/DOCSP-22100
https://jira.mongodb.org/browse/DOCSP-21563
https://jira.mongodb.org/browse/DOCSP-21227
https://jira.mongodb.org/browse/DOCSP-21676
https://jira.mongodb.org/browse/DOCSP-21423
https://jira.mongodb.org/browse/DOCSP-22098

Staging - [What's New](https://docs-mongodbcom-staging.corp.mongodb.com/java/docsworker-xlarge/DOCSP-22100-4.6-updates/whats-new/)
[CSFLE (mongodb-crypt reference)](https://docs-mongodbcom-staging.corp.mongodb.com/java/docsworker-xlarge/DOCSP-22100-4.6-updates/fundamentals/csfle/)

## Self-Review Checklist

### DIFF
--- a/source/whats-new.txt
+++ b/source/whats-new.txt
@@ -41,8 +41,8 @@ New features of the 4.6 Java driver release include:
   a custom DNS client. See the `DnsClient <https://mongodb.github.io/mongo-java-driver/4.6/apidocs/mongodb-driver-core/com/mongodb/spi/dns/DnsClient.html>`__
   and the `DnsClientProvider <https://mongodb.github.io/mongo-java-driver/4.6/apidocs/mongodb-driver-core/com/mongodb/spi/dns/DnsClientProvider.html>`__
   interface API documentation for more information.
-- Added driver support for encoding and decoding between Java Records and
-  BSON documents, which is enabled by default.
+- Added driver support for encoding and decoding between `Java records <https://docs.oracle.com/en/java/javase/17/language/records.html>`__
+  and BSON documents, which is enabled by default.
 
 .. _version-4.5:
 
--- a/source/whats-new.txt
+++ b/source/whats-new.txt
@@ -70,6 +70,12 @@ New features of the 4.5 Java driver release include:
   you order them as described in the section on :ref:`overriding codecs <codecs-override>`.
 - Resolved performance issues that impacted versions 4.4 and 4.3 of the
   driver. Performance in this version should be similar to performance in 4.2.
+- Resolved an issue in which errors originating from retrieving the cluster
+  description weren't passed to the ``onError`` Subscriber callback
+- Resolved an issue with releasing ``ByteBuf`` instances when you connect
+  with compression enabled.
+- Removed an unnecessary dependency on the ``javax.annotation.*`` packages
+  from the ``org.mongodb.driver-core`` OSGi bundle.
 
 .. _version-4.4:
 
--- a/source/whats-new.txt
+++ b/source/whats-new.txt
@@ -35,6 +35,10 @@ New features of the 4.6 Java driver release include:
   Encryption (CSFLE) the same way as specifying authentication credentials
   using :ref:`MONGODB-AWS <mongodb-aws-auth-mechanism>`.
 - Added CSFLE library support for Apple M1 ARM-based hardware.
+- Added the ``DnsClient`` and ``DnsClientProvider`` interfaces to allow
+  custom DNS client implementations. Added a service provider interface (SPI)
+  to specify the custom DNS client using ``java.util.ServiceLoader``. The
+  driver uses a default implementation if the application doesn't specify one.
 
 .. _version-4.5:
 
--- a/source/whats-new.txt
+++ b/source/whats-new.txt
@@ -41,6 +41,8 @@ New features of the 4.6 Java driver release include:
   a custom DNS client. See the `DnsClient <https://mongodb.github.io/mongo-java-driver/4.6/apidocs/mongodb-driver-core/com/mongodb/spi/dns/DnsClient.html>`__
   and the `DnsClientProvider <https://mongodb.github.io/mongo-java-driver/4.6/apidocs/mongodb-driver-core/com/mongodb/spi/dns/DnsClientProvider.html>`__
   interface API documentation for more information.
+- Added the `` org.mongodb:bson-record-codec`` library for encoding and
+  decoding Java records using the ``RecordCodec`` and ``RecordCodecProvider``.
 
 .. _version-4.5:
 
--- a/source/whats-new.txt
+++ b/source/whats-new.txt
@@ -41,8 +41,8 @@ New features of the 4.6 Java driver release include:
   a custom DNS client. See the `DnsClient <https://mongodb.github.io/mongo-java-driver/4.6/apidocs/mongodb-driver-core/com/mongodb/spi/dns/DnsClient.html>`__
   and the `DnsClientProvider <https://mongodb.github.io/mongo-java-driver/4.6/apidocs/mongodb-driver-core/com/mongodb/spi/dns/DnsClientProvider.html>`__
   interface API documentation for more information.
-- Added the `` org.mongodb:bson-record-codec`` library for encoding and
-  decoding Java records using the ``RecordCodec`` and ``RecordCodecProvider``.
+- Added driver support for encoding and decoding between Java Records and
+  BSON documents, which is enabled by default.
 
 .. _version-4.5:
 
--- a/source/whats-new.txt
+++ b/source/whats-new.txt
@@ -12,12 +12,30 @@ What's New
 
 Learn what's new in:
 
+* :ref:`Version 4.6 <version-4.6>`
 * :ref:`Version 4.5 <version-4.5>`
 * :ref:`Version 4.4 <version-4.4>`
 * :ref:`Version 4.3 <version-4.3>`
 * :ref:`Version 4.2 <version-4.2>`
 * :ref:`Version 4.1 <version-4.1>`
 * :ref:`Version 4.0 <version-4.0>`
+
+
+.. _version-4.6:
+
+What's New in 4.6
+-----------------
+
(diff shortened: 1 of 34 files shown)


source/whats-new.txt
```diff
@@ -12,13 +12,34 @@ What's New
...
+
+- The buffer pool is now shared across all ``MongoClient`` instances. By
+  default, the buffer pool prunes buffers that have been idle for one
+  minute.
+- Added an API to specify AWS credentials for Client-Side Field Level
+  Encryption (CSFLE) the same way as specifying authentication credentials
+  using :ref:`MONGODB-AWS <mongodb-aws-auth-mechanism>`.
+- Added CSFLE library support for Apple M1 ARM-based hardware.
+- Added the ``DnsClient`` and ``DnsClientProvider`` interfaces to allow
+  custom DNS client implementations. Added a service provider interface (SPI)
```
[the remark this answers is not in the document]
ccho-mongodb: It's the same feature; the first part describes the interface you can use for a custom implementation and the second part describes how you can specify the custom implementation.

source/whats-new.txt
```diff
@@ -12,13 +12,34 @@ What's New
...
+
+New features of the 4.6 Java driver release include:
+
+- The buffer pool is now shared across all ``MongoClient`` instances. By
+  default, the buffer pool prunes buffers that have been idle for one
+  minute.
+- Added an API to specify AWS credentials for Client-Side Field Level
+  Encryption (CSFLE) the same way as specifying authentication credentials
+  using :ref:`MONGODB-AWS <mongodb-aws-auth-mechanism>`.
+- Added CSFLE library support for Apple M1 ARM-based hardware.
```
stIncMale, line 37: Instead of mysteriously referring to the library as "CSFLE library", we may just identify it as `org.mongodb:mongodb-crypt` because users specify it explicitly as a dependency, i.e., are aware of its formal name.

source/whats-new.txt
```diff
@@ -12,13 +12,34 @@ What's New
...
+- Added the ``DnsClient`` and ``DnsClientProvider`` interfaces to allow
+  custom DNS client implementations. Added a service provider interface (SPI)
+  to specify the custom DNS client using ``java.util.ServiceLoader``. The
+  driver uses a default implementation if the application doesn't specify one.
```
stIncMale, line 41: Saying _"Added a service provider interface (SPI)"_ after _"Added the ``DnsClient`` and ``DnsClientProvider`` interfaces"_ is weird because `DnsClient` and `DnsClientProvider` together is the SPI we added. I propose to write the following:

> Added the ``DnsClient`` and ``DnsClientProvider`` interfaces which represent a service provider interface (SPI) that allows customizing the DNS client used by the driver.

We don't even need to mention ``java.util.ServiceLoader``, because the documentation of the `DnsClientProvider` states _"The driver discovers implementations of this interface via `java.util.ServiceLoader`."_

stIncMale (changes requested):
An important feature appears to be missing: users may now add a dependency on `org.mongodb:bson-record-codec` and get support for encoding/decoding Java records via `RecordCodecProvider` and `RecordCodec`: https://jira.mongodb.org/browse/JAVA-3567.

source/whats-new.txt
```diff
@@ -12,13 +12,34 @@ What's New
...
+-----------------
+
+New features of the 4.6 Java driver release include:
+
+- The buffer pool is now shared across all ``MongoClient`` instances. By
+  default, the buffer pool prunes buffers that have been idle for one
+  minute.
+- Added an API to specify AWS credentials for Client-Side Field Level
+  Encryption (CSFLE) the same way as specifying authentication credentials
+  using :ref:`MONGODB-AWS <mongodb-aws-auth-mechanism>`.
```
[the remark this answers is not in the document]
jyemin: Yeah, the API is different.  The part that's the same is the functionality.

* That AWS credentials are discovered in the same way (though with a different API)
* That there is a callback for the application to provide the credentials (thought with a different API)

source/whats-new.txt
jyemin: I don't mind repeating in the reference docs what's already in the API docs.  We do that in many places I am sure.

source/whats-new.txt
ccho-mongodb: I agree that including the library name is useful in this instance, but the feature name is also important to keep.

Given this is page is targeted towards both Jasmin and Lucas, I think having a high level description of what the library is for is important to us as well as it otherwise becomes "mysterious" to audiences not familiar with the dependency name.

source/whats-new.txt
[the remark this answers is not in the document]
ccho-mongodb: I can see how it's confusing in the current form. I'll rephrase and link to the methods listed in your suggestion.

ccho-mongodb:
@stIncMale and @jyemin : I addressed your PR comments in the latest commits, please take a look when you get the chance.

stIncMale (changes requested):
I still think that we need to mention the support for Java records. I suggested this originally [here](https://github.com/mongodb/docs-java/pull/227#pullrequestreview-952470122).

stIncMale:
I think the changes worth including in "what's new" can be identified as those marked with "Documentation Changes: Needed" in Jira. But then https://jira.mongodb.org/browse/JAVA-3567 has not been marked this way.

@jyemin, @ccho-mongodb, does the above suggestion looks OK to you? If yes, then let's agree to follow the approach.

jyemin:
It was an oversight that Java records support was not marked as Docs Changes Needed.  I think it's worth its own section with an example, in addition to What's New.

ccho-mongodb:
> It was an oversight that Java records support was not marked as Docs Changes Needed. I think it's worth its own section with an example, in addition to What's New.

I agree and created a DOCSP ticket a little earlier: https://jira.mongodb.org/browse/DOCSP-22293
Feel free to add any guidance there. Is the DBX team planning to create API documentation for `org.mongodb:bson-record-codec` ?

jyemin:
> Is the DBX team planning to create API documentation for org.mongodb:bson-record-codec?

We don't have any current plans to add more documentation beyond the minimal already created.

source/whats-new.txt
```diff
@@ -12,13 +12,38 @@ What's New
...
+  `AutoEncryptionSettings.Builder.kmsProviderPropertySuppliers() <https://mongodb.github.io/mongo-java-driver/4.6/apidocs/mongodb-driver-core/com/mongodb/AutoEncryptionSettings.Builder.html#kmsProviderPropertySuppliers(java.util.Map)>`__
+  and the `ClientEncryptionSettings.Builder.kmsProviderPropertySuppliers() <https://mongodb.github.io/mongo-java-driver/4.6/apidocs/mongodb-driver-core/com/mongodb/ClientEncryptionSettings.html#getKmsProviderPropertySuppliers()>`__
+  API documentation for more information.
+- Added compatibility to ``org.mongodb:mongodb-crypt``, a dependency for
+  CSFLE, for Apple M1 ARM-based hardware.
+- Added a service provider interface (SPI) to allow the driver to specify
+  a custom DNS client. See the `DnsClient <https://mongodb.github.io/mongo-java-driver/4.6/apidocs/mongodb-driver-core/com/mongodb/spi/dns/DnsClient.html>`__
+  and the `DnsClientProvider <https://mongodb.github.io/mongo-java-driver/4.6/apidocs/mongodb-driver-core/com/mongodb/spi/dns/DnsClientProvider.html>`__
+  interface API documentation for more information.
+- Added the `` org.mongodb:bson-record-codec`` library for encoding and
```
jyemin, line 44: I'm not sure this is the best way to document this, for a couple of reasons

1. RecordCodec is not a public class
2. RecordCodecProvider is including in the default codec registry, so to make use of it you don't have to use RecordCodecProvider directly at all.

We should say something like:

> The driver supports encoding and decoding Java records (link) to and from BSON documents.  This support is enabled by default.

And perhaps link to API docs for com.mongodb.MongoClientSettings#getDefaultCodecRegistry

jyemin (changes requested):

source/whats-new.txt
ccho-mongodb: @jyemin From browsing Maven, I wasn't able to find which current JAR lists bson-record-codec as a dependency, so if that's the case, don't they need to add it for default support?

jyemin: mongodb-driver-core does: https://repo1.maven.org/maven2/org/mongodb/mongodb-driver-core/4.6.0/mongodb-driver-core-4.6.0.pom

source/whats-new.txt
```diff
@@ -41,8 +41,8 @@ New features of the 4.6 Java driver release include:
   a custom DNS client. See the `DnsClient <https://mongodb.github.io/mongo-java-driver/4.6/apidocs/mongodb-driver-core/com/mongodb/spi/dns/DnsClient.html>`__
   and the `DnsClientProvider <https://mongodb.github.io/mongo-java-driver/4.6/apidocs/mongodb-driver-core/com/mongodb/spi/dns/DnsClientProvider.html>`__
   interface API documentation for more information.
-- Added the `` org.mongodb:bson-record-codec`` library for encoding and
-  decoding Java records using the ``RecordCodec`` and ``RecordCodecProvider``.
+- Added driver support for encoding and decoding between Java Records and
```
jyemin, line 44: Nit: lowercase `Records`.
Nit: Add a link to https://docs.oracle.com/en/java/javase/17/language/records.html or https://docs.oracle.com/javase/specs/jls/se18/html/jls-8.html#jls-8.10

jyemin (approved):

PR #234, opened 2022-06-22
DOCSP-22866 clustered indexes

kyuan-mongodb:
# Pull Request Info

Added a section on the Indexes page on how to specify a clustered index.

JIRA - https://jira.mongodb.org/browse/DOCSP-22866
Staging - https://docs-mongodbcom-staging.corp.mongodb.com/java/docsworker-xlarge/DOCSP-22866/fundamentals/indexes/#clustered-indexes

## Self-Review Checklist

### DIFF
--- a/source/fundamentals/indexes.txt
+++ b/source/fundamentals/indexes.txt
@@ -405,9 +405,9 @@ insert, update, and delete operations on **clustered collections**.
 Clustered collections store documents ordered by the clustered index key
 value.
 
-To create a clustered index, specify the clustered index option when you
-create your collection. The clustered index option must specify the
-``_id`` field as the key and the unique field as ``true``.
+To create a clustered index, specify the clustered index option with the
+``_id`` field as the key and the unique field as ``true`` when you
+create your collection.
 
 The following example creates a clustered index on the ``_id`` field in
 the ``vendors`` collection:
--- a/source/fundamentals/indexes.txt
+++ b/source/fundamentals/indexes.txt
@@ -405,12 +405,12 @@ insert, update, and delete operations on **clustered collections**.
 Clustered collections store documents ordered by the clustered index key
 value.
 
-To create a clustered index, specify a ``ClusteredIndexOptions`` in
-your ``CreateCollectionOptions``. The ``ClusteredIndexOptions`` must
-specify the ``_id`` field as the key and the unique field as ``true``.
+To create a clustered index, specify the clustered index option when you
+create your collection. The clustered index option must specify the
+``_id`` field as the key and the unique field as ``true``.
 
 The following example creates a clustered index on the ``_id`` field in
-the ``tea.vendors`` collection:
+the ``vendors`` collection:
 
 .. code-block:: java
    
@@ -420,10 +420,10 @@ the ``tea.vendors`` collection:
 
    database.createCollection("vendors", createCollectionOptions);
 
-See the MongoDB server manual sections on 
-:ref:`Clustered Indexes <db.createCollection.clusteredIndex>` and 
-:ref:`Clustered Collections <clustered-collections>`
-for more information.
+See the MongoDB server manual sections for more information:
+
+- :ref:`Clustered Indexes <db.createCollection.clusteredIndex>`
+- :ref:`Clustered Collections <clustered-collections>`
 
 Remove an Index
 ---------------
--- a/source/fundamentals/indexes.txt
+++ b/source/fundamentals/indexes.txt
@@ -405,9 +405,9 @@ insert, update, and delete operations on **clustered collections**.
 Clustered collections store documents ordered by the clustered index key
 value.
 
-To create a clustered index, specify the ``ClusteredIndexOptions`` in
-the ``CreateCollectionOptions``. The ``ClusteredIndexOptions`` must
-specify the ``_id`` field as the key, and the unique field as ``true``.
+To create a clustered index, specify a ``ClusteredIndexOptions`` in
+your ``CreateCollectionOptions``. The ``ClusteredIndexOptions`` must
+specify the ``_id`` field as the key and the unique field as ``true``.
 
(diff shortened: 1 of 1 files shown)


source/fundamentals/indexes.txt
```diff
@@ -397,6 +397,34 @@ Refer to the :manual:`Unique Indexes page </core/index-unique>` in the MongoDB s
...
+specify the ``_id`` field as the key, and the unique field as ``true``.
+
+The following example creates a clustered index on the ``_id`` field in
+the ``tea.vendors`` collection:
+
+.. code-block:: java
+   
+   MongoDatabase database = mongoClient.getDatabase("tea");
+   ClusteredIndexOptions cioptions = new ClusteredIndexOptions(Filters.eq("_id", 1), true);
+   CreateCollectionOptions coptions = new CreateCollectionOptions().clusteredIndexOptions(cioptions);
```
ccho-mongodb, line 419: Issue: the variable names should follow variable naming convention. Generally, this means that you should use words instead of first letters of words, and camelCase between the words.

Suggestion:
ClusteredIndexOptinos clusteredIndexOptions = ...
CreateCollectionOptions createCollectionOptions = ...

source/fundamentals/indexes.txt
```diff
@@ -397,6 +397,34 @@ Refer to the :manual:`Unique Indexes page </core/index-unique>` in the MongoDB s
...
+Clustered Indexes
+~~~~~~~~~~~~~~~~~
+
+**Clustered indexes** are indexes that improve the performance of
+insert, update, and delete operations on **clustered collections**.
+Clustered collections store documents ordered by the clustered index key
+value.
+
+To create a clustered index, specify a ``ClusteredIndexOptions`` in
+your ``CreateCollectionOptions``. The ``ClusteredIndexOptions`` must
```
ccho-mongodb, line 409: Issue:
Not sure what "ClusteredIndexOptions" and "CreateCollectionOptions" are because they haven't been introduced yet and aren't trivial to figure out.

Suggestion:
Identify what they are instead of only referring to them by name.

source/fundamentals/indexes.txt
```diff
@@ -397,6 +397,34 @@ Refer to the :manual:`Unique Indexes page </core/index-unique>` in the MongoDB s
...
+specify the ``_id`` field as the key and the unique field as ``true``.
+
+The following example creates a clustered index on the ``_id`` field in
+the ``tea.vendors`` collection:
```
ccho-mongodb, line 413: Issue: the collection name used in the example is "vendors" and not "tea.vendors".
I think you can actually have a collection name that includes a "." character, so it is not trivial to figure out.

source/fundamentals/indexes.txt
```diff
@@ -397,6 +397,34 @@ Refer to the :manual:`Unique Indexes page </core/index-unique>` in the MongoDB s
...
+   MongoDatabase database = mongoClient.getDatabase("tea");
+   ClusteredIndexOptions clusteredIndexOptions = new ClusteredIndexOptions(Filters.eq("_id", 1), true);
+   CreateCollectionOptions createCollectionOptions = new CreateCollectionOptions().clusteredIndexOptions(clusteredIndexOptions);
+
+   database.createCollection("vendors", createCollectionOptions);
+
+See the MongoDB server manual sections on 
+:ref:`Clustered Indexes <db.createCollection.clusteredIndex>` and 
+:ref:`Clustered Collections <clustered-collections>`
+for more information.
```
ccho-mongodb, line 426: Suggestion: for consistency with other sections, when presenting multiple links, use a bulleted list.

ccho-mongodb (changes requested):
Looks mostly good; I found a couple minor issues that I'd like to review again. Thanks!

ccho-mongodb (approved):
LGTM, left one suggestion.

source/fundamentals/indexes.txt
```diff
@@ -397,6 +397,34 @@ Refer to the :manual:`Unique Indexes page </core/index-unique>` in the MongoDB s
 
 .. driver-content-end
 
+Clustered Indexes
+~~~~~~~~~~~~~~~~~
+
+**Clustered indexes** are indexes that improve the performance of
+insert, update, and delete operations on **clustered collections**.
+Clustered collections store documents ordered by the clustered index key
+value.
```
stIncMale, line 406: The part _"clustered indexes are indexes that improve the performance of ..."_ appears to be posed as the essence of what it means for an index to be clustered. However, it is actually not, and the essence is that the presence of a clustered index makes the collection to store documents ordered by the key (such a collection is called a clustered collection). I propose to specify only the essence on this page and refer users to the server documentation for any other details.

source/fundamentals/indexes.txt
```diff
@@ -397,6 +397,34 @@ Refer to the :manual:`Unique Indexes page </core/index-unique>` in the MongoDB s
...
+``_id`` field as the key and the unique field as ``true`` when you
+create your collection.
+
+The following example creates a clustered index on the ``_id`` field in
+the ``vendors`` collection:
+
+.. code-block:: java
+   
+   MongoDatabase database = mongoClient.getDatabase("tea");
+   ClusteredIndexOptions clusteredIndexOptions = new ClusteredIndexOptions(Filters.eq("_id", 1), true);
```
stIncMale, line 418: Since there is no API in the driver that is intended to construct a valid `Bson` key for `ClusteredIndexOptions`, we should not recommend users to use unrelated API solely because it so happens that the API creates a document valid for our situation (`Projections.include("_id")` also happens to produce a valid document). If we do this, then users may reasonably expect that specifying the output of other `Filters` methods here is allowed, but it's not. I think that we instead should write `new ClusteredIndexOptions(new Document("_id", 1), true)`.

stIncMale (changes requested):

stIncMale (approved):

PR #242, opened 2022-07-13
DOCSP-23027: BsonExtraElements Annotation

ccho-mongodb:
# Pull Request Info

JIRA - https://jira.mongodb.org/browse/DOCSP-23027
Staging - https://docs-mongodbcom-staging.corp.mongodb.com/java/docsworker-xlarge/DOCSP-23027-bsonextraelements-example/fundamentals/data-formats/pojo-customization/

## Self-Review Checklist

### DIFF
--- a/source/fundamentals/data-formats/pojo-customization.txt
+++ b/source/fundamentals/data-formats/pojo-customization.txt
@@ -424,9 +424,10 @@ the following JSON object:
 
 .. code-block:: json
    :copyable: false
+   :emphasize-lines: 5-9
 
    {
-     "name": "MDB0123,
+     "name": "MDB0123",
      "serialNumber": "62e2...",
      "relatedItems": null,
      "additionalInfo": {
--- a/source/fundamentals/data-formats/pojo-customization.txt
+++ b/source/fundamentals/data-formats/pojo-customization.txt
@@ -300,13 +300,17 @@ package:
        nested within the field.
 
    * - ``BsonExtraElements``
-     - Specifies the POJO field on which to deserialize any elements that are
+     - Specifies the POJO field on which to deserialize all elements that are
        not mapped to a field. The POJO field must be one of the following
        types:
 
        - `Document <{+api+}/apidocs/bson/org/bson/Document.html>`__
        - `BsonDocument <{+api+}/apidocs/bson/org/bson/BsonDocument.html>`__
        - Map<String, Object>
+
+       When serializing a POJO through an insert or replace operation, the
+       driver stores the value of the annotated field as a subdocument in
+       MongoDB.
 
        .. seealso::
 
@@ -382,45 +386,56 @@ BsonExtraElements Example
 
 The ``@BsonExtraElements`` annotation allows you to specify a field to
 deserialize data from a MongoDB document that lacks a corresponding POJO
-field mapping. Consider a situation in which your application needs to handle
-data that conforms to different data schema versions. When your POJO fields
-change to accommodate the schema, you can continue to read, but not
-update, the remainder of the MongoDB document data from the field that uses
-this annotation.
+field mapping. This is useful when your application needs to work with data in
+a partially-defined schema. You can use this annotation to access data from
+any fields that do not correspond to the fields on your POJO.
 
-Consider a situation in which you need to update the ``Product`` POJO from the
-:ref:`previous example <bson-annotation-code-example>` to accommodate the
-following changes to your data schema:
-
-- The ``name`` field is no longer needed, but must the data must remain for
-  legacy support.
-- The stock keeping unit (SKU) needs to be added.
-
-The updated version of the ``Product`` POJO should resemble the following code
-example:
+Consider a situation in which you store and retrieve data for an online store
+using the :ref:`Product POJO <bson-annotation-code-example>` from the previous
+example. As you offer a greater variety of products for the store, you
+discover that you need additional fields to describe them. Instead of
+mapping each additional field to the POJO, you can access them from a
+single field annotated with ``@BsonExtraElements`` as shown in the following
+code example:
 
 .. code-block:: java
-   :emphasize-lines: 6-7,9-10
+   :emphasize-lines: 12-13
 
    public class Product {
+       @BsonProperty("modelName")
+       private String name;
+
        @BsonId()
        @BsonRepresentation(BsonType.OBJECT_ID)
        private String serialNumber;
 
-       @BsonProperty("sku")
-       private String sku;
+       @BsonIgnore
+       private List<Product> relatedItems;
 
        @BsonExtraElements
-       private Document legacyInfo;
+       private Document additionalInfo;
 
        // ...
 
-Once you update the product records with SKU values in MongoDB, the data you
-retrieve with the ``Product`` POJO should resemble the following:
+If a database client updated the documents in the collection to include
+additional fields that describe the products such as ``dimensions`` and
+``weight``, a document retrieved using the ``Product`` POJO might resemble
+the following JSON object:
 
-.. code-block:: none
+.. code-block:: json
+   :copyable: false
 
-   [Product [sku=MDB0..., serialNumber=62cf..., legacyInfo=Document{{ name=file cabinet }]]
+   {
+     "name": "MDB0123,
+     "serialNumber": "62e2...",
+     "relatedItems": null,
+     "additionalInfo": {
+       {
+         "dimensions": "3x4x5",
+         "weight": "256g"
+       }
+   }
+
 
 .. _pojo-discriminators:
 
--- a/source/fundamentals/data-formats/pojo-customization.txt
+++ b/source/fundamentals/data-formats/pojo-customization.txt
@@ -370,6 +370,53 @@ The annotations in the example POJO specify the following behavior:
 - Omit the ``relatedItems`` field and value when converting data
 - Use the ``Product(String name)`` constructor when instantiating the POJO
 
+
+.. _bsonextraelements-annotation-code-example:
+
+BsonExtraElements Example
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The ``@BsonExtraElements`` annotation allows you to specify a field to
+deserialize data from a MongoDB document that lacks a corresponding POJO
+field mapping. Consider a situation in which your application needs to handle
+data that conforms to different data schema versions. When your POJO fields
+change to accommodate the schema, you can continue to read, but not
+update, the remainder of the MongoDB document data from the field that uses
+this annotation.
+
+Suppose you need to update the ``Product`` POJO from the
+:ref:`previous example <bson-annotation-code-example>` to accommodate the
+following changes to your data schema:
+
+- The ``name`` field is no longer needed, but should remain for legacy support
+- The stock keeping unit (SKU) needs to be added
+
+The updated version of the ``Product`` POJO should resemble the following code
+example:
+
+.. code-block:: java
+   :emphasize-lines: 6-7,9-10
+
+   public class Product {
+       @BsonId()
+       @BsonRepresentation(BsonType.OBJECT_ID)
+       private String serialNumber;
+
+       @BsonProperty("sku")
+       private String sku;
+
+       @BsonExtraElements
+       private Document legacyInfo;
+
+       // ...
+
+Once you update the product records with SKU values in MongoDB, the data you
+retrieve with the ``Product`` POJO should resemble the following:
+
+.. code-block:: none
+
+   [NewProduct [sku=MDB0..., serialNumber=62cf..., legacyInfo=Document{{ name=file cabinet }]]
+
 .. _pojo-discriminators:
 
 Discriminators
--- a/source/fundamentals/data-formats/pojo-customization.txt
+++ b/source/fundamentals/data-formats/pojo-customization.txt
@@ -306,11 +306,7 @@ package:
 
        - `Document <{+api+}/apidocs/bson/org/bson/Document.html>`__
        - `BsonDocument <{+api+}/apidocs/bson/org/bson/BsonDocument.html>`__
-       - Map<String, Object>
-
-       When serializing a POJO through an insert or replace operation, the
-       driver stores the value of the annotated field as a subdocument in
-       MongoDB.
+       - ``Map<String, Object>``
 
        .. seealso::
 
--- a/source/fundamentals/data-formats/pojo-customization.txt
+++ b/source/fundamentals/data-formats/pojo-customization.txt
@@ -304,12 +304,16 @@ package:
        not mapped to a field. The POJO field must be one of the following
        types:
 
-       - Document
-       - BsonDocument
+       - `Document <{+api+}/apidocs/bson/org/bson/Document.html>`__
+       - `BsonDocument <{+api+}/apidocs/bson/org/bson/BsonDocument.html>`__
        - Map<String, Object>
 
+       .. seealso::
+
+          :ref:`BsonExtraElements Annotation Example <bsonextraelements-annotation-code-example>`
+
 The following code snippet shows a sample POJO called ``Product`` that uses
-the preceding annotations.
+several of the preceding annotations.
 
 .. _bson-annotation-code-example:
 
@@ -384,12 +388,13 @@ change to accommodate the schema, you can continue to read, but not
 update, the remainder of the MongoDB document data from the field that uses
 this annotation.
 
-Suppose you need to update the ``Product`` POJO from the
+Consider a situation in which you need to update the ``Product`` POJO from the
 :ref:`previous example <bson-annotation-code-example>` to accommodate the
 following changes to your data schema:
 
-- The ``name`` field is no longer needed, but should remain for legacy support
-- The stock keeping unit (SKU) needs to be added
+- The ``name`` field is no longer needed, but must the data must remain for
+  legacy support.
+- The stock keeping unit (SKU) needs to be added.
 
 The updated version of the ``Product`` POJO should resemble the following code
 example:
@@ -415,7 +420,7 @@ retrieve with the ``Product`` POJO should resemble the following:
 
 .. code-block:: none
 
-   [NewProduct [sku=MDB0..., serialNumber=62cf..., legacyInfo=Document{{ name=file cabinet }]]
+   [Product [sku=MDB0..., serialNumber=62cf..., legacyInfo=Document{{ name=file cabinet }]]
 
 .. _pojo-discriminators:
 
--- a/source/fundamentals/data-formats/pojo-customization.txt
+++ b/source/fundamentals/data-formats/pojo-customization.txt
@@ -415,24 +415,19 @@ code example:
 
 If a database client updated the documents in the collection to include
 additional fields that describe the products such as ``dimensions`` and
-``weight``, a document retrieved using the ``Product`` POJO might resemble
-the following JSON object:
+``weight``, the document retrieved using the ``Product`` POJO might resemble
+the following:
 
-.. code-block:: json
+.. code-block::
    :copyable: false
-   :emphasize-lines: 5-9
+   :emphasize-lines: 5
 
-   {
-     "name": "MDB0123",
-     "serialNumber": "62e2...",
-     "relatedItems": null,
-     "additionalInfo": {
-       {
-         "dimensions": "3x4x5",
-         "weight": "256g"
-       }
-   }
-
+   ProductWithBsonExtraElements [
+     name=MDB0123,
+     serialNumber=62eb...,
+     relatedItems=null,
+     additionalInfo=Document{{dimensions=3x4x5, weight=256g}}
+   ]
 
 .. _pojo-discriminators:
 
--- a/source/fundamentals/data-formats/pojo-customization.txt
+++ b/source/fundamentals/data-formats/pojo-customization.txt
@@ -300,7 +300,13 @@ package:
        nested within the field.
 
    * - ``BsonExtraElements``
-     - Specifies any extra elements that aren't already mapped to a field.
+     - Specifies the POJO field on which to deserialize any elements that are
+       not mapped to a field. The POJO field must be one of the following
+       types:
+
+       - Document
+       - BsonDocument
+       - Map<String, Object>
 
 The following code snippet shows a sample POJO called ``Product`` that uses
 the preceding annotations.
--- a/source/fundamentals/data-formats/pojo-customization.txt
+++ b/source/fundamentals/data-formats/pojo-customization.txt
@@ -413,10 +413,22 @@ code example:
 
        // ...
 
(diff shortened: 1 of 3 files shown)


source/fundamentals/data-formats/pojo-customization.txt
```diff
@@ -364,6 +374,54 @@ The annotations in the example POJO specify the following behavior:
...
+data that conforms to different data schema versions. When your POJO fields
+change to accommodate the schema, you can continue to read, but not
+update, the remainder of the MongoDB document data from the field that uses
+this annotation.
+
+Consider a situation in which you need to update the ``Product`` POJO from the
+:ref:`previous example <bson-annotation-code-example>` to accommodate the
+following changes to your data schema:
+
+- The ``name`` field is no longer needed, but must the data must remain for
```
jyemin, line 395: but must -> but

source/fundamentals/data-formats/pojo-customization.txt
```diff
@@ -364,6 +374,54 @@ The annotations in the example POJO specify the following behavior:
...
+
+BsonExtraElements Example
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The ``@BsonExtraElements`` annotation allows you to specify a field to
+deserialize data from a MongoDB document that lacks a corresponding POJO
+field mapping. Consider a situation in which your application needs to handle
+data that conforms to different data schema versions. When your POJO fields
+change to accommodate the schema, you can continue to read, but not
+update, the remainder of the MongoDB document data from the field that uses
```
jyemin, line 388: I'm not sure what you mean that you can read but not update the remainder of the data.  It seems like you can both read _and_ update.

source/fundamentals/data-formats/pojo-customization.txt
```diff
@@ -364,6 +374,54 @@ The annotations in the example POJO specify the following behavior:
...
+this annotation.
+
+Consider a situation in which you need to update the ``Product`` POJO from the
```
jyemin, line 391: This is a great example, and actually one I hadn't thought of.  I think the more common case though is where you have a semi-structured schema where you want to model the structured part as POJO fields while still being able to read/update the unstructured part as a Document.  Perhaps we could add an example for that as well, or at least mention the use case?

jyemin (changes requested):

source/fundamentals/data-formats/pojo-customization.txt
ccho-mongodb: To clarify, you're referring to generally being able to update the data and not when using a POJO field with the BsonExtraElements annotation? When reading the [description of the annotation](https://jira.mongodb.org/browse/DOCSP-22048), I inferred that the BsonExtraElements annotation would decode to a POJO, but not encode the field's value and pass it back to MongoDB.

I think while rewriting this section, I placed this info in a spot that made it confusing as to whether it was a requirement of the hypothetical situation or general capabilities of using a POJO to work with MongoDB data.

ccho-mongodb: From reading your next comment, it sounds like you can update the unstructured part and that the field data gets encoded in an update operation. Will test it out and update the descriptions accordingly.

ccho-mongodb: Wasn't able to update the MongoDB data using a POJO field marked with the annotation, so I'm assuming you meant you could update POJO fields generally, but not with the BsonExtraElements annotation.

jyemin: I just tried it and it worked as I expected.  Can you share your reproducer?

ccho-mongodb: https://gist.github.com/ccho-mongodb/98a7a6839c2b7627fb0cc2637151b2fa

I was able to set the "material" field on the "metadata" subdocument (annotated with @BsonExtraElements) using a POJO only on insert. Any attempt to update that field using the POJO (shown in the gist) resulted in no changes.

Isn't that the expected behavior? If not, what am I doing incorrectly?

jyemin: OK, I think I see the problem.  The `metadata` field does not result in a sub-document.  The name of a field annotated with `BsonExtraElements` is actually ignored, and the fields within the document itself are added at the same level as the other fields.  So when you insert the `NewProductPojo` instance in your example, you should end up with:

```
{ 
    _id: "MDB0123",
   material: "iron"
}
```

and not

```
{ 
    _id: "MDB0123",
   metadata: { material: "iron"}
}
```

Does that clear it up?

ccho-mongodb: Yes, from your explanation, I think my original understanding matched what you're saying, but needed to be more specific in the wording of the explanation in the docs.

To summarize, you can change the underlying data, but not from the BsonExtraElements-annotated field since it's only decoded and not encoded. E.g. I can update the MongoDB document by directly updating the "material" field, but not from the "metadata" POJO field.

jyemin: I don't think what you're saying is correct.  It _is_ encoded, and either insertOne or replaceOne will write the context of the field annotated with BsonExtraElements into the top-level document.  

We can touch base tomorrow to clarify

source/fundamentals/data-formats/pojo-customization.txt
ccho-mongodb: Based on our conversation, I'm planning to change the situation to talk about a situation in which you have a partially-structured schema.

I'd like any input you have on which example you think would be most suitable:

1) Once the BsonExtraElements field is established in the POJO, update the document and application's usage of that data without having to change the serial version id. E.g. Add a "hidden: true" field/value to the document to instruct the application from surfacing that POJO from an API call.

2) Lazy data migration: the goal is to get all the data to resemble the new schema reflected in the POJO, but it's updated only when you access the document rather than in bulk. Not common, but have encountered this situation before.

3) Overload application models to contain related metadata that other processes generate (such as search weights/relevance and ML data). Doesn't provide a very strong case for using the field unless the application needed to surface that data.

ccho-mongodb:
@jyemin I fixed the description and updated the example, please take a look when you get the chance.

source/fundamentals/data-formats/pojo-customization.txt
```diff
@@ -300,14 +300,18 @@ package:
...
-     - Specifies the POJO field on which to deserialize any elements that are
+     - Specifies the POJO field on which to deserialize all elements that are
        not mapped to a field. The POJO field must be one of the following
        types:
 
        - `Document <{+api+}/apidocs/bson/org/bson/Document.html>`__
        - `BsonDocument <{+api+}/apidocs/bson/org/bson/BsonDocument.html>`__
        - Map<String, Object>
 
+       When serializing a POJO through an insert or replace operation, the
```
jyemin, line 311: I don't think this is true.  Rather, I think each field in the Map is stored at the same level as the other fields in the POJO.

source/fundamentals/data-formats/pojo-customization.txt
```diff
@@ -382,45 +386,57 @@ BsonExtraElements Example
...
-.. code-block:: none
+.. code-block:: json
+   :copyable: false
+   :emphasize-lines: 5-9
+
+   {
+     "name": "MDB0123",
+     "serialNumber": "62e2...",
+     "relatedItems": null,
+     "additionalInfo": {
```
jyemin, line 433: This is not how it's supposed to work.  Rather, it should be:

```
   {
     "name": "MDB0123",
     "serialNumber": "62e2...",
     "relatedItems": null,
     "dimensions": "3x4x5",
     "weight": "256g"
   }
```

See [this test case](https://github.com/mongodb/mongo-java-driver/blob/95cb47a61f2896044e69ab73d59306e1527686e2/bson/src/test/unit/org/bson/codecs/pojo/PojoRoundTripTest.java#L528-L531) in the driver.

Do you have an example application that shows it working this way?

jyemin (changes requested):

source/fundamentals/data-formats/pojo-customization.txt
ccho-mongodb: Sorry, I was focused on updating the other part and forgot to update this mention.

source/fundamentals/data-formats/pojo-customization.txt
ccho-mongodb: Sorry, I think the disconnect is that I was portraying what was in the POJO, not MongoDB since the latter is redundant. I'll show the output using a generic toString() method to make it more apparent since that would show the Java class names.

ccho-mongodb: E.g.
```
ProductWithBsonExtraElements [name=MDB0123, serialNumber=62e2b9c5c32cb23e8cb0c95e, relatedItems=null, additionalInfo=Document{{dimensions=5x5x5}}]
```

ccho-mongodb:
Hey @jyemin, I made updates to remove the content in the annotation table and updated the description and output to make it more clear what the POJO contains. I also added my sample code for testing purposes only: https://github.com/mongodb/docs-java/pull/242/files#diff-295533015998e6eb3072ac03e18e8fe140c7c2fc29b6b6592b82fb87bb2549d5

source/includes/fundamentals/code-snippets/bson-extra-elements/BsonExtraElementsExample.java
```diff
@@ -0,0 +1,74 @@
...
+import com.mongodb.client.model.Updates;
+import com.mongodb.client.result.UpdateResult;
+
+public class BsonExtraElementsExample {
+
+	private static String DOC_NAME_VALUE = "MDB0123";
+
+    private static CodecRegistry getCodecRegistry() {
+        ClassModel<ProductWithBsonExtraElements> classModel = ClassModel.builder(ProductWithBsonExtraElements.class).
+                conventions(Arrays.asList(Conventions.ANNOTATION_CONVENTION)).build();
```
jyemin, line 30: I think this convention is enabled by default.  Any reason to include it here explicitly?  i.e. you can just do

    PojoCodecProvider.builder().automatic(true).build();

source/includes/fundamentals/code-snippets/bson-extra-elements/BsonExtraElementsExample.java
```diff
@@ -0,0 +1,74 @@
...
+import com.mongodb.client.MongoClients;
+import com.mongodb.client.MongoCollection;
+import com.mongodb.client.MongoDatabase;
+import com.mongodb.client.model.Filters;
+import com.mongodb.client.model.Updates;
+import com.mongodb.client.result.UpdateResult;
+
+public class BsonExtraElementsExample {
+
+	private static String DOC_NAME_VALUE = "MDB0123";
```
jyemin, line 26: Nit: fix indent

source/includes/fundamentals/code-snippets/bson-extra-elements/BsonExtraElementsExample.java
```diff
@@ -0,0 +1,74 @@
...
+        PojoCodecProvider pojoCodecProvider = PojoCodecProvider.builder().register(classModel).build();
+        return fromRegistries(getDefaultCodecRegistry(), fromProviders(pojoCodecProvider));
+
+    }
+
+    private static <T> void setup(MongoCollection<T> collection) {
+    	collection.drop();
+
+    	ProductWithBsonExtraElements doc = new ProductWithBsonExtraElements(DOC_NAME_VALUE);
+    	collection.insertOne((T) doc);
```
jyemin, line 40: If you do the above, this (unsafe) cast is redundant and should be removed.

source/includes/fundamentals/code-snippets/bson-extra-elements/BsonExtraElementsExample.java
```diff
@@ -0,0 +1,74 @@
...
+
+    private static CodecRegistry getCodecRegistry() {
+        ClassModel<ProductWithBsonExtraElements> classModel = ClassModel.builder(ProductWithBsonExtraElements.class).
+                conventions(Arrays.asList(Conventions.ANNOTATION_CONVENTION)).build();
+        PojoCodecProvider pojoCodecProvider = PojoCodecProvider.builder().register(classModel).build();
+        return fromRegistries(getDefaultCodecRegistry(), fromProviders(pojoCodecProvider));
+
+    }
+
+    private static <T> void setup(MongoCollection<T> collection) {
```
jyemin, line 36: Don't make this generic.  Instead just take an instance of `MongoCollection<ProductWithBsonExtraElements>`

source/includes/fundamentals/code-snippets/bson-extra-elements/BsonExtraElementsExample.java
```diff
@@ -0,0 +1,74 @@
...
+    	collection.drop();
+
+    	ProductWithBsonExtraElements doc = new ProductWithBsonExtraElements(DOC_NAME_VALUE);
+    	collection.insertOne((T) doc);
+    }
+
+    // Update without the POJO; match the annotations for field names
+    private static void addOtherFields(MongoClient client) {
+    	MongoCollection collection = client.getDatabase("sample_pojo").getCollection("ProductsWithMoreInfo");
```
jyemin, line 45: This should be of type `MongoCollection<Document>`

source/includes/fundamentals/code-snippets/bson-extra-elements/ProductWithBsonExtraElements.java
```diff
@@ -0,0 +1,75 @@
...
+    @BsonExtraElements
+    private Document additionalInfo;
+
+    @BsonCreator
+    public ProductWithBsonExtraElements(
+            @BsonProperty("modelName") String name) {
+        this.name = name;
+    }
+
+	public String getName() {
```
jyemin, line 37: The indents in this file should be made consistent

source/includes/fundamentals/code-snippets/bson-extra-elements/ProductWithBsonExtraElements.java
```diff
@@ -0,0 +1,75 @@
...
+public class ProductWithBsonExtraElements {
+
+    @BsonProperty("modelName")
+    private String name;
+
+    @BsonId()
+    @BsonRepresentation(BsonType.OBJECT_ID)
+    private String serialNumber;
+
+    @BsonIgnore
```
jyemin, line 25: Does this property serve any purpose in the example?  If not, it should be removed.

source/includes/fundamentals/code-snippets/bson-extra-elements/ProductWithBsonExtraElements.java
```diff
@@ -0,0 +1,75 @@
...
+import org.bson.Document;
+import org.bson.codecs.pojo.annotations.BsonCreator;
+import org.bson.codecs.pojo.annotations.BsonDiscriminator;
+import org.bson.codecs.pojo.annotations.BsonExtraElements;
+import org.bson.codecs.pojo.annotations.BsonId;
+import org.bson.codecs.pojo.annotations.BsonIgnore;
+import org.bson.codecs.pojo.annotations.BsonProperty;
+import org.bson.codecs.pojo.annotations.BsonRepresentation;
+
+@BsonDiscriminator(value="ProductWithBsonExtraElements", key="_cls")
```
jyemin, line 15: Is this annotation necessary the example?  If not, it should be removed

source/includes/fundamentals/code-snippets/bson-extra-elements/BsonExtraElementsExample.java
jyemin, line 26: Nit: make it final

source/includes/fundamentals/code-snippets/bson-extra-elements/BsonExtraElementsExample.java
```diff
@@ -0,0 +1,74 @@
...
+    			Filters.eq("modelName", DOC_NAME_VALUE),
+    			Updates.combine(
+    					Updates.set("dimensions", "3x4x5"),
+    					Updates.set("weight", "256g")));
+    	System.out.println(result);
+
+    }
+
+    private static <T> void printDocuments(MongoCollection<T> collection) {
+    	 List<T> products = new ArrayList<T>();
```
jyemin, line 57: Can be shortened to just:

    collection.find().into(new ArrayList<>()).forEach(System.out::println);

source/fundamentals/data-formats/pojo-customization.txt
```diff
@@ -419,24 +415,19 @@ code example:
 
 If a database client updated the documents in the collection to include
 additional fields that describe the products such as ``dimensions`` and
-``weight``, a document retrieved using the ``Product`` POJO might resemble
-the following JSON object:
+``weight``, the document retrieved using the ``Product`` POJO might resemble
```
jyemin, line 418: I still feel like it would aid clarity if you show the JSON exactly instead of just describing it.  The way it is, there is room for misinterpretation.

jyemin (changes requested):

source/includes/fundamentals/code-snippets/bson-extra-elements/BsonExtraElementsExample.java
ccho-mongodb: I did this intentionally so other writers could use this class as a model for creating other small applications to test their code examples. Are there any disadvantages in using generic types for this use case?

source/includes/fundamentals/code-snippets/bson-extra-elements/ProductWithBsonExtraElements.java
ccho-mongodb: The context of the example mentions an addition to the ProductPojo, so keeping everything else was implied. It's not strictly necessary, but would likely make more sense to the next writer.

source/includes/fundamentals/code-snippets/bson-extra-elements/ProductWithBsonExtraElements.java
ccho-mongodb: Similar reason as mentioned in this comment https://github.com/mongodb/docs-java/pull/242#discussion_r939022903

source/includes/fundamentals/code-snippets/bson-extra-elements/BsonExtraElementsExample.java
ccho-mongodb: Is there any disadvantage to keeping it as is? I was thinking that for writer testing purposes, keeping it as a lambda could allow them to insert debugging/testing logic more easily.

source/includes/fundamentals/code-snippets/bson-extra-elements/BsonExtraElementsExample.java
jyemin: Yes, because the cast to `T` is unsafe and generates a warning.

source/includes/fundamentals/code-snippets/bson-extra-elements/BsonExtraElementsExample.java
jyemin: Just less idiomatic Java but it's ok.

ccho-mongodb:
@jyemin Thanks for all the feedback, ready for another review.

source/includes/fundamentals/code-snippets/bson-extra-elements/ProductWithBsonExtraElements.java
jyemin: The indentation is still inconsistent.

source/includes/fundamentals/code-snippets/bson-extra-elements/BsonExtraElementsExample.java
```diff
@@ -0,0 +1,70 @@
...
+import com.mongodb.client.result.UpdateResult;
+
+public class BsonExtraElementsExample {
+
+	private static final String DOC_NAME_VALUE = "MDB0123";
+
+	private static CodecRegistry getCodecRegistry() {
+		PojoCodecProvider pojoCodecProvider = PojoCodecProvider.builder().automatic(true).build();
+		return fromRegistries(getDefaultCodecRegistry(), fromProviders(pojoCodecProvider));
+
```
jyemin, line 29: Nit: extra whitespace

source/includes/fundamentals/code-snippets/bson-extra-elements/BsonExtraElementsExample.java
```diff
@@ -0,0 +1,70 @@
...
+import com.mongodb.client.model.Updates;
+import com.mongodb.client.result.UpdateResult;
+
+public class BsonExtraElementsExample {
+
+	private static final String DOC_NAME_VALUE = "MDB0123";
+
+	private static CodecRegistry getCodecRegistry() {
+		PojoCodecProvider pojoCodecProvider = PojoCodecProvider.builder().automatic(true).build();
+		return fromRegistries(getDefaultCodecRegistry(), fromProviders(pojoCodecProvider));
```
jyemin, line 28: Just FYI, this can be simplified to:

    return fromProviders(getDefaultCodecRegistry(), pojoCodecProvider);

since a registry is also a provider

source/includes/fundamentals/code-snippets/bson-extra-elements/BsonExtraElementsExample.java
```diff
@@ -0,0 +1,70 @@
...
+import com.mongodb.client.MongoClients;
+import com.mongodb.client.MongoCollection;
+import com.mongodb.client.MongoDatabase;
+import com.mongodb.client.model.Filters;
+import com.mongodb.client.model.Updates;
+import com.mongodb.client.result.UpdateResult;
+
+public class BsonExtraElementsExample {
+
+	private static final String DOC_NAME_VALUE = "MDB0123";
```
jyemin, line 24: Have we standardized on 8 space indents?  4 is more typical.

jyemin (approved):
Just whitespace issues.  LGTM!

source/includes/fundamentals/code-snippets/bson-extra-elements/BsonExtraElementsExample.java
ccho-mongodb: I think it's just a settings issue with my IDE; haven't had time to set the formatting. Will convert tabs to spaces and use 4 space indents.

source/includes/fundamentals/code-snippets/bson-extra-elements/BsonExtraElementsExample.java
ccho-mongodb: > The indentation is still inconsistent.

Sorry, strange conversion of tabs and spaces makes it look ok in the IDE. Need to adjust my IDE settings as mentioned in https://github.com/mongodb/docs-java/pull/242#discussion_r940305061

PR #263, opened 2022-08-17
Add workflow action - copy compat tables to docs-shared

mongoKart:
Environment secret has already been added to docs-java.


terakilobyte (approved):
LGTM. Will there be a follow on task to update the iconography and bring in the legend?

mongoKart:
> LGTM. Will there be a follow on task to update the iconography and bring in the legend?

Completed as part of this PR.

PR #265, opened 2022-08-18
allow copy workflow to trigger manually

mongoKart:
Add option to trigger copy to docs-shared manually instead of on PR


terakilobyte (approved):
lgtm, let's give it a shot

PR #289, opened 2022-10-10
(DOCSP-25230) Min Server Version 3.4 (min wire version 6)

biniona-mongodb:
# Pull Request Info

Add information about new min server/wire version.

[JIRA](https://jira.mongodb.org/browse/DOCSP-25230)
Staging - https://docs-mongodbcom-staging.corp.mongodb.com/drivers/docsworker-xlarge/NNNNN/

## Self-Review Checklist

- [ ] Is this free of any warnings or errors in the RST?

### DIFF
--- a/source/upgrade.txt
+++ b/source/upgrade.txt
@@ -50,6 +50,19 @@ changes between the current and upgrade versions. For example, if you
 are upgrading the driver from v4.0 to v4.5, address all breaking changes from
 the version after v4.0 including any listed under v4.5.
 
+.. _java-breaking-changes-v4.8:
+
+Version 4.8 Breaking Changes
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+- The v4.8 driver drops support for MongoDB v3.4 and earlier. To use
+  the v4.8 driver, your MongoDB server must be v3.6 or later. To learn
+  how to upgrade your MongoDB server to v3.6, see the following resources:
+
+  - :ref:`<3.6-upgrade-replica-set>`
+  - :ref:`<3.6-upgrade-standalone>`
+  - :ref:`<3.6-upgrade-sharded-cluster>`
+
 .. _java-breaking-changes-v4.7:
 
 Version 4.7 Breaking Changes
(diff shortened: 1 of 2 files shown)


jordan-smith721 (approved):
LGTM

jyemin (approved):

source/upgrade.txt
```diff
@@ -50,6 +50,20 @@ changes between the current and upgrade versions. For example, if you
 are upgrading the driver from v4.0 to v4.5, address all breaking changes from
 the version after v4.0 including any listed under v4.5.
 
+.. _java-breaking-changes-v4.8:
+
+Version 4.8 Breaking Changes
```
jyemin, line 55: Though now that I think of it, do we want to call it a breaking change?  It makes it seem like a semantic versioning violation to express it this way. I would prefer something different, though I can't think of quite the right terminology.  Maybe just be more explicit: 

> Version 4.8 Drops Support for EOL server releases

jyemin (changes requested):

source/upgrade.txt
terakilobyte: Should we update the compat table to remove a checkmark next to all versions < 4.2

jyemin: < 3.4 ?

terakilobyte: If we say eol server versions, that's everything < 4.2

source/upgrade.txt
```diff
@@ -35,6 +35,34 @@ Before you upgrade, perform the following actions:
...
+.._java-server-release-changes:
+
+Server Release Compatibility Changes
+------------------------------------
+
+A server release compatibility change is a modification
+to the {+driver-long+} that discontinues support for a set
+of **end-of-life** MongoDB server versions. An end-of-life MongoDB
+server version is a version of the MongoDB database that is no
+longer supported by MongoDB the company.
```
terakilobyte, line 47:
```suggestion
of MongoDB server versions.
```

jyemin (approved):
One grammar nit, but otherwise LGTM

PR #294, opened 2022-11-10
DOCSP-24606: add info about v4.8 server breaking change and upgrading

ccho-mongodb:
# Pull Request Info

JIRA - https://jira.mongodb.org/browse/DOCSP-24606

Moved Server Release Compatibility changes to the bottom since the driver breaking changes are more common. Made sure that there's a link from the driver version breaking change entry to the server compatibility changes.

Staging:

[What's new](https://docs-mongodbcom-staging.corp.mongodb.com/java/docsworker-xlarge/DOCSP-24606-3.6-unsupported/whats-new/#what-s-new-in-4.8)
[v4.8 Breaking changes](https://docs-mongodbcom-staging.corp.mongodb.com/java/docsworker-xlarge/DOCSP-24606-3.6-unsupported/upgrade/#version-4.8-breaking-changes)

## Self-Review Checklist

### DIFF
--- a/source/upgrade.txt
+++ b/source/upgrade.txt
@@ -76,6 +76,14 @@ changes between the current and upgrade versions. For example, if you
 are upgrading the driver from v4.0 to v4.5, address all breaking changes from
 the version after v4.0 including any listed under v4.5.
 
+.. _java-breaking-changes-v4.8:
+
+Version 4.8 Breaking Changes
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+- The driver ends support for connecting to MongoDB Server versions v3.4 and
+  earlier. To learn more about this change, see the :ref:`<java-server-release-change-v4.8>`
+  section.
 
 .. _java-breaking-changes-v4.7:
 
--- a/source/upgrade.txt
+++ b/source/upgrade.txt
@@ -27,39 +27,15 @@ Before you upgrade, perform the following actions:
   page for this information.
 - Address any breaking changes between the current version of the driver
   your application is using and your planned upgrade version in the
-  section on :ref:`Breaking Changes <java-breaking-changes>` below.
+  section on :ref:`Breaking Changes <java-breaking-changes>` below. To learn
+  more about the MongoDB server release compatibility changes, see the
+  section on :ref:`<java-server-release-changes>`.
 
 .. tip::
 
    To minimize the amount of changes your application may require when
    upgrading driver versions in the future, use the
    :ref:`{+stable-api+} <stable-api-java>`.
-
-.. _java-server-release-changes:
-
-Server Release Compatibility Changes
-------------------------------------
-
-A server release compatibility change is a modification
-to the {+driver-long+} that discontinues support for a set of
-MongoDB server versions.
-
-To learn more about end-of-life MongoDB products,
-see the `Legacy Support Policy <https://www.mongodb.com/support-policy/legacy>`__.
-
-.. _java-server-release-change-v4.8:
-
-Version 4.8 Server Release Support Changes
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-
-- The v4.8 driver drops support for MongoDB v3.4 and earlier.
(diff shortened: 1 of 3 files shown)


source/upgrade.txt
```diff
@@ -27,40 +27,16 @@ Before you upgrade, perform the following actions:
   page for this information.
 - Address any breaking changes between the current version of the driver
   your application is using and your planned upgrade version in the
-  section on :ref:`Breaking Changes <java-breaking-changes>` below.
+  section on :ref:`Breaking Changes <java-breaking-changes>` below. To learn
+  more about the MongoDB server release compatibility changes, see the
+  section on :ref:`<java-server-release-changes>`.
```
mongoKart, line 32: Issue: Per [style guide](https://www.mongodb.com/docs/meta/style-guide/terminology/alphabetical-terms/#term-about), don't use "on" to mean "about."

Suggestion: 
```suggestion
  more about the MongoDB server release compatibility changes, see :ref:`<java-server-release-changes>`.
```

source/upgrade.txt
```diff
@@ -76,6 +52,14 @@ changes between the current and upgrade versions. For example, if you
 are upgrading the driver from v4.0 to v4.5, address all breaking changes from
 the version after v4.0 including any listed under v4.5.
 
+.. _java-breaking-changes-v4.8:
+
+Version 4.8 Breaking Changes
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+- The driver ends support for connecting to MongoDB Server versions v3.4 and
```
mongoKart, line 60: Issue: AFAICT, this is the only place on the page that capitalizes the S in "MongoDB Server."

Suggestion: Standardize capitalization across the page.

source/upgrade.txt
```diff
@@ -76,6 +52,14 @@ changes between the current and upgrade versions. For example, if you
...
+  earlier. To learn more about this change, see the :ref:`<java-server-release-change-v4.8>`
+  section.
```
mongoKart, line 62: Optional: At the end of this page, you just say, "see <section>" instead of "see the <section> section." Maybe a little cleaner to do that here too.

mongoKart (changes requested):
Content and structure look good. A couple quick nitpicks. LMK if you think I should have approved this PR and made the changes optional instead of requesting changes.

source/upgrade.txt
ccho-mongodb: Good catch. I'll change it to "java-server-release-changes section". When linking content, I prefer to identify the type so that the reader has some idea of what they're going to be looking at or for (guide, tutorial, page, section, etc.).

source/upgrade.txt
ccho-mongodb: Good catch. I'll change this instance. And standardizing beyond that such as across the docs set is definitely a worthwhile task, so I'll ticket that.

source/upgrade.txt
ccho-mongodb: I have a preference similar to described in this previous response. I think it helps the reader identify what the UX will be, and which heading to look under if it doesn't happen to target the exact position on their browser due to how the page gets rendered.

For context, this was something the team had also agreed was a good idea a while back, but open to reasons why we shouldn't do it and changing that. If we decide to discontinue this, we should also ticket the work to standardize across docs.

ccho-mongodb:
> Content and structure look good. A couple quick nitpicks. LMK if you think I should have approved this PR and made the changes optional instead of requesting changes.

If there's an issue, it's definitely a candidate, but up to you. There's some efficiency in not requiring a re-review in that you might be blocking it if you're not able to re-review quickly, but maybe you'll identify something new from the edits or existing content.

source/upgrade.txt
ccho-mongodb: For further clarification, I think "MongoDB Server" is the correct product name, and MongoDB server could be referencing "a server/process that happens to be MongoDB Server". I don't know that our Server Manual uses this consistently, but will do some additional investigation.

mongoKart (approved):
lgtm!

PR #299, opened 2022-11-16
DOCSP-25900: recursive type support

rustagir:
# Pull Request Info

JIRA - https://jira.mongodb.org/browse/DOCSP-25900
Staging - https://docs-mongodbcom-staging.corp.mongodb.com/java/docsworker-xlarge/DOCSP-25900-recursive-type-support/fundamentals/data-formats/document-data-format-record/#operations-with-recursive-types

## Self-Review Checklist

### DIFF
--- a/source/fundamentals/data-formats/document-data-format-record.txt
+++ b/source/fundamentals/data-formats/document-data-format-record.txt
@@ -199,8 +199,29 @@ provides an example of a recursive class design:
 
 .. code-block:: java
 
-   public record RecordTree<T>(
-           T content,
-           RecordTree<T> left,
-           RecordTree<T> right
+   public record RecordTree(
+           String content,
+           RecordTree left,
+           RecordTree right
    ) {}
+
+You can perform read and write operations on recursive records the same
+way you would for other records. The following code shows how you can
+execute a find operation on a collection of cyclical ``RecordTree`` types:
+
+.. io-code-block::
+   :copyable: true
+
+   .. input::
+      :language: java
+
+      MongoDatabase database = mongoClient.getDatabase("myDB");
+      MongoCollection<RecordTree> collection = database.getCollection("myCollection", RecordTree.class);
+      
+      Bson filter = Filters.eq("left.left.right.content", "Ikatere");
+      collection.find(filter).forEach(doc -> System.out.println(doc));
+
+   .. output::
+      :language: sh
+
+      RecordTree[content=Ranginui, left=RecordTree[content=..., left=RecordTree[content=..., right=RecordTree[content=Ikatere...]]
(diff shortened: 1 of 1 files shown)


source/fundamentals/data-formats/document-data-format-record.txt
```diff
@@ -191,3 +190,38 @@ as shown in the following code:
 
       NetworkDeviceRecord[deviceId=fedc..., name=Enterprise Wi-fi, deviceType=router]
 
+Operations with Recursive Types
+-------------------------------
+
+The driver natively supports encoding and decoding cyclical
+references without causing runtime recursion. The following code
+provides an example of a recursive class design:
```
mongoKart, line 198: Issue: My Java's a little rusty, but isn't this a method named RecordTree that returns a record? If not, please ignore :)

Suggestion: Change code sample to a recursive class.

mongoKart (changes requested):
One question. Also double-checking you're sending to technical review at some point so I don't have to scrutinize code samples

source/fundamentals/data-formats/document-data-format-record.txt
```diff
@@ -191,3 +190,38 @@ as shown in the following code:
 
       NetworkDeviceRecord[deviceId=fedc..., name=Enterprise Wi-fi, deviceType=router]
 
+Operations with Recursive Types
+-------------------------------
+
+The driver natively supports encoding and decoding cyclical
+references without causing runtime recursion. The following code
```
jyemin, line 197: It's not the references that are cyclical.  The driver doesn't actually handle that.  It's when there are cycles in the type definitions.  In this case, RecordTree has two components of type RecordTree.  So there is a (short) cycle in the type definitions.  But the support generalizes to any sort of cycle in the type system, e.g.  RecordOne with a component of type RecordTwo with a component of type RecordThree with a component of type RecordOne.

jyemin (changes requested):

source/fundamentals/data-formats/document-data-format-record.txt
rustagir: do you think that I should expand the example to add a second (or third) record definition to fully illustrate support for cycles? I will also reword.

jyemin: Probably not necessary, I was just trying to explain it to you.  But I don't think we should use the phrase "cyclical references" as that will most likely be interpreted as references of instances, not types.

source/fundamentals/data-formats/document-data-format-record.txt
jyemin: This feature is about records, not classes, so I think it should stay as is.

jyemin (approved):
LGTM

mongoKart (approved):
LGTM!

PR #330, opened 2023-02-15
DOCSP-28004: boolean operators

rustagir:
# Pull Request Info

JIRA - https://jira.mongodb.org/browse/DOCSP-28004
Staging - https://docs-mongodbcom-staging.corp.mongodb.com/java/docsworker-xlarge/DOCSP-28004-boolean-operators/fundamentals/aggregation/aggregation-expression-operations/#boolean-expressions

## Self-Review Checklist


source/fundamentals/aggregation/aggregation-expression-operations.txt
```diff
@@ -25,8 +25,109 @@ more naturally readable.
...
+         :emphasize-lines: 5-6
+
+         var pipeline = Arrays.asList(project(
+                 fields(
+                         exclude("_id"),
+                         include("location"),
+                         computed("extremeTemp",
+                                 current().getInteger("temp").lt(of(10)).or(current().getInteger("temp").gt(of(90))))
+                 )
+         ));
```
katcharov, line 104:
```suggestion
         var pipeline = Arrays.asList(project(fields(
                 exclude("_id"),
                 include("location"),
                 computed("extremeTemp",
                         current().getInteger("temp").lt(of(10))
                             .or(current().getInteger("temp").gt(of(90))))
         )));
```
The example scrolls horizontally - the above indentation should make it more readable.

You might also consider including 
```
                             .cond("extreme!", "not extreme")
```
on a new line after the "or" (with correct parens). The way the "cond" operation works (along with switch, and a few others) might be unexpected to some users, so it seems worth addressing it in some way.

source/fundamentals/aggregation/aggregation-expression-operations.txt
```diff
@@ -25,8 +25,109 @@ more naturally readable.
...
+         :language: none
+         :visible: false
+      
+         {"location": "New York, NY", "extremeTemp": false}
+         {"location": "Duluth, MN", "extremeTemp": true}
+         {"location": "Riverside, CA", "extremeTemp": true}
+         {"location": "Tucson, AZ", "extremeTemp": true}
+         
+   The following snippet provides the equivalent expression in
+   the Query API:
```
katcharov, line 118: Perhaps we should make clear somewhere that the API makes no guarantee about what MQL might be generated.

katcharov (changes requested):

source/fundamentals/aggregation/aggregation-expression-operations.txt
```diff
@@ -28,5 +28,76 @@ more naturally readable.
...
+.. code-block:: java
+   :copyable: true
+
+   var pipeline = Arrays.asList(project(fields(
+           computed("result",
+                   current().getInteger("field1").lt(of(0))
+                       .or(current().getInteger("field1").gt(of(10))))
+   )));
+      
+The following snippet provides an equivalent expression in
```
ccho-mongodb, line 89: Suggestion:
I wasn't able to tell if there's anything that identifies as a "snippet" or "code", but it could be good just to stay consistent if there's no description of each term.

source/fundamentals/aggregation/aggregation-expression-operations.txt
```diff
@@ -28,5 +28,76 @@ more naturally readable.
...
+   corresponding Query API expression, they may differ in expected
+   parameters and implementation.
+   
+The following code computes the ``result`` field based on if the value
+of ``field1`` is less than ``0`` *or* greater than ``10``:
+
+.. code-block:: java
+   :copyable: true
+
+   var pipeline = Arrays.asList(project(fields(
```
ccho-mongodb, line 83: Suggestion:
I think it would be helpful to show the static imports used in this code example.

source/fundamentals/aggregation/aggregation-expression-operations.txt
```diff
@@ -28,5 +28,76 @@ more naturally readable.
...
+     - :manual:`$cond </reference/operator/aggregation/cond/>`
+
+.. note::
+
+   While each Java method is functionally equivalent to the
+   corresponding Query API expression, they may differ in expected
+   parameters and implementation.
+   
+The following code computes the ``result`` field based on if the value
+of ``field1`` is less than ``0`` *or* greater than ``10``:
```
ccho-mongodb, line 78: Suggestion:
I see this is similar to https://www.mongodb.com/docs/manual/reference/operator/aggregation/or/#example but not the same. However, I wonder if there might be conditions more familiar to a reader which could aid in understanding it more quickly.
Also, it could be useful to identify which operator this shows.

E.g. 
"The following example shows how to use the or() operator to ..."

ccho-mongodb (changes requested):
Good start! A couple issues and suggestions.

ccho-mongodb (changes requested):
Good start! A couple issues and suggestions.

source/fundamentals/aggregation/aggregation-expression-operations.txt
```diff
@@ -28,5 +28,76 @@ more naturally readable.
...
+
+   The Query API expression that the driver generates from each
+   aggregation expression operation might not be
+   identical to the Query API expression provided in each example.
+   However, they are functionally equivalent.
+
+Boolean Expressions
+~~~~~~~~~~~~~~~~~~~
+
+You can invoke boolean operations on boolean values of type ``MqlBoolean``.
```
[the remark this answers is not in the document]
rustagir: I think the first clause and the second sentence are somewhat redundant, so I combined them.

source/fundamentals/aggregation/aggregation-expression-operations.txt
```diff
@@ -28,5 +28,76 @@ more naturally readable.
...
+Each of the following sections provides information and examples for
+aggregation expression operations available in the {+driver-short+}.
+This guide sorts these operations into categories based on general
+purpose and functionality.
+
+.. important::
+
+   The Query API expression that the driver generates from each
+   aggregation expression operation might not be
+   identical to the Query API expression provided in each example.
```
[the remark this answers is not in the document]
rustagir: Changed for style

source/fundamentals/aggregation/aggregation-expression-operations.txt
```diff
@@ -28,5 +28,76 @@ more naturally readable.
...
+   However, they are functionally equivalent.
+
+Boolean Expressions
+~~~~~~~~~~~~~~~~~~~
+
+You can invoke boolean operations on boolean values of type ``MqlBoolean``.
+
+The following table describes the boolean aggregation operation methods
+available in the {+driver-short+} and the equivalent expressions in the
```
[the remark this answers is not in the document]
rustagir: removed
or changed

source/fundamentals/aggregation/aggregation-expression-operations.txt
```diff
@@ -28,5 +28,76 @@ more naturally readable.
...
+Query API:
+
+.. list-table::
+   :header-rows: 1
+   :widths: 40 60
+
+   * - Java Aggregation Expression
```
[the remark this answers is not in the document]
rustagir: changed to just Aggregation Method

source/fundamentals/aggregation/aggregation-expression-operations.txt
```diff
@@ -28,5 +28,76 @@ more naturally readable.
...
+   However, they are functionally equivalent.
+
+Boolean Expressions
+~~~~~~~~~~~~~~~~~~~
+
+You can invoke boolean operations on boolean values of type ``MqlBoolean``.
+
+The following table describes the boolean aggregation operation methods
+available in the {+driver-short+} and the equivalent expressions in the
+Query API:
```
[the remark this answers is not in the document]
rustagir: None of these methods have multiple signatures, but I can still call out that the driver includes multiple methods of the same name for some of these methods.

source/fundamentals/aggregation/aggregation-expression-operations.txt
rustagir: How can I show this without showing the whole code structure including the main method etc.?

source/fundamentals/aggregation/aggregation-expression-operations.txt
rustagir: Would it be familiar to users to refer to a `qty` field with no introduction about what the document schema is?

source/fundamentals/aggregation/aggregation-expression-operations.txt
ccho-mongodb: Not recommending this exact format, but here's one example in the current Java docs:
https://www.mongodb.com/docs/drivers/java/sync/current/fundamentals/crud/read-operations/sort/#methods-for-sorting

source/fundamentals/aggregation/aggregation-expression-operations.txt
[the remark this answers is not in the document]
ccho-mongodb: The current naming (in the Manual) is confusing to me, so not sure I can point you to the right option.

However, I wonder if Aggregation Method might generate confusion since that term is used to categorize "aggregate()" and "mapReduce()" in the Manual here: https://www.mongodb.com/docs/v6.0/reference/operator/aggregation/interface/#aggregation-methods

On the other hand, those are also called Aggregation Commands here:
https://www.mongodb.com/docs/v6.0/reference/aggregation-commands-comparison/#aggregation-commands-comparison-table

Since this page refers to them as "Aggregation Expression Operators", maybe that would be the best choice? Or qualify them with "Java", e.g. "Java Aggregation Expression Operators".

source/fundamentals/aggregation/aggregation-expression-operations.txt
ccho-mongodb: I wasn't specifically commenting on the field names, but I think using a more recognizable situation and field names might help. The current example is "Add a field called 'result' with the value of field1 if it's less than "0" or greater than "10". As far as I know, there isn't any natural association one might make as to why you would want to do that.

Not the most imaginative examples, but hopefully demonstrate what I mean:
1) if "name" equals "mouse" or "keyboard", set the value of the "type" field to "computer peripheral".
2) if "body_temperature" is less than "96" or "body_temperature" is greater than "100", set the value of the "action" to "see the doctor".

ccho-mongodb: Looks like you had something along the lines of what I was looking for in a prior revision about extreme temperatures which could still work. Was there any specific reason why you changed it other than that maybe it contained more than just the "or" operator?

source/fundamentals/aggregation/aggregation-expression-operations.txt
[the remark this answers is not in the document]
ccho-mongodb: > None of these methods have multiple signatures, but I can still call out that the driver includes multiple methods of the same name for some of these methods.

Cool, then no need to mention that if you're linking to the API docs since that should disambiguate them.

source/fundamentals/aggregation/aggregation-expression-operators.txt
```diff
@@ -0,0 +1,32 @@
+=================================
```
ccho-mongodb, line 1: Question: 

Did you mean to commit this file? It looks like mostly a duplicate of [aggregation-expression-operations.txt](https://github.com/mongodb/docs-java/pull/330/files#diff-3efcbcb27373cfe178572891a6efd03125e78706f1a6232fa904ec73df71ff5f)

Currently, it's generating a warning because it uses the driver-short constant.

source/fundamentals/aggregation/aggregation-expression-operations.txt
ccho-mongodb: Another example:
https://www.mongodb.com/docs/drivers/java/sync/current/fundamentals/builders/aggregates/

source/fundamentals/aggregation/aggregation-expression-operations.txt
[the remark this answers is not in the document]
rustagir: chose "Java Aggregation Expression Operator"

source/fundamentals/aggregation/aggregation-expression-operations.txt
rustagir: Choosing the model on [this page](https://www.mongodb.com/docs/drivers/java/sync/current/fundamentals/crud/read-operations/sort/#methods-for-sorting), since I'm not sure we should direct users to include imports that will be unused, depending on their code.

source/fundamentals/aggregation/aggregation-expression-operations.txt
rustagir: I just thought that following the discussion Mike, you, and I had, that we had agreed on generic field names. I have changed the example to something easily understood.

source/fundamentals/aggregation/aggregation-expression-operators.txt
rustagir: deleted this file. I think this issue emerged when i merged master into this branch

source/fundamentals/aggregation/aggregation-expression-operations.txt
ccho-mongodb: Closing out since we discussed it synchronously and decided each example can either be:
1) generic field names when it's possible to clearly demonstrate/communicate its purpose without explanation
or
2) Provide a short explanation of what it does to help clarify what the aggregation does

ccho-mongodb (approved):
LGTM

source/fundamentals/aggregation/aggregation-expression-operations.txt
```diff
@@ -25,8 +25,109 @@ more naturally readable.
...
+
 Operations
 ----------
 
+Boolean Expressions
+~~~~~~~~~~~~~~~~~~~
+
+Boolean expressions evaluate arguments as booleans and return ``true`` or
+``false``. These expressions evaluate ``false``, ``null``, ``0``, and
+``undefined`` values as ``false``, and all other values as ``true``.
```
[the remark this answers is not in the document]
rustagir: Does the current wording of the intro text in this section address these points?

source/fundamentals/aggregation/aggregation-expression-operations.txt
```diff
@@ -25,8 +25,109 @@ more naturally readable.
...
+     - Query API Expression
+
+   * - ``and()``
+     - :manual:`$and </reference/operator/aggregation/and/>`
+       
+   * - ``not()``
+     - :manual:`$not </reference/operator/aggregation/not/>`
+
+   * - ``or()``
+     - :manual:`$or </reference/operator/aggregation/or/>`
```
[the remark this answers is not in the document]
rustagir: throughout our documentation, we refer to methods with a `()` marker, irrespective of the # of parameters they take. This format distinguishes methods from class references, etc. We think it is confusing for users to see placeholders such as `a` and `b`.

source/fundamentals/aggregation/aggregation-expression-operations.txt
```diff
@@ -25,8 +25,109 @@ more naturally readable.
...
+
+.. list-table::
+   :header-rows: 1
+   :widths: 40 60
+
+   * - Java Aggregation Expression
+     - Query API Expression
+
+   * - ``and()``
+     - :manual:`$and </reference/operator/aggregation/and/>`
```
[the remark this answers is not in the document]
rustagir: See above comment about full expressions.
Addressed the second comment with a note.

source/fundamentals/aggregation/aggregation-expression-operations.txt
rustagir: We are trying to use only one aggregation method per example, so that users can clearly see what a given method does individually.

source/fundamentals/aggregation/aggregation-expression-operations.txt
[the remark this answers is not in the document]
katcharov: It looks like these are covered as far as the Boolean section goes.

source/fundamentals/aggregation/aggregation-expression-operations.txt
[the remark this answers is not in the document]
katcharov: Fair enough.

I think users must be ok with placeholders to some extent, since they are used prominently in the [expressions docs](https://www.mongodb.com/docs/manual/reference/operator/aggregation/or/), but I can see how they might also be too much here.

source/fundamentals/aggregation/aggregation-expression-operations.txt
katcharov: Tough call. Out of the dozens of expressions, there were several comments during design review mentioning that `.cond` seemed unusual. It would be nice to have two examples, but if I had to pick I would go for $cond.

PR #334, opened 2023-02-23
Docsp 27977 - documents stage

mongoKart:
# Pull Request Info

JIRA - https://jira.mongodb.org/browse/DOCSP-27977
Staging - https://docs-mongodbcom-staging.corp.mongodb.com/java/docsworker-xlarge/docsp-27977-documents-stage/fundamentals/builders/aggregates/

## Self-Review Checklist

- [X] Are all the links working?

### DIFF
--- a/source/fundamentals/builders/aggregates.txt
+++ b/source/fundamentals/builders/aggregates.txt
@@ -2,8 +2,6 @@
 Aggregates Builders
 ===================
 
-.. default-domain:: mongodb
-
 .. contents:: On this page
    :local:
    :backlinks: none
@@ -50,6 +48,32 @@ aggregation as a list:
    :language: java
    :dedent:
 
+Documents
+---------
+
+Use the ``documents()`` method to create a
+:manual:`$documents </reference/operator/aggregation/documents/>`
+pipeline stage that returns literal documents from query values.
+
+.. important::
+
+   If you use the ``$documents`` stage in an aggregation pipeline, it must be the first 
+   stage in the pipeline.
+
+The following example creates a pipeline stage that creates 
+sample documents, each with a ``"title"`` field: 
+
+.. literalinclude:: /includes/fundamentals/code-snippets/builders/AggBuilders.java
+   :start-after: // begin documents
+   :end-before: // end documents
+   :language: java
+   :dedent:
+
+.. tip::
+
+   If you use the ``documents()`` method to provide the input to an aggregation pipeline, 
+   you can call the ``aggregate()`` method on a database object rather than on a 
+   collection.
 
 Match
 -----
@@ -71,7 +95,6 @@ The following example creates a pipeline stage that matches all documents where 
    :end-before: end match
    :language: java
    :dedent:
-
 
 
 Project
@@ -946,3 +969,4 @@ aggregation stage:
 
 Learn more about this helper from the
 `searchMeta() API documentation <{+api+}/apidocs/mongodb-driver-core/com/mongodb/client/model/Aggregates.html#searchMeta(com.mongodb.client.model.search.SearchCollector)>`__.
+
--- a/source/fundamentals/builders/aggregates.txt
+++ b/source/fundamentals/builders/aggregates.txt
@@ -48,32 +48,6 @@ aggregation as a list:
    :language: java
    :dedent:
 
-Documents
----------
-
-Use the ``documents()`` method to create a
-:manual:`$documents </reference/operator/aggregation/documents/>`
-pipeline stage that returns literal documents from input values.
-
-.. important::
-
-   If you use a ``$documents`` stage in an aggregation pipeline, it must be the first 
-   stage in the pipeline.
-
-The following example creates a pipeline stage that creates 
-sample documents with a ``title`` field: 
-
-.. literalinclude:: /includes/fundamentals/code-snippets/builders/AggBuilders.java
-   :start-after: // begin documents
-   :end-before: // end documents
-   :language: java
-   :dedent:
-
-.. tip::
-
-   If you use the ``documents()`` method to provide the input to an aggregation pipeline, 
-   you can call the ``aggregate()`` method on a database instead of on a 
-   collection.
 
 Match
 -----
@@ -131,6 +105,34 @@ into a new field called ``rating``, effectively renaming the field.
    :end-before: end computed
    :language: java
    :dedent:
+
+
+Documents
+---------
+
+Use the ``documents()`` method to create a
+:manual:`$documents </reference/operator/aggregation/documents/>`
+pipeline stage that returns literal documents from input values.
+
+.. important::
+
+   If you use a ``$documents`` stage in an aggregation pipeline, it must be the first 
+   stage in the pipeline.
+
+The following example creates a pipeline stage that creates 
+sample documents with a ``title`` field: 
+
+.. literalinclude:: /includes/fundamentals/code-snippets/builders/AggBuilders.java
+   :start-after: // begin documents
+   :end-before: // end documents
+   :language: java
+   :dedent:
+
+.. tip::
+
+   If you use the ``documents()`` method to provide the input to an aggregation pipeline, 
+   you can call the ``aggregate()`` method on a database instead of on a 
+   collection.
 
 
 Sample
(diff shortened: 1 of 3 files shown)


source/fundamentals/builders/aggregates.txt
```diff
@@ -50,6 +48,32 @@ aggregation as a list:
    :language: java
    :dedent:
 
+Documents
+---------
```
rustagir, line 52: Q: Is this new feature significant enough to be at the top of the page, even before match and project?

source/fundamentals/builders/aggregates.txt
```diff
@@ -50,6 +48,32 @@ aggregation as a list:
...
+
+Use the ``documents()`` method to create a
+:manual:`$documents </reference/operator/aggregation/documents/>`
+pipeline stage that returns literal documents from query values.
```
rustagir, line 56: Q: What is a "query value"? I noticed that the manual page refers to these as input values.

source/fundamentals/builders/aggregates.txt
```diff
@@ -50,6 +48,32 @@ aggregation as a list:
...
+
+.. important::
+
+   If you use the ``$documents`` stage in an aggregation pipeline, it must be the first 
+   stage in the pipeline.
```
rustagir, line 61:
```suggestion
   If you use a ``$documents`` stage in an aggregation pipeline, it must be the first 
   stage in the pipeline.
```

source/fundamentals/builders/aggregates.txt
```diff
@@ -50,6 +48,32 @@ aggregation as a list:
...
+
+The following example creates a pipeline stage that creates 
+sample documents with a ``"title"`` field:
```
rustagir, line 64: I: elsewhere on the page, field names are not in quotes

S:
```suggestion
sample documents with a ``title`` field: 
```

source/fundamentals/builders/aggregates.txt
```diff
@@ -50,6 +48,32 @@ aggregation as a list:
...
+.. literalinclude:: /includes/fundamentals/code-snippets/builders/AggBuilders.java
+   :start-after: // begin documents
+   :end-before: // end documents
+   :language: java
+   :dedent:
+
+.. tip::
+
+   If you use the ``documents()`` method to provide the input to an aggregation pipeline, 
+   you can call the ``aggregate()`` method on a database object rather than on a
```
rustagir, line 75:
```suggestion
   you can call the ``aggregate()`` method on a database instead of on a 
```

source/includes/fundamentals/code-snippets/builders/AggBuilders.java
```diff
@@ -302,4 +302,13 @@ private void matchStage() {
...
         // end match
     }
+
+    private void documents() {
+        // begin documents
+        documents(Arrays.asList(
+                new Document("title", "The Shawshank Redemption"),
+                new Document("title", "Back to the Future"),
+                new Document("title", "Jurassic Park")));
+        // end documents
```
rustagir, line 312: Q: Did you confirm that this is the correct syntax to create this stage?

rustagir: Just making sure -- it seems consistent with what is on this page, but doesn't exactly match the [tests from the feature PR](https://github.com/mongodb/mongo-java-driver/blob/a19fad2b3d93ee818132f14377e42cfaa268e8ee/driver-core/src/test/functional/com/mongodb/client/model/AggregatesTest.java#L170-L171).

rustagir (changes requested):
some small changes, but overall looks great!

source/fundamentals/builders/aggregates.txt
mongoKart: Good question. My reasoning was that when this stage it's used, it will be the first in the pipeline. But I'm happy to move it down if you don't find that reason compelling.

source/fundamentals/builders/aggregates.txt
mongoKart: That's...strange. I thought I copied the text directly from the manual. Changed.

source/fundamentals/builders/aggregates.txt
mongoKart: Good catch!

source/includes/fundamentals/code-snippets/builders/AggBuilders.java
mongoKart: I confirmed that it _works_, but not whether it's the preferred syntax. I was planning to send to tech review after this.

source/fundamentals/builders/aggregates.txt
rustagir: I think that match will likely still be the first stage for most users use cases... maybe this can go after project so it's near the top but not first. But up to you!

rustagir (approved):
lgtm with one suggestion that is up to your discretion

source/fundamentals/builders/aggregates.txt
```diff
@@ -110,6 +107,34 @@ into a new field called ``rating``, effectively renaming the field.
...
+---------
+
+Use the ``documents()`` method to create a
+:manual:`$documents </reference/operator/aggregation/documents/>`
+pipeline stage that returns literal documents from input values.
+
+.. important::
+
+   If you use a ``$documents`` stage in an aggregation pipeline, it must be the first 
+   stage in the pipeline.
```
mongoKart, line 120: This is based on a sentence in [this Google doc](https://docs.google.com/document/d/1yRcAcMUJysh4B-OMUDp2XsjgPgozYj7Mo2PwfepTC_U/edit#). Did I interpret it correctly?

source/includes/fundamentals/code-snippets/builders/AggBuilders.java
```diff
@@ -302,4 +302,13 @@ private void matchStage() {
         match(eq("title", "The Shawshank Redemption"));
         // end match
     }
+
+    private void documents() {
+        // begin documents
+        documents(asList(
+                new Document("title", "The Shawshank Redemption"),
+                new Document("title", "Back to the Future"),
+                new Document("title", "Jurassic Park")));
```
mongoKart, line 311: Is this an acceptable syntax for this stage? One of my teammates noticed that it varies from what is used in the [unit tests.](https://github.com/mongodb/mongo-java-driver/blob/a19fad2b3d93ee818132f14377e42cfaa268e8ee/driver-core/src/test/functional/com/mongodb/client/model/AggregatesTest.java#L170-L171)

source/fundamentals/builders/aggregates.txt
katcharov: This is correct.

The aggregation must also run on the database, rather than on a specific collection. This is called the [aggregation level](https://github.com/mongodb/mongo-java-driver/commit/dee0358bff6bae9b65371a0967180bd342260245#diff-cb241c9477e484cb6edcd5e8e0df80078e9a94f813167089bdc5c7f1d79e1dd7R289), but practically it means that one must use:

            mongoClient.getDatabase("dbName").aggregate(...);

rather than:

            mongoClient.getDatabase("dbName").getCollection("cName").aggregate(...);

source/includes/fundamentals/code-snippets/builders/AggBuilders.java
katcharov: Yes.

While this variant can be more verbose than the "parse" variant, I think we want to avoid encouraging the use of parse (it will be mildly slower, and has other caveats that are incidentally unimportant for our unit tests).

source/fundamentals/builders/aggregates.txt
```diff
@@ -110,6 +107,34 @@ into a new field called ``rating``, effectively renaming the field.
...
+   :start-after: // begin documents
+   :end-before: // end documents
+   :language: java
+   :dedent:
+
+.. tip::
+
+   If you use the ``documents()`` method to provide the input to an aggregation pipeline, 
+   you can call the ``aggregate()`` method on a database instead of on a 
+   collection.
```
katcharov, line 135: (Ah, I see that you do mention what I stated in an earlier comment.)

Consider whether this is a tip, since this is not just a "you may/can", but a "you must", like the "must be first" requirement (iirc, an error is produced if $documents is used with a collection).

mongoKart: I wasn't aware this was a requirement rather than an option. Updated to an Important box.

katcharov (approved):
lgtm!

PR #336, opened 2023-02-24
DOCSP-28004: boolean aggregation expression operators

rustagir:
# Pull Request Info

JIRA - https://jira.mongodb.org/browse/DOCSP-28004
Staging - https://docs-mongodbcom-staging.corp.mongodb.com/java/docsworker-xlarge/DOCSP-28004-boolean-operators/fundamentals/aggregation/aggregation-expression-operations/#boolean-expressions

**This PR is a duplicate of #330 -- merged before technical review was completed, so reverted and reopened this PR.**


source/fundamentals/aggregation/aggregation-expression-operations.txt
```diff
@@ -28,5 +27,90 @@ more naturally readable.
...
+available in the driver and equivalent expressions in the
+Query API. The methods link to API documentation and the Query API
+expressions link to Server manual documentation.
+
+.. note::
+
+   Some aggregation methods share a name with other methods available in
+   the driver. Check the parameter types and the type of
+   the object that the method is called on to confirm that that you are
+   referencing the correct documentation.
```
katcharov, line 57: The other methods (like the `and` [method](https://mongodb.github.io/mongo-java-driver/4.9/apidocs/mongodb-driver-core/com/mongodb/client/model/Filters.html#and(org.bson.conversions.Bson...))) cannot be called on an object, since they are static. (Also, this seems more general than just Booleans.)

I don't think that the issue is that users will end up looking at the wrong docs, but rather that they might read our example below (`and()`), think "I know that method" without clicking the link, and then try to use it. Perhaps we need an initial section that discusses how this integrates with the existing API, and this can be more fittingly mentioned there? Very roughly:

> To use all of the operators below, look for a method that takes a `TExpression`, like `Filters.expr` or `Projections.computed`. You can then start by obtaining the current document using `current()`. This API is typed, so you will need to specify the type of your field via, for example, `current().getInteger("myFieldName")`. [You can use IntelliSense to show which operations are available; maybe screenshot of code completion via `.`]. Values can also be specified via methods like `of`. While this API can be used within [existing] aggregation methods that take `TExpression`, various [existing] methods that return Bson values are not accepted by this API.

(The last sentence is intended to cover what I think is the original point.)

katcharov (changes requested):

source/fundamentals/aggregation/aggregation-expression-operations.txt
```diff
@@ -28,5 +27,90 @@ more naturally readable.
...
+   The driver generates a Query API expression that may be different
+   from the Query API expression provided in each example. However,
+   the expressions are equivalent when evaluated.
+
+Boolean Expressions
+~~~~~~~~~~~~~~~~~~~
+
+You can construct a boolean expression from values of type
+``MqlBoolean`` and boolean aggregation operations. These 
+expressions evaluate to ``true`` or ``false``.
```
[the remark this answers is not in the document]
rustagir: To address these concerns, I removed the sentence about what these operations evaluate to, and changed some of the wording to omit the word "expression" when discussing the Java methods.

source/fundamentals/aggregation/aggregation-expression-operations.txt
rustagir: Will discuss this point with my team to figure out how to incorporate this content into the intro. In the meantime, I have removed this admonition.

source/fundamentals/aggregation/aggregation-expression-operations.txt
```diff
@@ -28,5 +27,78 @@ more naturally readable.
...
+
+   The driver generates a Query API expression that may be different
+   from the Query API expression provided in each example. However,
+   the expressions are equivalent when evaluated.
+
+Boolean Operations
+~~~~~~~~~~~~~~~~~~
+
+You can construct a boolean operation from at least one value of type
+``MqlBoolean`` and one or more boolean aggregation methods.
```
katcharov, line 44: In this Java API, for something like `oneValue.not()`, the "one value" is the `MqlBoolean` `oneValue` that the method is invoked on.

Only one (strike "or more") of the methods below is used per operation. (I think perhaps we should avoid introducing a novel term "boolean aggregation methods".)

```suggestion
You can perform an operation on an ``MqlBoolean`` value using the methods below.
```

(Just for reference, especially in the context of Java: operations are performed, objects are constructed (when a constructor is involved), methods are invoked.)

source/fundamentals/aggregation/aggregation-expression-operations.txt
```diff
@@ -28,5 +27,78 @@ more naturally readable.
...
+available in the driver and equivalent expressions in the
+Query API. The methods link to API documentation and the Query API
+expressions link to Server manual documentation.
+
+.. list-table::
+   :header-rows: 1
+   :widths: 40 60
+
+   * - Java Aggregation Operation
+     - Query API Expression
```
katcharov, line 56:
```suggestion
   * - Java Method
     - Aggregation Pipeline Operator
```
Trying to simplify / avoid introducing new terms.

Since we're just listing methods, and the context is clear enough, so I don't think we need qualifiers (and "aggregation method" sounds odd or misleading). For "Query API Expression", unsure if any docs call `$not` an expression (it isn't one, formally), but in the docs all of [these pages](https://www.mongodb.com/docs/manual/reference/operator/aggregation/) are under the [Aggregation Pipeline Operator](https://www.mongodb.com/docs/manual/reference/operator/aggregation/) page, which seems like a fine enough term.

katcharov (changes requested):

source/fundamentals/aggregation/aggregation-expression-operations.txt
```diff
@@ -28,5 +27,78 @@ more naturally readable.
...
 
+The following sections provide information and examples for
+aggregation expression operations available in the driver.
+The operations are categorized by purpose and functionality.
+
+.. important::
+
+   The driver generates a Query API expression that may be different
+   from the Query API expression provided in each example. However,
+   the expressions are equivalent when evaluated.
```
katcharov, line 38:
```suggestion
.. note::

   The driver generates a Query API expression that may be different
   from the Query API expression provided in each example. However,
   the expressions will produce the same results.
```

Unsure if this is important to know this before using, rather than a "note".

You may mention that this is "because the driver may perform post-processing or other optimizations" to reassure users, but maybe this is too much detail.

source/fundamentals/aggregation/aggregation-expression-operations.txt
```diff
@@ -28,5 +27,78 @@ more naturally readable.
...
+
+Boolean Operations
+~~~~~~~~~~~~~~~~~~
+
+You can construct a boolean operation from at least one value of type
+``MqlBoolean`` and one or more boolean aggregation methods.
+
+The following table describes boolean aggregation expression operators
+available in the driver and equivalent expressions in the
```
katcharov, line 47:
```suggestion
available in the driver and corresponding expressions in the
```

source/fundamentals/aggregation/aggregation-expression-operations.txt
```diff
@@ -28,5 +27,78 @@ more naturally readable.
...
+   from the Query API expression provided in each example. However,
+   the expressions are equivalent when evaluated.
+
+Boolean Operations
+~~~~~~~~~~~~~~~~~~
+
+You can construct a boolean operation from at least one value of type
+``MqlBoolean`` and one or more boolean aggregation methods.
+
+The following table describes boolean aggregation expression operators
```
katcharov, line 46:
```suggestion
The following table describes boolean methods
```

source/fundamentals/aggregation/aggregation-expression-operations.txt
rustagir: Good point. converting to note and adding an appropriate amount of explanation.

katcharov (approved):
LGTM!

(Did not review the "how to use..." section.)

PR #345, opened 2023-03-18
DOCSP-28005: Adding comparison agg operators

DBirtolo-mdb:
# Pull Request Info

JIRA - https://jira.mongodb.org/browse/DOCSP-28005
Staging - https://docs-mongodbcom-staging.corp.mongodb.com/java/docsworker-xlarge/DOCSP-28005-comparison-operations/fundamentals/aggregation/aggregation-expression-operations/#comparison-operations

## Self-Review Checklist

### DIFF
--- a/source/fundamentals/aggregation/aggregation-expression-operations.txt
+++ b/source/fundamentals/aggregation/aggregation-expression-operations.txt
@@ -213,7 +213,7 @@ Query API:
 
    [ { $match: {
          $and: [
-            { $eq: [ "$tempUnit", "Celsius"] }
+            { $eq: [ "$location", "California"] }
    ] } } ]
 
 More Information
--- a/source/fundamentals/aggregation/aggregation-expression-operations.txt
+++ b/source/fundamentals/aggregation/aggregation-expression-operations.txt
@@ -176,22 +176,22 @@ expressions link to Server manual documentation.
    * - Java Method
      - Aggregation Pipeline Operator
 
-   * - `eq() <{+api+}/apidocs/mongodb-driver-core/com/mongodb/client/model/mql/MqlNumber.html#eq(com.mongodb.client.model.mql.MqlValue)>`__
+   * - `eq() <{+api+}/apidocs/mongodb-driver-core/com/mongodb/client/model/mql/MqlValue.html#eq(com.mongodb.client.model.mql.MqlValue)>`__
      - :manual:`$eq </reference/operator/aggregation/eq/>`
 
-   * - `gt() <{+api+}/apidocs/mongodb-driver-core/com/mongodb/client/model/mql/MqlNumber.html#gt(com.mongodb.client.model.mql.MqlValue)>`__
+   * - `gt() <{+api+}/apidocs/mongodb-driver-core/com/mongodb/client/model/mql/MqlValue.html#gt(com.mongodb.client.model.mql.MqlValue)>`__
      - :manual:`$gt </reference/operator/aggregation/gt/>`
 
-   * - `gte() <{+api+}/apidocs/mongodb-driver-core/com/mongodb/client/model/mql/MqlNumber.html#gte(com.mongodb.client.model.mql.MqlValue)>`__
+   * - `gte() <{+api+}/apidocs/mongodb-driver-core/com/mongodb/client/model/mql/MqlValue.html#gte(com.mongodb.client.model.mql.MqlValue)>`__
      - :manual:`$gte </reference/operator/aggregation/gte/>`
 
-   * - `lt() <{+api+}/apidocs/mongodb-driver-core/com/mongodb/client/model/mql/MqlNumber.html#lt(com.mongodb.client.model.mql.MqlValue)>`__
+   * - `lt() <{+api+}/apidocs/mongodb-driver-core/com/mongodb/client/model/mql/MqlValue.html#lt(com.mongodb.client.model.mql.MqlValue)>`__
      - :manual:`$lt </reference/operator/aggregation/lt/>`
 
-   * - `lte() <{+api+}/apidocs/mongodb-driver-core/com/mongodb/client/model/mql/MqlNumber.html#lte(com.mongodb.client.model.mql.MqlValue)>`__
+   * - `lte() <{+api+}/apidocs/mongodb-driver-core/com/mongodb/client/model/mql/MqlValue.html#lte(com.mongodb.client.model.mql.MqlValue)>`__
      - :manual:`$lte </reference/operator/aggregation/lte/>`
 
-   * - `ne() <{+api+}/apidocs/mongodb-driver-core/com/mongodb/client/model/mql/MqlNumber.html#ne(com.mongodb.client.model.mql.MqlValue)>`__
+   * - `ne() <{+api+}/apidocs/mongodb-driver-core/com/mongodb/client/model/mql/MqlValue.html#ne(com.mongodb.client.model.mql.MqlValue)>`__
      - :manual:`$ne </reference/operator/aggregation/ne/>`
 
 The following example creates a pipeline stage that matches all the documents
--- a/source/fundamentals/aggregation/aggregation-expression-operations.txt
+++ b/source/fundamentals/aggregation/aggregation-expression-operations.txt
@@ -160,9 +160,9 @@ the Query API:
 Comparison Operations
 ~~~~~~~~~~~~~~~~~~~~~
 
-You can use comparison operations to compare two different values. You can
-perform a comparison operation on ``MqlValue`` objects using the methods
-described in this section.
+You can use comparison operations to compare two values. The methods
+described in this section perform a comparison operation on ``MqlValue``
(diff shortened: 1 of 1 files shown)


source/fundamentals/aggregation/aggregation-expression-operations.txt
```diff
@@ -157,5 +157,75 @@ the Query API:
                                       then: "extreme temperature",
                                       else: "no advisory" } } } }
 
+Comparison Operations
+~~~~~~~~~~~~~~~~~~~~~
+
+You can use comparison operations to compare two different values. You can
```
mongoKart, line 163: I: Let's get pedantic! Technically, the values can be identical.

```suggestion
You can use comparison operations to compare two values. You can
```

source/fundamentals/aggregation/aggregation-expression-operations.txt
```diff
@@ -157,5 +157,75 @@ the Query API:
...
+perform a comparison operation on ``MqlValue`` objects using the methods
```
mongoKart, line 164: I: Per [style guide](https://www.mongodb.com/docs/meta/style-guide/terminology/alphabetical-terms/#u), avoid "using"

S:

```suggestion
perform a comparison operation on ``MqlValue`` objects by using the methods
```

source/fundamentals/aggregation/aggregation-expression-operations.txt
mongoKart, line 163: S: Can either the first or second sentence be rephrased to avoid two consecutive "You can..." sentences?

source/fundamentals/aggregation/aggregation-expression-operations.txt
```diff
@@ -157,5 +157,75 @@ the Query API:
...
+described in this section.
+
+The following table lists the comparison methods
+available in the driver and corresponding expressions in the
+Query API. The methods link to API documentation and the Query API
+expressions link to Server manual documentation.
```
mongoKart, line 170: S: "Server manual documentation" is redundant

```suggestion
expressions link to Server documentation.
```

source/fundamentals/aggregation/aggregation-expression-operations.txt
```diff
@@ -157,5 +157,75 @@ the Query API:
...
+
+   * - `lte() <{+api+}/apidocs/mongodb-driver-core/com/mongodb/client/model/mql/MqlNumber.html#lte(com.mongodb.client.model.mql.MqlValue)>`__
+     - :manual:`$lte </reference/operator/aggregation/lte/>`
+
+   * - `ne() <{+api+}/apidocs/mongodb-driver-core/com/mongodb/client/model/mql/MqlNumber.html#ne(com.mongodb.client.model.mql.MqlValue)>`__
+     - :manual:`$ne </reference/operator/aggregation/ne/>`
+
+The following example creates a pipeline stage that matches all the documents
+where the ``tempUnit`` is Celsius and the ``tempValue`` is between
+``20`` and ``50``:
```
mongoKart, line 199: S: I don't mind this phrasing, but some users might appreciate absolute clarity

```suggestion
where the ``tempUnit`` field has the value ``"Celsius"`` and the ``tempValue`` field has a value between
``20`` and ``50``:
```

source/fundamentals/aggregation/aggregation-expression-operations.txt
```diff
@@ -157,5 +157,75 @@ the Query API:
...
+               .eq(of("Celsius"))
+            .and(current()
+               .getInteger("tempValue")
+                  .gt(of(20)))
+            .and(current()
+               .getInteger("tempValue")
+                  .lt(of(50))))
+   ));
+
+The following code provides an equivalent aggregation pipeline in the
```
mongoKart, line 216: S: 
```suggestion
The following code shows an equivalent aggregation pipeline in the
```

mongoKart (changes requested):
A couple of small things, but looks really good overall! If I left feedback on anything the group has already decided on, just let me know.

source/fundamentals/aggregation/aggregation-expression-operations.txt
DBirtolo-mdb: This is one of the copied sections.

source/fundamentals/aggregation/aggregation-expression-operations.txt
```diff
@@ -157,5 +157,75 @@ the Query API:
...
+     - :manual:`$gte </reference/operator/aggregation/gte/>`
+
+   * - `lt() <{+api+}/apidocs/mongodb-driver-core/com/mongodb/client/model/mql/MqlNumber.html#lt(com.mongodb.client.model.mql.MqlValue)>`__
+     - :manual:`$lt </reference/operator/aggregation/lt/>`
+
+   * - `lte() <{+api+}/apidocs/mongodb-driver-core/com/mongodb/client/model/mql/MqlNumber.html#lte(com.mongodb.client.model.mql.MqlValue)>`__
+     - :manual:`$lte </reference/operator/aggregation/lte/>`
+
+   * - `ne() <{+api+}/apidocs/mongodb-driver-core/com/mongodb/client/model/mql/MqlNumber.html#ne(com.mongodb.client.model.mql.MqlValue)>`__
+     - :manual:`$ne </reference/operator/aggregation/ne/>`
```
mongoKart, line 195: I: It seems a little strange (and less than ideal, from a usefulness perspective) to have all of the Java methods link to the same API page--especially when I couldn't find some of the methods I was looking for there.

S: Have each method link directly to its documentation in the API. E.g., here's eq():

https://mongodb.github.io/mongo-java-driver/4.9/apidocs/mongodb-driver-core/com/mongodb/client/model/mql/MqlValue.html#eq(com.mongodb.client.model.mql.MqlValue)

mongoKart (changes requested):
One issue with the method links

source/fundamentals/aggregation/aggregation-expression-operations.txt
DBirtolo-mdb: Weird - I don't know how that happened. If you look at the text, I had links to each one individually - or so I thought. I must have just copied the # and not noticed that I it was `MqlNumber` instead of `MqlValue`. Nice catch!

mongoKart (approved):
lgtm

source/fundamentals/aggregation/aggregation-expression-operations.txt
```diff
@@ -157,5 +157,72 @@ the Query API:
                                       then: "extreme temperature",
                                       else: "no advisory" } } } }
 
+Comparison Operations
+~~~~~~~~~~~~~~~~~~~~~
+
+You can use comparison operations to compare two values. The methods
+described in this section perform a comparison operation on ``MqlValue``
+objects.
```
katcharov, line 165: It would help to clarify that all types are of MqlValue. For consistency with the other sections, perhaps something like:

> You can perform a comparison operation on any value of type `MqlValue` using the methods described in this section. All values such as MqlBoolean, Mql...etc. are of type MqlValue.

The term "object" should be avoided, since this might be confused with MqlDocument (which correspond to what the server docs call objects aka documents).

One implication of any type being able to be compared against any other type is that [the comparison order](https://www.mongodb.com/docs/manual/reference/bson-type-comparison-order/#std-label-bson-types-comparison-order) applies (though, that list will be confusing from the point of view of our API, because the API hides several irrelevant items, and we don't use the term "object" but "document" instead).

source/fundamentals/aggregation/aggregation-expression-operations.txt
```diff
@@ -157,5 +157,72 @@ the Query API:
...
+         expr(current()
+            .getString("tempUnit")
+               .eq(of("Celsius"))
+            .and(current()
+               .getInteger("tempValue")
+                  .gt(of(20)))
+            .and(current()
+               .getInteger("tempValue")
+                  .lt(of(50))))
+   ));
```
katcharov, line 211: The example seems quite similar to the boolean example. It seems that we are using "and" a lot in the examples. If this was just a single check, I don't think the example would suffer:

    Arrays.asList(match(
         expr(current()
            .getString("someField")
            .eq(of("some value"))
    )))

katcharov (changes requested):

source/fundamentals/aggregation/aggregation-expression-operations.txt
```diff
@@ -218,11 +213,8 @@ Query API:
 
    [ { $match: {
          $and: [
-            { $eq: [ "$tempUnit", "Celsius"] }, 
-            { $and: [ 
-               { $gt: [ "$tempValue", 20 ] },
-               { $lt: [ "$tempValue", 50 ] } ]
-   } ] } } ]
+            { $eq: [ "$tempUnit", "Celsius"] }
```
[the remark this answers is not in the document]
DBirtolo-mdb: Wow... Yeah, it should

katcharov (approved):

PR #361, opened 2023-04-04
DOCSP-28934: Cleaning up aggregation expressions

DBirtolo-mdb:
# Pull Request Info

JIRA - https://jira.mongodb.org/browse/DOCSP-28934
Staging - https://docs-mongodbcom-staging.corp.mongodb.com/java/docsworker-xlarge/DOCSP-28934/fundamentals/aggregation/aggregation-expression-operations/

## Self-Review Checklist

### DIFF
--- a/source/fundamentals/aggregation/aggregation-expression-operations.txt
+++ b/source/fundamentals/aggregation/aggregation-expression-operations.txt
@@ -425,6 +425,163 @@ the Query API:
      }
    ]
 
+Comparison Operations
+~~~~~~~~~~~~~~~~~~~~~
+
+You can perform a comparison operation on any value of type ``MqlValue``
+using the methods described in this section. For more information about how
+the driver compares values of different types, see
+:manual:`Aggregation Pipeline Operators </reference/bson-type-comparison-order/#std-label-bson-types-comparison-order>`
+in the Server manual.
+
+.. list-table::
+   :header-rows: 1
+   :widths: 40 60
+
+   * - Java Method
+     - Aggregation Pipeline Operator
+
+   * - `eq() <{+api+}/apidocs/mongodb-driver-core/com/mongodb/client/model/mql/MqlValue.html#eq(com.mongodb.client.model.mql.MqlValue)>`__
+     - :manual:`$eq </reference/operator/aggregation/eq/>`
+
+   * - `gt() <{+api+}/apidocs/mongodb-driver-core/com/mongodb/client/model/mql/MqlValue.html#gt(com.mongodb.client.model.mql.MqlValue)>`__
+     - :manual:`$gt </reference/operator/aggregation/gt/>`
+
+   * - `gte() <{+api+}/apidocs/mongodb-driver-core/com/mongodb/client/model/mql/MqlValue.html#gte(com.mongodb.client.model.mql.MqlValue)>`__
+     - :manual:`$gte </reference/operator/aggregation/gte/>`
+
+   * - `lt() <{+api+}/apidocs/mongodb-driver-core/com/mongodb/client/model/mql/MqlValue.html#lt(com.mongodb.client.model.mql.MqlValue)>`__
+     - :manual:`$lt </reference/operator/aggregation/lt/>`
+
+   * - `lte() <{+api+}/apidocs/mongodb-driver-core/com/mongodb/client/model/mql/MqlValue.html#lte(com.mongodb.client.model.mql.MqlValue)>`__
+     - :manual:`$lte </reference/operator/aggregation/lte/>`
+
+   * - `ne() <{+api+}/apidocs/mongodb-driver-core/com/mongodb/client/model/mql/MqlValue.html#ne(com.mongodb.client.model.mql.MqlValue)>`__
+     - :manual:`$ne </reference/operator/aggregation/ne/>`
+
+.. include:: /includes/fundamentals/agg-mql-diff.rst
+
+The following example creates a pipeline stage that matches all the documents
+where the ``location`` field has the value ``"California"``:
+
+.. code-block:: java
+   :copyable: true
+
+   Arrays.asList(match(
+         expr(current()
+            .getString("location")
+               .eq(of("California"))
+   )));
+
+The following code shows an equivalent aggregation pipeline in the
+Query API:
+
+.. code-block:: javascript
+   :copyable: true
+
+   [ { $match: {
+         $and: [
+            { $eq: [ "$location", "California"] }
+   ] } } ]
+
+Conditional Operations
+~~~~~~~~~~~~~~~~~~~~~~
+
+You can perform conditional operations using the methods described in
+this section.
+
+.. list-table::
+   :header-rows: 1
+   :widths: 40 60
+
+   * - Java Method
+     - Aggregation Pipeline Operator
+
+   * - `cond() <{+api+}/apidocs/mongodb-driver-core/com/mongodb/client/model/mql/MqlBoolean.html#cond(T,T)>`__
+     - :manual:`$cond </reference/operator/aggregation/cond/>`
+
+   * - `switchOn() <{+api+}/apidocs/mongodb-driver-core/com/mongodb/client/model/mql/MqlValue.html#switchOn(java.util.function.Function)>`__
+
+       | **Typed Variants**:
+       | `switchArrayOn() <{+api+}/apidocs/mongodb-driver-core/com/mongodb/client/model/mql/MqlArray.html#switchArrayOn(java.util.function.Function)>`__
+       | `switchBooleanOn() <{+api+}/apidocs/mongodb-driver-core/com/mongodb/client/model/mql/MqlBoolean.html#switchBooleanOn(java.util.function.Function)>`__
+       | `switchDateOn() <{+api+}/apidocs/mongodb-driver-core/com/mongodb/client/model/mql/MqlDate.html#switchDateOn(java.util.function.Function)>`__
+       | `switchDocumentOn() <{+api+}/apidocs/mongodb-driver-core/com/mongodb/client/model/mql/MqlDocument.html#switchDocumentOn(java.util.function.Function)>`__
+       | `switchIntegerOn() <{+api+}/apidocs/mongodb-driver-core/com/mongodb/client/model/mql/MqlInteger.html#switchIntegerOn(java.util.function.Function)>`__
+       | `switchMapOn() <{+api+}/apidocs/mongodb-driver-core/com/mongodb/client/model/mql/MqlMap.html#switchMapOn(java.util.function.Function)>`__
+       | `switchNumberOn() <{+api+}/apidocs/mongodb-driver-core/com/mongodb/client/model/mql/MqlNumber.html#switchNumberOn(java.util.function.Function)>`__
+       | `switchStringOn() <{+api+}/apidocs/mongodb-driver-core/com/mongodb/client/model/mql/MqlString.html#switchStringOn(java.util.function.Function)>`__
+
+     - :manual:`$switch </reference/operator/aggregation/switch/>`
+
+.. include:: /includes/fundamentals/agg-mql-diff.rst
+
+Suppose you have a collection of customers with their membership information.
+Originally, customers were either members or not. Over time, membership levels
+were introduced and used the same field. The information stored in this field
+can be one of a few different types, and you want to create a consistent field
+indicating their membership level.
+
+To change what you are doing based on the data type of ``member``, use the
+``switchOn()`` method. If the data type is a string, it indicates the
+membership level and you want to use the data without altering it. If the data
+type is ``Boolean``, check the value with the ``cond()`` method and return a
+string depending on the result. The default membership level is "Gold". If the
+data type is ``Array``, it lists all of the customer's subscriptions, with the
+last one in the array being the current membership. Get the last element using
+the ``last()`` method. If it is not one of these types, label them as a "Guest"
+by using ``defaults()``. Regardless of the type, use the ``project()`` method
+to output the result as ``membershipLevel``.
+
+The following code shows the pipeline stage for this aggregation:
+
+.. code-block:: java
+   :copyable: true
+
+   Arrays.asList(project(fields(
+         computed("membershipLevel",
+               current().getField("member").switchOn(field -> field
+                        .isString(s -> s)
+                        .isBoolean(b -> b.cond(of("Gold"), of("Guest")))
+                        .<MqlString>isArray(a -> a.last())
+                        .defaults(d -> of("Guest"))
+               ))
+   )));
+
+The following code provides an equivalent aggregation pipeline in
+the Query API:
+
+.. code-block:: javascript
+   :copyable: true
+
+   [
+     {
+       $project: {
+         result: {
+           $switch: {
+             branches: [
+               { case:
+                 { $eq: [ { $type: "$member" }, "string" ] },
+                 then:
+                   "$member" },
+               { case:
+                 { $eq: [ { $type: "$member" }, "bool" ] },
+                 then:
+                   { $cond: { if: "$member",
+                     then: "Gold",
+                     else: "Guest" } } },
+               { case:
+                 { $eq: [ { $type: "$member" }, "array" ] },
+                 then:
+                   { $last: "$member" }
+             ],
+             default: "Guest"
+           }
+         }
+       }
+     }
+   ]
+
 Conversion Operations
 ~~~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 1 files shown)


source/fundamentals/aggregation/aggregation-expression-operations.txt
```diff
@@ -438,50 +474,127 @@ types using the methods described in this section.
...
+indicating their membership level.
 
-To convert the ``graduationYear`` field to an integer,
-use the ``parseInteger()`` method. To add five years to this value, use the
-``add()`` method. To store the result as ``reunionYear``,
-use the ``addFields()`` method.
+To change what you are doing based on the data type of ``member``, use the
+``switchOn()`` method. If the data type is a string, it indicates the
+membership level and you want to use the data without altering it. If the data
+type is ``Boolean``, check the value with the ``cond()`` method and return a
```
ccho-mongodb, line 575: Question:
This reads as tutorial instructions rather than a description of the logic. Would it be possible to describe as a description of the logic instead?

ccho-mongodb (changes requested):
Looks pretty good.
I think the cleanup item "Figure out where this page should be in the repo" hasn't been completed yet.
Left a few suggestions and questions.

source/fundamentals/aggregation/aggregation-expression-operations.txt
```diff
@@ -199,12 +210,18 @@ Query API:
...
-                              "$quantity" ] },
-               "$discount" ] } } } ]
+           $subtract: [
+             { $multiply: [ "$itemPrice", "$quantity" ] },
+             "$discount"
+            ]
+         }
+       }
+     }
+   ]
```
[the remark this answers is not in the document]
DBirtolo-mdb: I was modeling it off of other samples and thought it was an unwritten code standard. I am all in favor of compacting it and will do so.

source/fundamentals/aggregation/aggregation-expression-operations.txt
```diff
@@ -438,50 +474,127 @@ types using the methods described in this section.
...
-       
-   * - `asMap() <{+api+}/apidocs/mongodb-driver-core/com/mongodb/client/model/mql/MqlDocument.html#asMap()>`__
+
+   * - `max() for MqlNumber <{+api+}/apidocs/mongodb-driver-core/com/mongodb/client/model/mql/MqlNumber.html#max(com.mongodb.client.model.mql.MqlNumber)>`__
      - *No corresponding operator*
 
-   * - `asString() <{+api+}/apidocs/mongodb-driver-core/com/mongodb/client/model/mql/MqlValue.html#asString()>`__
-     - :manual:`$toString </reference/operator/aggregation/toString/>`
+   * - `min() for MqlInteger <{+api+}/apidocs/mongodb-driver-core/com/mongodb/client/model/mql/MqlInteger.html#min(com.mongodb.client.model.mql.MqlInteger)>`__
+     - *No corresponding operator*
```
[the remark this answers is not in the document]
DBirtolo-mdb: Yes this is different. $min takes in an expression or looks at all the fields and returns the minimum value. This version of min is specific to integers and returns the minimum between the integer you are calling it on and the integer you provide.

```
quantity.min(5)
```

would return either `quantity` or 5, whichever is lower, whereas

```
$min("$quantity")
```

would return the minimum value of the `quantity` field. So while they both do very similar things, you can't use `$min` to do what this version of `min` does, which is why I didn't link them.

source/fundamentals/aggregation/aggregation-expression-operations.txt
```diff
@@ -438,50 +474,127 @@ types using the methods described in this section.
...
+
+   * - `lte() <{+api+}/apidocs/mongodb-driver-core/com/mongodb/client/model/mql/MqlValue.html#lte(com.mongodb.client.model.mql.MqlValue)>`__
+     - :manual:`$lte </reference/operator/aggregation/lte/>`
+
+   * - `max() for MqlInteger <{+api+}/apidocs/mongodb-driver-core/com/mongodb/client/model/mql/MqlInteger.html#max(com.mongodb.client.model.mql.MqlInteger)>`__
      - *No corresponding operator*
-       
-   * - `asMap() <{+api+}/apidocs/mongodb-driver-core/com/mongodb/client/model/mql/MqlDocument.html#asMap()>`__
+
+   * - `max() for MqlNumber <{+api+}/apidocs/mongodb-driver-core/com/mongodb/client/model/mql/MqlNumber.html#max(com.mongodb.client.model.mql.MqlNumber)>`__
```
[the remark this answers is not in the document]
DBirtolo-mdb: Similar to my explanation for above with `$min`, this is a very similar, but different function.

source/fundamentals/aggregation/aggregation-expression-operations.txt
```diff
@@ -809,30 +978,81 @@ operations treat the arrays as sets.
...
+       "street": "9286 Bethany Glens",
+       "city": "Vasqueztown",
+       "state": "CO"
+       "zip": 22939
+     },
+   }
+
+To access the nested ``currentAddress`` document, use the ``getDocument()``
+method. To get the string code for the state, use the ``getString()`` method.
+To compare the string field, use the ``.eq()`` method.
```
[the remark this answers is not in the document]
DBirtolo-mdb: typo - thanks!

source/fundamentals/aggregation/aggregation-expression-operations.txt
```diff
@@ -421,15 +453,19 @@ the Query API:
...
+Comparison Operations
 ~~~~~~~~~~~~~~~~~~~~~
 
-You can perform a conversion operation to convert between ``MqlValue``
-types using the methods described in this section.
+You can perform a comparison operation on a value of type ``MqlValue``
+using the methods described in this section. For more information about how
+the driver compares values of different types, see
+:manual:`Aggregation Pipeline Operators </reference/bson-type-comparison-order/#std-label-bson-types-comparison-order>`
+in the Server manual.
```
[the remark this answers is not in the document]
DBirtolo-mdb: There is not a one-to-one mapping between categories. I can check to see if there are other sections that have explanations that need to be considered, but we can't map each section.

source/fundamentals/aggregation/aggregation-expression-operations.txt
DBirtolo-mdb: I tried to pull the descriptions into the previous paragraph and make this paragraph modeled like the others in this topic. I'm not sure if this addresses your concerns, but I think it now models what the other sections have.

source/fundamentals/aggregation/aggregation-expression-operations.txt
```diff
@@ -138,9 +138,8 @@ the Server manual documentation.
 Arithmetic Operations
 ~~~~~~~~~~~~~~~~~~~~~
 
-You can use arithmetic operations to manipulate numbers. You can perform an
-arithmetic operation on a value of ``MqlInteger`` or ``MqlNumber`` using the
-methods described in this section.
+You can perform an arithmetic operation on a value of type ``MqlInteger`` or
+``MqlNumber`` using the methods described in this section.
```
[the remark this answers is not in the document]
DBirtolo-mdb: This was filtered out through multiple internal and external reviews, so I'm hesitant to change it at this point in time.

DBirtolo-mdb:
I missed the bullet about location - it was added to the ticket after I started working on it and I missed that element.

DBirtolo-mdb:
Updated the staging link since it's a new location in the docs.

source/fundamentals/aggregation/aggregation-expression-operations.txt
```diff
@@ -149,22 +148,34 @@ methods described in this section.
    * - Java Method
      - Aggregation Pipeline Operator
 
-   * - `abs() <{+api+}/apidocs/mongodb-driver-core/com/mongodb/client/model/mql/MqlNumber.html#abs()>`__
+   * - `abs() for MqlInteger <{+api+}/apidocs/mongodb-driver-core/com/mongodb/client/model/mql/MqlInteger.html#abs()>`__
```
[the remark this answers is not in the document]
ccho-mongodb: Could you change the table column widths to prevent wrapping?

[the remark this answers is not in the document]
ccho-mongodb: Here's what I currently see: 

Screenshot 2023-04-07 at 10 19 34 AM

The content in the left column is wrapping, but could probably fit if wider.

source/fundamentals/aggregation/aggregation-expression-operations.txt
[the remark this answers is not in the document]
ccho-mongodb: I think it would probably be good to remove this since it's the only one. That would help present the sections uniformly and prevent confusion about why only one category on this page is related to the category in the manual.

source/fundamentals/aggregation/aggregation-expression-operations.txt
```diff
@@ -438,50 +474,127 @@ types using the methods described in this section.
...
+can be one of a few different types, and you want to create a consistent field
+indicating their membership level.
 
-To convert the ``graduationYear`` field to an integer,
-use the ``parseInteger()`` method. To add five years to this value, use the
-``add()`` method. To store the result as ``reunionYear``,
-use the ``addFields()`` method.
+To change what you are doing based on the data type of ``member``, use the
+``switchOn()`` method. If the data type is a string, it indicates the
+membership level and you want to use the data without altering it. If the data
```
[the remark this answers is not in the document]
ccho-mongodb: Issue:

The new version seems to imply the other data types do not result in a membership level.

Suggestion:

Make this passage easier to read by using consistent terminology.

Currently:
If the type is a string, it indicates the membership level. If the data type is a boolean value, it indicates whether the member is a Gold member. If the type is an array, it lists all of the customer's subscriptions, with the last one in the array being the current membership.

Suggested:
If the data type is a string, it returns the membership level value.
If the data type is a boolean, it returns either a Gold or Guest membership level.
If the data type is an array, it returns an array of the customer's subscriptions and appends the current membership level.

source/fundamentals/builders/aggregates.txt
```diff
@@ -19,7 +19,10 @@ In this guide, you can learn how to use the `Aggregates <{+api+}/apidocs/mongodb
 class which provides static factory methods that build :manual:`aggregation pipeline
 stages </meta/aggregation-quick-reference/#stages>` in the MongoDB Java driver.
 
-For a more thorough introduction to Aggregation, see our :ref:`Aggregation guide <java-aggregation>`.
+These builders serve as a wrapper for aggregation expressions and provide some
+aggregation functionality. For a more thorough introduction to Aggregation, see
```
ccho-mongodb, line 23: Question:

I could be remembering incorrectly, but didn't Maxim want to avoid any mention of these expressions as a "wrapper"? If so, could the first sentence be omitted?

source/fundamentals/builders/aggregates.txt
```diff
@@ -19,7 +19,10 @@ In this guide, you can learn how to use the `Aggregates <{+api+}/apidocs/mongodb
...
+our :ref:`Aggregation guide <java-aggregation>`. For a more complete syntax
```
ccho-mongodb, line 24: Issue:
To me, "for a more complete...enables greater aggregation functionality" implies that the Aggregation Expressions subsumes the content on this page, which I don't think is the case.

Suggestion:
Omit this paragraph.

source/fundamentals/builders/aggregation-expressions.txt
```diff
@@ -1,6 +1,6 @@
-=================================
-Aggregation Expression Operations
-=================================
+=======================
+Aggregation Expressions
```
ccho-mongodb, line 2: Question:

Is this the correct term? From our initial discussions with Maxim, I thought we wanted to call them "Aggregation Expression Operators"

ccho-mongodb (changes requested):
Looks like it's nearly there, but would like to take another look once comments are addressed.

source/fundamentals/builders/aggregation-expressions.txt
DBirtolo-mdb: You are correct - since the ticket changed the name, I thought that was what it was supposed to be, but that was not correct.

source/fundamentals/aggregation/aggregation-expression-operations.txt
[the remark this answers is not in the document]
DBirtolo-mdb: Updated the language and moved it back into the paragraph below since now it seems more appropriate for describing what the example does.

source/fundamentals/builders/aggregates.txt
DBirtolo-mdb: I think it is important for people to have a link to the greater topic on aggregation, and also important for folks to know that this is one solution, while there is another. I did some rewording, but I think we need something here.

source/fundamentals/builders/aggregation-expression-operations.txt
```diff
@@ -138,52 +140,55 @@ the Server manual documentation.
...
-You can use arithmetic operations to manipulate numbers. You can perform an
-arithmetic operation on a value of ``MqlInteger`` or ``MqlNumber`` using the
-methods described in this section.
+You can perform an arithmetic operation on a value of type ``MqlInteger`` or
+``MqlNumber`` using the methods described in this section.
 
 .. list-table::
    :header-rows: 1
-   :widths: 40 60
+   :widths: 60 60
```
ccho-mongodb, line 148: Suggestion:

There isn't a good reason to use a total that's not equal to 100. If you want them to be equal, "50 50" is equivalent.

source/fundamentals/builders/aggregation-expression-operations.txt
```diff
@@ -694,73 +813,80 @@ the Query API:
...
-string depending on the result. The default membership level is "Gold". If the
-data type is ``Array``, it lists all of the customer's subscriptions, with the
-last one in the array being the current membership. Get the last element using
-the ``last()`` method. If it is not one of these types, label them as a "Guest"
-by using ``defaults()``. Regardless of the type, use the ``project()`` method
-to output the result as ``membershipLevel``.
+The ``dayOfWeek()`` method determines which day of the week it is and checks
+if it's a Monday by comparing it to ``2`` using the ``eq()`` method. The
+``"Anerica/New_York"`` parameter specifies which number corresponds to
+Monday.
```
ccho-mongodb, line 876: Question:
Doesn't the America/New_York parameter specify the time zone, not "which number corresponds to Monday"?

source/fundamentals/builders/aggregation-expression-operations.txt
```diff
@@ -694,73 +813,80 @@ the Query API:
...
-membership level and you want to use the data without altering it. If the data
-type is ``Boolean``, check the value with the ``cond()`` method and return a
-string depending on the result. The default membership level is "Gold". If the
-data type is ``Array``, it lists all of the customer's subscriptions, with the
-last one in the array being the current membership. Get the last element using
-the ``last()`` method. If it is not one of these types, label them as a "Guest"
-by using ``defaults()``. Regardless of the type, use the ``project()`` method
-to output the result as ``membershipLevel``.
+The ``dayOfWeek()`` method determines which day of the week it is and checks
+if it's a Monday by comparing it to ``2`` using the ``eq()`` method. The
```
ccho-mongodb, line 874: Question:
Does dayOfWeek() actually check both those things? If not, use separate sentences to separate the subject.

source/fundamentals/builders/aggregation-expression-operations.txt
```diff
@@ -919,71 +1133,66 @@ the Query API:
...
 
-The following example creates a pipeline stage that matches all the documents
-where the ``location`` field has the value ``"California"``:
+   * - `isStringOr() <{+api+}/apidocs/mongodb-driver-core/com/mongodb/client/model/mql/MqlValue.html#isStringOr(com.mongodb.client.model.mql.MqlString)>`__
+     - *No corresponding operator*
 
-.. code-block:: java
-   :copyable: true
+Suppose you have a collection of rating data. Early versions of the reviews
+allowed users to submit reviews without a star rating. You want to clean up
```
ccho-mongodb, line 1177: Suggestion:

Seems unnecessary to mention multiple prior versions (plural). Also, it might be helpful to identify what it was a version of.

```suggestion
Suppose you have a collection of rating data. An earlier version of the reviews schema
allowed users to submit reviews without a star rating. You want to clean up
```

source/fundamentals/builders/aggregation-expression-operations.txt
```diff
@@ -919,71 +1133,66 @@ the Query API:
...
+the data, and convert any reviews without a star rating to be the minimum
+value of 1 star.
 
-   Arrays.asList(match(
-         expr(current()
-            .getString("location")
-               .eq(of("California"))
-   )));
+The ``isNumberOr()`` method returns either the value of ``originalNumber``, or
```
ccho-mongodb, line 1181: Question:
Should this be "originalRating" to match the example? If not, I think it needs some context/description or will otherwise confuse the reader (like it did to me).

ccho-mongodb (approved):
A few questions and suggestions, but otherwise LGTM!

source/fundamentals/builders/aggregation-expression-operations.txt
DBirtolo-mdb: It's actually a thing that specifies which day is 1 - whether the week starts on Sunday or Monday.

source/fundamentals/builders/aggregation-expression-operations.txt
DBirtolo-mdb: Whoops. Nice catch!

ccho-mongodb:
Merged in favor of a review of the complete feature branch.

PR #367, opened 2023-04-12
DOCSP-27390 aggregation operations

ccho-mongodb:
# Pull Request Info

JIRA - https://jira.mongodb.org/browse/DOCSP-27390
[**Staging**](https://docs-mongodbcom-staging.corp.mongodb.com/java/docsworker-xlarge/docsp-27390-aggregation-operations/fundamentals/aggregation-expression-operations/)

## Self-Review Checklist


source/fundamentals/builders/aggregation-expression-operations.txt
```diff
@@ -0,0 +1,1198 @@
...
+                  .filter(showtime -> {
+                        var totalSeats = showtime
+                              .<MqlInteger>getArray("seats")
+                              .sum(n -> n);
+                        var isAvailable = showtime
+                              .getInteger("ticketsBought")
+                              .lt(totalSeats);
+                        return isAvailable;
+                  })
+   ))));
```
katcharov, line 354:
```
   asList(project(fields(
       computed("availableShowtimes", current()
           ...
   ))));
```
As a convention for formatting all Java examples:
1. Let's indent each level by 4 spaces (this one had 9, 3, 5...), to save space and for consistency.
2. If there are multiple stages, we should put `project` etc. on a new level. If there are not multiple, keep them on the same line (as I think we do everywhere).
3. However, even single fields ("computed") should go on separate lines, because things get too long otherwise?
4. Just as in `.filter(showtime -> showtime` [...], showtime would be on the same line as filter, current should be on the same line in `computed("availableShowtimes", current()` (and then a linebreak to chain off of current, or whatever initiates the expression).
5. (pending a prior comment) We should extract current fields into variables: `var abc = current().getSomething("abc")`.
6. I think that `asList(match(expr(current()` belong on one line, because this would correspond to `.match(c -> c` if stages used the same pattern (or, for that matter, `.filter(x -> x` even in this new API). But I'm a bit hesitant to suggest this one because things get a bit long, and mildly inconsistent with compute/fields.
7. Parenthesis should line up with the lines that introduce them. For example, above, there are 3 parens on the first line, but 4 on the last one. One of those should be moved to a new line directly before, to match the paren on `computed(`. Alternatively, if it really should be collapsed, it should collapse with the prior line (so, `}))`).

source/fundamentals/builders/aggregation-expression-operations.txt
```diff
@@ -0,0 +1,1198 @@
...
+
+The following code shows the pipeline for this aggregation:
+
+.. code-block:: java
+   :copyable: true
+
+   asList(match(
+         expr(current()
+               .getDocument("currentAddress")
+               .getString("state")
```
katcharov, line 993: I wonder if we should include a field that has a `.` in it to demonstrate that these `get...` methods handle those properly and do not support "paths" in the way that raw MQL JSON does (like `$currentAddress.state` below, which is not a single field with a dot but one path with two fields).

source/fundamentals/builders/aggregation-expression-operations.txt
```diff
@@ -0,0 +1,1198 @@
...
+
+.. code-block:: javascript
+   :copyable: true
+
+   [
+     {
+       $project: {
+         extremeTemp: { $or: [ { $lt: ["$temperature", 10] },
+                               { $gt: ["$temperature", 95] } ] }
+   }}]
```
katcharov, line 435:
```suggestion
   [ { $project: {
       extremeTemp: { $or: [ { $lt: ["$temperature", 10] },
                             { $gt: ["$temperature", 95] } ] }
   }}]
```
I'm not sure what the spacing should be. Some of the }]}}]s are spaced, but others are not. But if we're going to condense, I think we should condense the first two.

source/fundamentals/builders/aggregation-expression-operations.txt
```diff
@@ -0,0 +1,1198 @@
...
+                     then: "Gold",
+                     else: "Guest" } } 
+                 },
+               { case:
+                 { $eq: [ { $type: "$member" }, "array" ] },
+                 then: { $last: "$member" }
+               }
+             ],
+             default: "Guest"}
+   }}}]
```
katcharov, line 589: I think the formatting makes this harder to follow what is going on. For example, the { after case is broken onto the same level as case, but after then (which is on the same level as case), the line is broken after the {, and then the $cond is indented. The } after default is collapsed to that same line, but not the one after that last `then`.

I'm not sure what the convention should be, but maybe something like:
1. To start, expand everything using a typical json formatter.
2. If a "level" is collapsed, it should all be put one one line: `case: { $eq: [ { $type: "$member" }, "string" ] },`.
3. Where it cannot be put on one line, but collapsing is helpful, keep all ({[s strictly matching some line they were introduced on. So, `[ { $project: {` is an acceptable collapse, but the trailing `} } ]` must be on a similarly-indented line at the bottom, with nothing else on that line.
4. When choosing indentation, attempt to mimic what was done in the Java, for the sake of easier comparison.

source/fundamentals/builders/aggregation-expression-operations.txt
```diff
@@ -0,0 +1,1198 @@
...
+The following code provides an equivalent aggregation pipeline in
+the Query API:
+
+.. code-block:: javascript
+   :copyable: true
+
+   [
+     {
+       $project: {
+         result: {
```
katcharov, line 569: I think this should be `membershipLevel`?

Consider doublechecking the other examples to make sure they match up with the Java.

katcharov (changes requested):
Gave this an initial pass, it's looking quite good!

.github/.backportrc.json
```diff
@@ -1,13 +1,17 @@
...
   "autoMerge": true,
-  "autoMergeMethod": "squash",
-
-  "branchLabelMapping": {
-    "^auto-backport-to-(.+)$": "$1"
-  }
-}
\ No newline at end of file
+  "autoMergeMethod": "squash"
+}
```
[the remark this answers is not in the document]
rustagir: I think this might be a change from rebase or merge. @ccho-mongodb Do you have any insight?

source/fundamentals/builders.txt
```diff
@@ -7,6 +7,7 @@ Builders
 .. toctree::
 
    /fundamentals/builders/aggregates
+   /fundamentals/builders/aggregation-expression-operations
```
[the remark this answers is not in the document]
rustagir: Slating this for discussion with Dylan and Chris. Maybe we can potentially leave it here for now to get the page out, and then move it

source/fundamentals/builders/aggregates.txt
```diff
@@ -17,7 +17,9 @@ In this guide, you can learn how to use the `Aggregates <{+api+}/apidocs/mongodb
 class which provides static factory methods that build :manual:`aggregation pipeline
 stages </meta/aggregation-quick-reference/#stages>` in the MongoDB Java driver.
 
-For a more thorough introduction to Aggregation, see our :ref:`Aggregation guide <java-aggregation>`.
+To learn more about aggregation, see our :ref:`Aggregation guide <java-aggregation>`.
+For a different way to use aggregation that includes a complete syntax solution,
+see :ref:`Aggregation Expressions <java-aggregation-expression-operations>`.
```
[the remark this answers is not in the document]
rustagir: reworded to include the phrasing you highlighted

source/fundamentals/builders/aggregation-expression-operations.txt
```diff
@@ -0,0 +1,1198 @@
...
+
+   * - Java Method
+     - Aggregation Pipeline Operator
+
+   * - `abs() for MqlInteger <{+api+}/apidocs/mongodb-driver-core/com/mongodb/client/model/mql/MqlInteger.html#abs()>`__
+       | `abs() for MqlNumber <{+api+}/apidocs/mongodb-driver-core/com/mongodb/client/model/mql/MqlNumber.html#abs()>`__
+     - :manual:`$abs </reference/operator/aggregation/abs/>`
+
+   * - `add() for MqlInteger <{+api+}/apidocs/mongodb-driver-core/com/mongodb/client/model/mql/MqlInteger.html#add(int)>`__
+       | `add() for MqlNumber <{+api+}/apidocs/mongodb-driver-core/com/mongodb/client/model/mql/MqlNumber.html#add(com.mongodb.client.model.mql.MqlNumber)>`__
```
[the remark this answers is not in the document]
rustagir: I think this might have just been a formatting error where an extra line break was needed

source/includes/fundamentals/agg-mql-diff.rst
```diff
@@ -0,0 +1,5 @@
+.. note::
+
+   While each Java method is functionally equivalent to the
```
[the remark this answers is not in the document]
rustagir: Ok -- moving the content from this note into the text under the main operations heading. Removing from each section.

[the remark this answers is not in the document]
rustagir: will delete the .rst include file when we approve this approach

source/fundamentals/builders/aggregation-expression-operations.txt
```diff
@@ -0,0 +1,1198 @@
...
+     {
+       $addFields: {
+         reunionYear: {
+           $add: [ { $toInt: "$graduationYear" }, 5 ]}}
+   }]
+
+Custom Operations
+~~~~~~~~~~~~~~~~~
+
+You can apply custom functions to values of type
```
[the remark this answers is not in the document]
rustagir: I would prefer to keep the custom operations section separate, perhaps we can rename "Convenience" to something else. Or, we could move this to a separate section altogether and call it Representation of Literals or something like that.

source/fundamentals/builders/aggregation-expression-operations.txt
```diff
@@ -0,0 +1,1198 @@
...
+
+Suppose you have a collection of theater data where each document represents
+an individual theater. One of the fields in that document is a map of the
+ticket prices for different age groups. You want to find all the theaters that
+have a rate for a senior age group.
+
+The ``has()`` method determines whether the ``rates`` map has a ``senior``
+field.
+
+The following code shows the pipeline for this aggregation:
```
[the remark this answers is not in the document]
rustagir: changed example and added a tip

source/fundamentals/builders/aggregation-expression-operations.txt
```diff
@@ -0,0 +1,1198 @@
...
+where the ``location`` field has the value ``"California"``:
+
+.. code-block:: java
+   :copyable: true
+
+   asList(match(
+         expr(current()
+            .getString("location")
+            .eq(of("California")))
+   ));
```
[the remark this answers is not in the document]
rustagir: the array section has a complicated example, so let me know if that one should not be extracted

source/fundamentals/builders/aggregation-expression-operations.txt
rustagir: Yes, these are tricky.

rustagir: also making these 4-spaced

source/fundamentals/builders.txt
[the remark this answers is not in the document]
ccho-mongodb: We can move out of builders, but I don't think it should be a sub-page of Aggregation. Perhaps, it can be its own item that comes sequentially after Aggregation.

source/fundamentals/builders/aggregation-expression-operations.txt
katcharov: Agreed, but the deviation would be intentional to demonstrate a "gotcha", and we can mention it in the wording "...has legacy data that uses a `.` in the field name..."

source/fundamentals/builders/aggregation-expression-operations.txt
```diff
@@ -150,68 +153,66 @@ You can perform an arithmetic operation on a value of type ``MqlInteger`` or
...
-                  .subtract(current()
-                     .getInteger("discount")))
-   )));
+   var precip = current().getInteger("precipitation");
+
+   asList(group(
+       "$month",
+       avg(
+           "avgPrecipMM",
+           precip.multiply(25.4))
```
katcharov, line 202:
```suggestion
       avg("avgPrecipMM", precip.multiply(25.4))
```
I think one line will fit? (Following convention of "computed")

source/fundamentals/builders/aggregation-expression-operations.txt
```diff
@@ -150,68 +153,66 @@ You can perform an arithmetic operation on a value of type ``MqlInteger`` or
...
-                  .getInteger("itemPrice")
-                  .multiply(current()
-                     .getInteger("quantity"))
-                  .subtract(current()
-                     .getInteger("discount")))
-   )));
+   var precip = current().getInteger("precipitation");
+
+   asList(group(
+       "$month",
```
katcharov, line 199: Does this accept something like `var date = current().getDate("$date")` `date.getMonth()` instead of `$month`? This would demonstrate an expression in the group here.

source/includes/fundamentals/agg-mql-diff.rst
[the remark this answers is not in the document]
katcharov: Looks good to me

source/fundamentals/builders/aggregation-expression-operations.txt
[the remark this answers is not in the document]
katcharov: I agree about separate - they do look different. It just seems that "pass..." is definitely a "convenience" method (operation?)

But the things being called convenience at the moment are... I don't know what to call them. "Conversions"? But we have a section like that. Values? I guess they are not literals from the perspective of Java (since one can put `of(myVariable)`).

source/fundamentals/builders/aggregation-expression-operations.txt
```diff
@@ -415,24 +411,22 @@ The following code shows the pipeline for this aggregation:
    var temperature = current().getInteger("temperature");
 
    asList(project(fields(
-            computed("extremeTemp",
-               temperature
-                  .lt(of(10))
-                  .or(temperature.gt(of(95)))
-   ))));
+       computed("extremeTemp",
+           temperature
```
katcharov, line 415:
```suggestion
       computed("extremeTemp", temperature
```
(and un-indent following lines)

source/fundamentals/builders/aggregation-expression-operations.txt
```diff
@@ -462,38 +456,38 @@ using the methods described in this section.
...
-   asList(match(
-         expr(current()
-            .getString("location")
-            .eq(of("California")))
-   ));
+   var location = current().getString("location");
+
+   asList(match(expr(location
+       .eq(of("California"))
+   )));
```
katcharov, line 482: Looks better. I am not sure, but maybe on the same line? Up to you.
```suggestion
   asList(match(expr(location.eq(of("California")))));
```

source/fundamentals/builders/aggregation-expression-operations.txt
```diff
@@ -1045,24 +1029,26 @@ You can perform a map operation on a value of either type ``MqlMap`` or
...
 
-   asList(match(
-         expr(current()
-               .getMap("rates")
-               .has(of("senior")))
-   ));
+   var inventory = current().getMap("inventory");
+
+   asList(project(fields(
+       computed("inventoryArray", inventory.entries())
```
katcharov, line 1050: How about something like

`warehouseInventory.entries().sum(warehouse -> warehouse.getValue())`

to get the sum of all inventory in the warehouse? (I am assuming that it maps from warehouse ids to integers, and the integers count the inventory).

This would demonstrate usage of the entry type.

source/fundamentals/builders/aggregation-expression-operations.txt
```diff
@@ -1187,11 +1168,11 @@ The following code shows the pipeline for this aggregation:
...
 
+   var rating = current().getField("originalRating");
+
    asList(project(fields(
-            computed("filteredRating",
-               current()
-                  .getField("originalRating")
-                  .isNumberOr(of(1)))
+       computed("filteredRating", rating
+           .isNumberOr(of(1)))
```
katcharov, line 1175: Should this have the MQL below?

source/fundamentals/builders/aggregation-expression-operations.txt
```diff
@@ -987,27 +966,32 @@ The following code shows the pipeline for this aggregation:
...
 Map Operations
 ~~~~~~~~~~~~~~
 
 You can perform a map operation on a value of either type ``MqlMap`` or
 ``MqlEntry`` using the methods described in this section.
 
+.. tip::
+   
+   You should represent data in a map if the data maps
+   arbitrary keys such as dates or item IDs to values.
```
katcharov, line 993:
```suggestion
   You should represent data as a map if the data maps
   arbitrary keys such as dates or item IDs to values.
```

source/fundamentals/builders/aggregation-expression-operations.txt
rustagir: changed example to use minN.

source/fundamentals/aggregation-expression-operations.txt
```diff
@@ -979,7 +976,11 @@ the Query API:
 .. code-block:: javascript
    :copyable: true
 
-   [ { $match: { $eq: [ "$currentAddress.state", "WA" ] } } ]
+   [ 
+       { $addFields: { a: { $objectToArray: { $getField: "mailing.address" } } } },
+       { $match: { "a.k": "state", "a.v": "WA" } },
+       { $project: { a: 0 } }
```
katcharov, line 982: I think this should only have one stage, because the Java only specifies one stage. I think something like: 

`[ { $match: { $eq: [  { $getField: "mailing.address" }, "WA" ] } } ]`

but it would need to be confirmed.

source/fundamentals/builders/aggregation-expression-operations.txt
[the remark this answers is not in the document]
katcharov: Let's try and see...

source/fundamentals/aggregation-expression-operations.txt
```diff
@@ -0,0 +1,1211 @@
...
+   asList(project(fields(
+       computed("availableShowtimes", showtimes
+           .filter(showtime -> {
+               var totalSeats = showtime
+                   .<MqlInteger>getArray("seats")
+                   .sum(n -> n);
+               var isAvailable = showtime
+                   .getInteger("ticketsBought")
+                   .lt(totalSeats);
+               return isAvailable;
```
katcharov, line 350:
```suggestion
            var seats = showtime.<MqlInteger>getArray("seats");
            var totalSeats = seats.sum(n -> n);
            var ticketsBought = showtime.getInteger("ticketsBought")
            var isAvailable = ticketsBought.lt(totalSeats);
            return isAvailable;
```
I think maybe you are right that this is cleaner. At least it saves two lines.

source/fundamentals/builders.txt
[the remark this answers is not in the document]
katcharov: I think this looks ok, since it is somewhat in line with the CRUD Operations heading above...

source/fundamentals/aggregation-expression-operations.txt
```diff
@@ -0,0 +1,1211 @@
...
+     - :manual:`$max </reference/operator/aggregation/max/>`
+
+   * - `maxN() <{+api+}/apidocs/mongodb-driver-core/com/mongodb/client/model/mql/MqlArray.html#maxN(com.mongodb.client.model.mql.MqlInteger)>`__
+     - :manual:`$maxN </reference/operator/aggregation/maxN-array-element/>`
+
+   * - `min() <{+api+}/apidocs/mongodb-driver-core/com/mongodb/client/model/mql/MqlArray.html#min(T)>`__
+     - :manual:`$min </reference/operator/aggregation/min/>`
+
+   * - `minN() <{+api+}/apidocs/mongodb-driver-core/com/mongodb/client/model/mql/MqlArray.html#minN(com.mongodb.client.model.mql.MqlInteger)>`__
+     - :manual:`$map </reference/operator/aggregation/minN-array-element/>`
```
katcharov, line 319: minN should map to minN

and I think the map method is missing

source/fundamentals/aggregation-expression-operations.txt
rustagir: I did a lot of testing for this, and ran into a lot of trouble because the top level field name contains a period. I really couldn't figure out how to write the valid MQL for this using a plain match stage!

source/fundamentals/aggregation-expression-operations.txt
```diff
@@ -0,0 +1,1211 @@
...
+   * - `union() <{+api+}/apidocs/mongodb-driver-core/com/mongodb/client/model/mql/MqlArray.html#union(com.mongodb.client.model.mql.MqlArray)>`__
+     - :manual:`$setUnion </reference/operator/aggregation/setUnion/>`
+
+   * - `unionArrays() <{+api+}/apidocs/mongodb-driver-core/com/mongodb/client/model/mql/MqlArray.html#unionArrays(java.util.function.Function)>`__
+     - :manual:`$setUnion </reference/operator/aggregation/setUnion/>`
+
+.. tip::
+
+   When retrieving arrays, you must specify the type of array that is returned
+   if you want to continue working with the return value.
```
katcharov, line 342: It might help if this had the example in it, I think it is referring to specifying MqlDocument in `current().<MqlDocument>getArray("showtimes");`? And the "continue working with", maybe "if you need to work with the values of the array as that specific type".

Maybe it is better to put this point together with the example, like "the getArray method is used to ... . Specifying the type using <MqlDocument> ensures that `showtime` is an `MqlDocument`" or something like this.

I am unsure, because the "tip" does highlight this important point which might be skipped if it was buried in a paragraph, but when I read it here, it's not clear what it means to specify the type of the array, or to continue working with the return value (note that you can continue working with the return value even without specifying, since it will just assume that the elements are of the general type MqlValue).

source/fundamentals/aggregation-expression-operations.txt
```diff
@@ -477,17 +475,15 @@ where the ``location`` field has the value ``"California"``:
...
+   asList(match(expr(location.eq(of("California")))));
 
 The following code provides an equivalent aggregation pipeline in
 the Query API:
 
 .. code-block:: javascript
    :copyable: true
 
-   [ { $match: { $eq: [ "$location", "California" ] } } ]
+   [ { $match: { location: { $eq: "California" } } } ]
```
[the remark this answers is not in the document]
rustagir: discussed in slack

source/fundamentals/aggregation-expression-operations.txt
```diff
@@ -0,0 +1,1211 @@
...
+     - *No corresponding operator*
+
+   * - `ofMap() <{+api+}/apidocs/mongodb-driver-core/com/mongodb/client/model/mql/MqlValues.html#ofMap()>`__
+     - *No corresponding operator*
+
+   * - `ofNull() <{+api+}/apidocs/mongodb-driver-core/com/mongodb/client/model/mql/MqlValues.html#ofNull()>`__
+     - *No corresponding operator*
+
+Refer to any of the other sections for examples using these convenience
+operators.
```
katcharov: Let's add a note that says something like:

> Values provided to these methods are treated literally. For example, `of("$x")` represents the string value `"$x"`, rather than a field `x`.

source/fundamentals/aggregation-expression-operations.txt
```diff
@@ -0,0 +1,1211 @@
...
+                   .lt(totalSeats);
+               return isAvailable;
+           }))
+   )));
+
+.. note::
+
+   To improve readability, the previous example assigns intermediary values to
+   the ``totalSeats`` and ``isAvailable`` variables. If you don't pull
+   out these intermediary values as variables, the code still produces
```
katcharov, line 400:
```suggestion
   out these intermediary values into variables, the code still produces
```

source/fundamentals/aggregation-expression-operations.txt
```diff
@@ -0,0 +1,1211 @@
...
+
+   * - | `max() for MqlInteger <{+api+}/apidocs/mongodb-driver-core/com/mongodb/client/model/mql/MqlInteger.html#max(com.mongodb.client.model.mql.MqlInteger)>`__
+       | `max() for MqlNumber <{+api+}/apidocs/mongodb-driver-core/com/mongodb/client/model/mql/MqlNumber.html#max(com.mongodb.client.model.mql.MqlNumber)>`__
+
+     - *No corresponding operator*
+
+   * - | `min() for MqlInteger <{+api+}/apidocs/mongodb-driver-core/com/mongodb/client/model/mql/MqlInteger.html#min(com.mongodb.client.model.mql.MqlInteger)>`__
+       | `min() for MqlNumber <{+api+}/apidocs/mongodb-driver-core/com/mongodb/client/model/mql/MqlNumber.html#min(com.mongodb.client.model.mql.MqlNumber)>`__
+
+     - *No corresponding operator*
```
katcharov, line 507: max and min correspond to $max and $min

katcharov (changes requested):
Completes final review up to and including Comparison.

source/fundamentals/aggregation-expression-operations.txt
```diff
@@ -0,0 +1,1211 @@
...
+   The driver generates a Query API expression that may be different
+   from the Query API expression provided in each example. However,
+   both expressions will produce the same aggregation result.
+
+.. important::
+
+   The driver does not provide methods for all aggregation pipeline operators in
+   the Query API. If you need to use an unsupported operation in an
+   aggregation, you can create an expression as a BSON ``Document``
+   type. To learn more about the ``Document`` type, see :ref:`<fundamentals-documents>`.
```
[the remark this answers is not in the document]
rustagir: omitted compound because it introduces new terminology that is not needed

source/fundamentals/aggregation-expression-operations.txt
rustagir: https://docs-mongodbcom-staging.corp.mongodb.com/java/docsworker-xlarge/docsp-27390-aggregation-operations/fundamentals/aggregation-expression-operations/#array-operations:~:text=%24last-,map(),-%24map

Here?

source/fundamentals/aggregation-expression-operations.txt
rustagir: I will leave it as a tip but move it down and change the content

source/fundamentals/aggregation-expression-operations.txt
```diff
@@ -0,0 +1,1221 @@
...
+
+.. code-block:: java
+   :copyable: true
+
+   var inventory = current().getMap("inventory");
+
+   asList(project(fields(
+       computed("lowInventory", inventory
+           .entries()
+           .minN(of(2)))
```
katcharov, line 1064: I don't think that the minN will work here in a non-accidental way. While it's true that it will compare entry documents, this is unreliable if the entry's fields are "out of order" (for example, the key, which is the name of the item, gets evaluated first to determine the "smallest" value, but actually it's the value that should be compared). (And there's no real way to force order for fields.)

I don't think we have a sort operator, so I'm not sure this can work at all. If I had to do it, I would map entries into arrays that placed the inventory first, and the name of the item second (via `entries().map(e -> ofArray(e.getValue(), e.getKey())`), and then I think (but would need to verify the docs on this) that the order would be assured because the first item of the array would be guaranteed to be compared before the second. But this gets a bit convoluted.

The example might have to be changed to use something that takes a function (like `v->v.getValue()`). Maybe instead of warehouses with items, there are items with warehouses, and you do `currentItem.warehouses.entries().sum(v -> v.getValue())` to count up the item's total quantity across all warehouses it's stored in? Or something with a similar pattern.

source/fundamentals/aggregation-expression-operations.txt
```diff
@@ -0,0 +1,1221 @@
...
+   [ { $project: {
+       username: {
+           $toLower: { $concat: ["$lastName", "$employeeID"] } }
+   } } ]
+
+Type-Checking Operations
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+You can perform a type-check operation on a value of type ``MqlValue``
+using the methods described in this section.
```
katcharov, line 1152:
```suggestion
using the methods described in this section. These methods do not return boolean values, but instead allow the user to specify a default value to use when the original value, of unknown type, is not of the specified type. To perform branching logic based on type checks, see switchOn.
```
I added the (imperfect) wording above because it seems odd to call these type checks. Usually, a type-check is like  `o instanceof String` in Java, which returns a boolean, and allows for branching logic. These "type checks" are more like "casts" with a default value in case the cast fails. I guess they could be called "Casting Operations" but I suspect that would be more confusing.

source/fundamentals/aggregation-expression-operations.txt
```diff
@@ -0,0 +1,1221 @@
...
+   * - `isNumberOr() <{+api+}/apidocs/mongodb-driver-core/com/mongodb/client/model/mql/MqlValue.html#isNumberOr(com.mongodb.client.model.mql.MqlNumber)>`__
+     - *No corresponding operator*
+
+   * - `isStringOr() <{+api+}/apidocs/mongodb-driver-core/com/mongodb/client/model/mql/MqlValue.html#isStringOr(com.mongodb.client.model.mql.MqlString)>`__
+     - *No corresponding operator*
+
+Suppose you have a collection of rating data. An early version of the review
+schema allowed users to submit reviews without a star rating. You want to clean up
+the data, and convert any reviews without a star rating to be the minimum
+value of 1 star.
```
katcharov, line 1188: ...but what if the review was very good!

Just a note - this is entirely optional - it seems a bit more usual to apply a general rating to a missing specific rating:

```
var comfort = current().getField("comfort");
var overall = current().getField("overall");

computed("comfort", comfort.isNumberOr(overall))
```

source/fundamentals/aggregation-expression-operations.txt
```diff
@@ -0,0 +1,1221 @@
...
+
+   * - `isNumberOr() <{+api+}/apidocs/mongodb-driver-core/com/mongodb/client/model/mql/MqlValue.html#isNumberOr(com.mongodb.client.model.mql.MqlNumber)>`__
+     - *No corresponding operator*
+
+   * - `isStringOr() <{+api+}/apidocs/mongodb-driver-core/com/mongodb/client/model/mql/MqlValue.html#isStringOr(com.mongodb.client.model.mql.MqlString)>`__
+     - *No corresponding operator*
+
+Suppose you have a collection of rating data. An early version of the review
+schema allowed users to submit reviews without a star rating. You want to clean up
+the data, and convert any reviews without a star rating to be the minimum
```
katcharov, line 1187:
```suggestion
schema allowed users to submit reviews without a star rating. You want to 
convert any reviews without a star rating to have the minimum
```
I would think "clean up" implies changing the stored data.

source/fundamentals/aggregation-expression-operations.txt
```diff
@@ -0,0 +1,1221 @@
...
+value of 1 star.
+
+The ``isNumberOr()`` method returns either the value of ``rating``, or
+a value of ``1`` if ``rating`` is not a number or is null. The
+``project()`` method returns this value as a new ``filteredRating`` field.
```
katcharov, line 1192: It is odd to call this a filtered rating, since items are not being `filter`ed, but rather `map`ped to 1.

katcharov (changes requested):
All sections have been reviewed.

source/fundamentals/aggregation-expression-operations.txt
DBirtolo-mdb: Fair point - I was trying to make it be a case where reviews used to be 0-based for negative reviews and the system is going to 1-based for negative reviews. I added some clarification to specify this was just for negative reviews.

source/fundamentals/aggregation-expression-operations.txt
```diff
@@ -0,0 +1,1221 @@
...
+
+   {
+     "_id": ...,
+     "customer.name": "Elizabeth Ray",
+     "mailing.address":
+     {
+       "street": "9286 Bethany Glens",
+       "city": "Vasqueztown",
+       "state": "CO",
+       "zip": 22939
```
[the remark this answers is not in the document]
DBirtolo-mdb: IT's not terribly important, but I am curious why you changed this? I pulled it out of our sample data that gets loaded into Atlas if you load sample data.

katcharov (approved):
LGTM!

PR #379, opened 2023-04-27
DOCSP-29523: fix skip and limit example

ccho-mongodb:
# Pull Request Info

JIRA - https://jira.mongodb.org/browse/DOCSP-29523
Staging - https://docs-mongodbcom-staging.corp.mongodb.com/java/docsworker-xlarge/DOCSP-29523-fix-skip-and-limit-example/fundamentals/crud/read-operations/limit/

## Self-Review Checklist

### DIFF
--- a/source/fundamentals/crud/read-operations/limit.txt
+++ b/source/fundamentals/crud/read-operations/limit.txt
@@ -24,8 +24,8 @@ documents that a read operation can return. If there are not enough documents
 to reach the specified limit, it can return a smaller number.
 If you use ``limit()`` with the ``skip()`` instance method, the skip applies
 first and the limit only applies to the documents left over after
-the skip. For more information on the ``skip()`` method, see our 
-:doc:`guide on Skipping Returned Documents </fundamentals/crud/read-operations/skip/>`. 
+the skip. For more information on the ``skip()`` method, see our
+:doc:`guide on Skipping Returned Documents </fundamentals/crud/read-operations/skip/>`.
 
 The following examples demonstrate, respectively, how to insert data into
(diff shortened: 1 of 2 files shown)


source/fundamentals/crud/read-operations/limit.txt
```diff
@@ -109,14 +113,13 @@ Combining Skip and Limit
 ------------------------
 
 To see the next three longest books, append the ``skip()`` method to your
```
DBirtolo-mdb, line 115: [question]

Would it be helpful to specify that you are skipping three books here? "See the next three longest books" only makes sense in the context of the whole topic. If it said something like "to skip the first three books and see the next..." then it would make sense even if someone jumps here.

DBirtolo-mdb (approved):
LGTM - just one question to consider, but not blocking.

source/fundamentals/crud/read-operations/limit.txt
ccho-mongodb: Thanks for spotting that. I think this raises a point about what type of documentation this is (as well as other Fundamentals content) and whether we would expect people to start from a prior section such as in a tutorial or how-to. When this page was authored, I think the assumption (whether correct) was that the reader would be reading straight through and trying the code.

I'll create a ticket to try to capture a summary of your comment/request and link back to here so that we can address it later.

Feel free to create tickets when you noticed other similar items. For this PR, I want to limit the scope directly to the problem defined in the ticket since it was user-reported feedback on broken content.

ccho-mongodb: https://jira.mongodb.org/browse/DOCSP-29547

PR #389, opened 2023-05-04
DOCSP-24433: SEO meta descriptions

ccho-mongodb:
# Pull Request Info

JIRA - https://jira.mongodb.org/browse/DOCSP-24433
Staging - None (meta descriptions only)

Currently, meta descriptions are not rendering on staging nor the published site, so will wait to merge this.

## Self-Review Checklist


source/usage-examples/delete-operations.txt
```diff
@@ -2,7 +2,8 @@
 Delete Operations
 =================
 
-.. default-domain:: mongodb
+.. meta::
+   :description: See examples of deleting data from MongoDB by using the {+driver-long+}.
```
rustagir, line 6: Comment; the phrase "deleting data from" sounds awkward. could it still be "deleting data IN.."?

source/usage-examples/find-operations.txt
```diff
@@ -2,7 +2,8 @@
 Find Operations
 ===============
 
-.. default-domain:: mongodb
+.. meta::
+   :description: See examples of querying and retrieving data from MongoDB by using the {+driver-long+}.
```
rustagir, line 6: C: since "querying" and "retrieving data from" aren't exactly parallel, could we just remove one of them?

source/usage-examples/insert-operations.txt
```diff
@@ -2,7 +2,8 @@
 Insert Operations
 =================
 
-.. default-domain:: mongodb
+.. meta::
+   :description: See examples of inserting data into MongoDB by using the {+driver-long+}.
```
rustagir, line 6: Suggestion: for the usage examples, does this sound a little more natural?

```suggestion
   :description: See how to insert data into MongoDB by using the {+driver-long+}.
```

source/fundamentals/crud/write-operations.txt
```diff
@@ -2,7 +2,8 @@
 Write Operations
 ================
 
-.. default-domain:: mongodb
+.. meta::
+   :description: Learn the concepts related to storing data to MongoDB by using the {+driver-long+}.
```
rustagir, line 6:
```suggestion
   :description: Learn the concepts related to writing data to MongoDB by using the {+driver-long+}.
```

rustagir (approved):
left some non-blocking comments. everything looks good, but I think some adjustments could be made to make the language a little more natural

source/usage-examples/delete-operations.txt
ccho-mongodb: Hm, I think I've always heard the preposition "from" associated with "delete".
E.g. "Delete from your computer" vs "Delete in your computer"
Have you heard the latter before?

source/usage-examples/find-operations.txt
ccho-mongodb: I think they are two separate actions, so maybe "querying data and retrieving data" would be appropriate.
Querying is the action of formulating instructions and sending the command to performing matching whereas retrieving is accessing the results, perhaps from a cursor.

source/usage-examples/insert-operations.txt
ccho-mongodb: By natural, did you mean avoiding using "examples" when the title of the page might include "examples"?
I don't have a strong preference for either your suggestion or the original text.

source/usage-examples/delete-operations.txt
rustagir: The way i was thinking about it was that we are working "in" MongoDB and any operations are executed from that point of view. But I see your point too!

source/usage-examples/insert-operations.txt
rustagir: I think including the word example is unnecessary-- including the word "how" might make it easier for users to pick up the point of these pages

ccho-mongodb: Makes sense. I think maybe I'll update this and all similar instances to "Learn how" since I think that seems more frequently used in our style guide examples.

ccho-mongodb: On second thought, the title of the page and how it would show up in search results is "Insert Operations", not "Usage Examples", so I think "examples" is important to mention. I will try rephrasing it a bit to avoid the gerund.

rustagir (commented):
Commenting to express my thoughts on how to describe running an operation

PR #390, opened 2023-05-04
DOCSP-21335 - 4.5.1 Release Note

mongoKart:
Added as release note only (see comments to Jira ticket). 
Modeled on 4.7.1 update on the same page.

# Pull Request Info

JIRA - https://jira.mongodb.org/browse/DOCSP-21335
Staging - https://docs-mongodbcom-staging.corp.mongodb.com/java/docsworker-xlarge/docsp-21335-nxdomain-error/whats-new/#std-label-version-4.5.1

## Self-Review Checklist

- [X] Did you run a grammar-check?
- [X] Are all the links working?

### DIFF
--- a/source/whats-new.txt
+++ b/source/whats-new.txt
@@ -20,6 +20,7 @@ Learn what's new in:
 * :ref:`Version 4.7.1 <version-4.7.1>`
 * :ref:`Version 4.7.0 <version-4.7.0>`
 * :ref:`Version 4.6 <version-4.6>`
+* :ref:`Version 4.5.1 <version-4.5.1>`
 * :ref:`Version 4.5 <version-4.5>`
 * :ref:`Version 4.4 <version-4.4>`
 * :ref:`Version 4.3 <version-4.3>`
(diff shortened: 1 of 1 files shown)


source/whats-new.txt
```diff
@@ -206,6 +207,16 @@ New features of the 4.6 Java driver release include:
...
 
 .. _version-4.5:
+.. _version-4.5.1:
+
+What's New in 4.5.1
+-------------------
+
+If the DNS server returns an NXDomain error,
+indicating a non-existent domain, the 4.5.1
+driver no longer throws an exception.
```
DBirtolo-mdb, line 217: [question]

Non-blocking question, but I'm curious - what does it do? Fail quietly, or does it work as normal and just log a message? This seems like something that should have some type of error message associated.

DBirtolo-mdb (approved):
LGTM

source/whats-new.txt
mongoKart: It ignores it. If you look at Jeff's commits, it's literally just an empty catch block for this exception.

I tend to agree with you about handling such things, though I assume that isn't the end of the lookup process and there's other logic to identify the right host address. (I'm also trusting that Jeff et al. know a lot more about this fix than I do...)

PR #402, opened 2023-05-10
DOCS-15368 - Clarify UUID changes

mongoKart:
# Pull Request Info

JIRA - https://jira.mongodb.org/browse/DOCS-15368
Staging - https://docs-mongodbcom-staging.corp.mongodb.com/java/docsworker-xlarge/docs-15368-clarify-uuid/upgrade/

## Self-Review Checklist

- [X] Did you run a grammar-check?

### DIFF
--- a/source/upgrade.txt
+++ b/source/upgrade.txt
@@ -132,6 +132,12 @@ Version 4.0 Breaking Changes
   UUID values must explicitly specify which representation to use. You can
   specify the representation in the ``uuidRepresentation`` property of
   ``MongoClientSettings``.
+
+  The UUID representation that you specify strictly controls how the driver
+  decodes UUIDs. For example, if you specify any of the ``LEGACY`` UUID
+  representations, the driver decodes only binary objects of subtype 3
+  as UUIDs. If you specify the ``STANDARD`` UUID representation, the
+  driver decodes only binary objects of subtype 4 as UUIDs.
 - The connection pool no longer restricts the number of wait queue threads
   or asynchronous tasks that require a connection to MongoDB. The
   application should throttle requests as necessary rather than depend on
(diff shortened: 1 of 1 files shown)


source/upgrade.txt
```diff
@@ -132,6 +132,12 @@ Version 4.0 Breaking Changes
   UUID values must explicitly specify which representation to use. You can
   specify the representation in the ``uuidRepresentation`` property of
   ``MongoClientSettings``.
+
+  The UUID representation that you specify strictly controls how the driver
+  decodes UUIDs. For example, if you specify any of the ``LEGACY`` UUID
+  representations, the driver decodes only binary objects of subtype 3
```
ccho-mongodb, line 138: Suggestion:
Link to the API documentation of UuidRepresentation enum since it may not be clear what the various LEGACY UUID representations are. For example, here's the API docs for v4.0 of the driver:

http://mongodb.github.io/mongo-java-driver/4.0/apidocs/bson/org/bson/UuidRepresentation.html

Alternatively, if the intention is to present a concrete example, use the JAVA_LEGACY representation as the example.

source/upgrade.txt
```diff
@@ -132,6 +132,12 @@ Version 4.0 Breaking Changes
...
+  as UUIDs. If you specify the ``STANDARD`` UUID representation, the
+  driver decodes only binary objects of subtype 4 as UUIDs.
```
ccho-mongodb, line 140: Suggestion: 
While I read that it "strictly controls how the driver decodes UUIDs", and it may currently be implied, I think is important to explicitly call out that the prior version disregarded the setting. Perhaps something like:

"Previous versions disregarded the value of this setting setting when decoding UUIDs."

Also, I think it could be helpful to mention what would happen if they had been using the wrong encoding. Does it throw an exception and if so, what type? Users searching for the symptom could identify the issue more easily and correct the problem.

ccho-mongodb (approved):
LGTM, a couple suggestions to make the implications of the change more clear.

source/upgrade.txt
mongoKart: My testing may have been incomplete, but I wasn't able to get an exception thrown. Leaving as is for now.

PR #416, opened 2023-06-29
DOCSP-26368: FAQ entry for state should be open exception

ccho-mongodb:
# Pull Request Info

JIRA - https://jira.mongodb.org/browse/DOCSP-26368
Staging - https://docs-mongodbcom-staging.corp.mongodb.com/java/docsworker-xlarge/DOCSP-26368-FAQ-illegalstateexception/faq/#how-do-i-prevent-the--illegalstateexception--state-should-be--open--error-

## Self-Review Checklist

### DIFF
--- a/source/faq.txt
+++ b/source/faq.txt
@@ -159,7 +159,7 @@ can be closed for one of the following reasons:
   to the operation call.
 - Another thread or process closed the connection. This can happen when
   the framework that your application uses to manage your MongoDB operations,
-  such as Spring Boot, manages the ``MongoClient``.
+  such as Spring Boot, closes connections.
 
 To avoid this exception, do not call operations on ``MongoClient`` instance 
 that closed its connections.
--- a/source/faq.txt
+++ b/source/faq.txt
@@ -162,12 +162,13 @@ after any code that calls ``close()`` on it.
 .. tip::
 
    The code that closes the ``MongoClient`` instance may be difficult to
-   locate in certain cases. To locate sources of this exception, search for
-   the following cases:
+   locate in certain cases. To locate potential sources of this exception,
+   search for the following cases:
 
    - Calls to ``close()`` on a ``MongoClient`` instance
-   - Operation calls outside of the scope of a ``MongoClient`` instance that
-     is declared in a try-with-resources statement.
+   - Operation calls on a ``MongoClient`` instance that are outside the scope
+     of the try-with-resources statement in which the ``MongoClient`` is
+     declared.
 
    If your application uses a framework to manage the ``MongoClient``
    such as Spring Boot, check the documentation of the framework to locate the
--- a/source/faq.txt
+++ b/source/faq.txt
@@ -152,27 +152,31 @@ How do I prevent the "IllegalStateException: state should be: open" error?
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 You may encounter this exception if you call an operation on a ``MongoClient``
-instance that closed its connections to MongoDB. The client's connections
-can be closed for one of the following reasons:
+instance that closed its connections to MongoDB. Once the ``close()`` method
+is called on the ``MongoClient``, any further operation calls on that instance
+throw this exception.
 
-- The ``close()`` method was called on that ``MongoClient`` instance prior
-  to the operation call.
-- Another thread or process closed the connection. This can happen when
+V- Another thread or process closed the connection. This can happen when
   the framework that your application uses to manage your MongoDB operations,
   such as Spring Boot, closes connections.
 
-To avoid this exception, do not call operations on ``MongoClient`` instance 
-that closed its connections.
+To avoid this exception, do not call operations on ``MongoClient`` instance
+after any code that calls ``close()`` on it.
 
-If your application uses a framework to manage the ``MongoClient``,
-check the documentation of the framework to locate the best practices for
-managing the connection behavior.
+.. tip::
 
-.. seealso::
+   The code that closes the ``MongoClient`` instance may be difficult to
+   locate in certain cases. The ``MongoClient`` uses the singleton pattern,
+   which allows the instance to be accessed by other threads. Check for any
+   code that might access that ``MongoClient`` instance and call ``close()``
+   on it.
+
+   If your application uses a framework to manage the ``MongoClient``
+   such as Spring Boot, check the documentation of the framework to locate the
+   best practices for managing the connection behavior.
 
    To learn more about accessing MongoDB from Spring Boot, see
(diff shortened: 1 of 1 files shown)


source/faq.txt
```diff
@@ -148,6 +148,26 @@ When ``MongoClient.close()`` is called by any thread, the driver
 closes all idle sockets and closes all sockets that are in
 use as they are returned to the pool.
 
+Why does the application throw "IllegalStateException: state should be: open"?
```
rustagir, line 151: Suggestion: make the question more oriented around the user

```suggestion
Why did I receive an "IllegalStateException: state should be: open" error?
```

source/faq.txt
```diff
@@ -148,6 +148,26 @@ When ``MongoClient.close()`` is called by any thread, the driver
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+You may encounter this exception if you call an operation on a ``MongoClient``
+instance that closed its connections to MongoDB. The connections could be
```
rustagir, line 155: Question: Does the client close the connections or does the application close the connections? It could be helpful to clarify the wording here. 

```suggestion
You may encounter this exception if you call an operation on a ``MongoClient``
instance with a closed connection to MongoDB. The connections could be
```

source/faq.txt
```diff
@@ -148,6 +148,26 @@ When ``MongoClient.close()`` is called by any thread, the driver
...
+closed due to one of the following conditions:
```
rustagir, line 156:
```suggestion
instance that closed its connections to MongoDB. The connections could be
closed for one of the following reasons:
```

source/faq.txt
```diff
@@ -148,6 +148,26 @@ When ``MongoClient.close()`` is called by any thread, the driver
...
+
+- The ``close()`` method was called on that ``MongoClient`` instance
```
rustagir, line 158:
```suggestion
- You previously called the ``close()`` method on the ``MongoClient`` instance.
```

source/faq.txt
```diff
@@ -148,6 +148,26 @@ When ``MongoClient.close()`` is called by any thread, the driver
...
+- The ``MongoClient`` instance is out of scope
```
rustagir, line 159:
```suggestion
- The ``MongoClient`` instance is out of scope.
```

source/faq.txt
```diff
@@ -148,6 +148,26 @@ When ``MongoClient.close()`` is called by any thread, the driver
...
+- The framework your application uses to manage your MongoDB operations, such
+  as Spring Boot, has closed the connection automatically.
```
rustagir, line 161:
```suggestion
- The framework your application uses to manage your MongoDB operations, such
  as Spring Boot, automatically closed the connection.
```

Also suggestion: link to something related to Spring Boot

rustagir (changes requested):
just a few things!

source/faq.txt
ccho-mongodb: I'm not sure I understand the question since they both are responsible for the closing of the connection depending on perspective. An application must call the code to instruct the client to close. The client must close the connection when that method is called. Could you clarify what you meant?

source/faq.txt
ccho-mongodb: Thanks, I reworded since "receive" doesn't seem like a common action to describe exceptions.

source/faq.txt
rustagir: I think I meant to ask whether there were times when a client closes a connection on its own, without the application containing/calling the close method. In any case, i think you can keep the wording as is

rustagir (approved):
lgtm!

source/faq.txt
ccho-mongodb: From what I understand, the client, which I'm assuming is MongoClient, can open and close connections based on load and network state (see [How Does Connection Pooling Work in the Java Driver](https://www.mongodb.com/docs/drivers/java/sync/current/faq/#how-does-connection-pooling-work-in-the-java-driver-)), but when you call an operation, whether a direct one such as ListDatabaseNames() or indirectly such as by calling find() on a MongoCollection, the MongoClient queues it up for the next available connection.

However, if you call close() on the instance, or perhaps another process signals for it to close its connections, it transitions into a state that is no longer "open" and does not accept any further operations.

source/faq.txt
```diff
@@ -148,6 +148,32 @@ When ``MongoClient.close()`` is called by any thread, the driver
...
+How do I prevent the "IllegalStateException: state should be: open" error?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+You may encounter this exception if you call an operation on a ``MongoClient``
+instance that closed its connections to MongoDB. The client's connections
+can be closed for one of the following reasons:
+
+- The ``close()`` method was called on that ``MongoClient`` instance prior
+  to the operation call.
+- Another thread or process closed the connection. This can happen when
```
jyemin, line 160: I don't understand what the difference is between the first and second bullet.  The only way to "close the connection" is to call the close method on the MongoClient.  The difference only seems to be where the close is happening (how close/far it is from the code that is getting the exception, whether in the same thread or a different thread.  

Also, it's not possible for another process to directly close a MongoClient.  It would have to happen indirectly through communication through a socket, or some signal, etc.

Also, it may be the

jyemin (changes requested):

source/faq.txt
ccho-mongodb: I see. I assumed that [this response](https://www.mongodb.com/community/forums/t/illegalstateexception-state-should-be-open/223912/4) was referring to an executor service operating on a pool of threads.  Would you prefer I remove this list item to avoid any confusion? Is there any info you'd like to add? Thanks!

jyemin: Yes, please remove.  You could point out that often MongoClient instances are singletons that are accessible by multiple threads, and so the code which closes the MongoClient may be distant from the code receiving the exception, perhaps in another thread of execution.

source/faq.txt
```diff
@@ -148,6 +148,32 @@ When ``MongoClient.close()`` is called by any thread, the driver
...
+is called on the ``MongoClient``, any further operation calls on that instance
+throw this exception.
+
+To avoid this exception, do not call operations on ``MongoClient`` instance
+after any code that calls ``close()`` on it.
+
+.. tip::
+
+   The code that closes the ``MongoClient`` instance may be difficult to
+   locate in certain cases. The ``MongoClient`` uses the singleton pattern,
```
jyemin, line 165: MongoClient doesn't use the singleton pattern itself.  It's just that frameworks are often configured to manage these instances as singletons.

jyemin (changes requested):

jyemin (approved):

PR #423, opened 2023-08-03
DOCSP-31894: fix href

ccho-mongodb:
# Pull Request Info

JIRA - https://jira.mongodb.org/browse/DOCSP-31894
Staging - https://docs-mongodbcom-staging.corp.mongodb.com/java/docsworker-xlarge/DOCSP-31894-fix-href/upgrade/

## Self-Review Checklist

### DIFF
--- a/source/upgrade.txt
+++ b/source/upgrade.txt
@@ -138,9 +138,9 @@ Version 4.0 Breaking Changes
   ``JAVA_LEGACY`` representation, the driver would decode binary objects
   of subtypes 3 and 4 as UUIDs. In version 4.0, the ``JAVA_LEGACY``
   representation works only with subtype 3.
-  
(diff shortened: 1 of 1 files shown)


rustagir (approved):
lgtm - QQ: is it necessary to link specifically to the v4.0 docs? do we in general want to use these versioned links on the What's New pages? (as opposed to using the api source constants)

ccho-mongodb:
> lgtm - QQ: is it necessary to link specifically to the v4.0 docs? do we in general want to use these versioned links on the What's New pages? (as opposed to using the api source constants)

That what's new entry refers to the 4.0 docs. I think it's better to use a versioned link because it refers to something at that point in time and that info could change or even be removed in the future.

PR #434, opened 2023-09-06
DOCSP-32483 programmatic tagging script for java sync repo

sarahemlin:
JIRA - https://jira.mongodb.org/browse/DOCSP-32483

Hey @mongoKart - here's the facet.toml file for the java sync driver repo.  Please merge if everything looks good!

Thanks!
Sarah

### DIFF
--- a/source/facets.toml
+++ b/source/facets.toml
@@ -3,11 +3,5 @@ category = "target_product"
 value = "drivers"
 
 [[facets]]
-category = "target_product"
-value = "drivers"
-[[facets]]
-category = "programming_language"
-value = "java"
-[[facets]]
 category = "programming_language"
 value = "java"
--- a/source/facets.toml
+++ b/source/facets.toml
@@ -1,0 +1,13 @@
+[[facets]]
+category = "target_product"
+value = "drivers"
+
+[[facets]]
+category = "target_product"
+value = "drivers"
+[[facets]]
+category = "programming_language"
+value = "java"
+[[facets]]
+category = "programming_language"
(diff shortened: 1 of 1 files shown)


source/facets.toml
```diff
@@ -0,0 +1,13 @@
...
+
+[[facets]]
+category = "target_product"
+value = "drivers"
+[[facets]]
+category = "programming_language"
+value = "java"
+[[facets]]
+category = "programming_language"
+value = "java"
```
mongoKart, line 13: Do each of these facets (product and language) need to be included twice in this file?

sarahemlin: @mongoKart - good catch!  No, they should not be.  The source file is correct, so I'll need to reach out to Seung & Cassidy.  I think the script is picking up the docs-java-other values, and with the info you shared about that one we might just need to delete the async driver repo from the source file.

seungpark: these doubled values are erroneous! good catch @mongoKart !

https://github.com/10gen/snooty-scripts/pull/8

created an update to match repo path exactly as is in CSV (no partial matching for `docs-java-other` etc)

sarahemlin: Thanks so much, @seungpark!  I'll delete these extra 2 lines so Mike can merge, knowing that the script is now updated so the next time we run it we won't have this problem.

sarahemlin:
@mongoKart - I removed the extra lines so I think we're good to go!

mongoKart (approved):
lgtm

PR #518, opened 2024-01-26
DOCSP-35821: link to jar download

rustagir:
# Pull Request Info

JIRA - <https://jira.mongodb.org/browse/DOCSP-35821>
Staging - https://preview-mongodbrustagir.gatsbyjs.io/java/DOCSP-35821-link-jar-file/quick-start/#add-mongodb-as-a-dependency

## Self-Review Checklist

- [x] Are the [facets and meta keywords](https://wiki.corp.mongodb.com/display/DE/Docs+Taxonomy) accurate?

### DIFF
--- a/source/quick-start.txt
+++ b/source/quick-start.txt
@@ -56,7 +56,7 @@ If you are using `Gradle <https://gradle.org/>`__, add the following to your
    {+driver-short+}. However, if you must download the driver
    and dependencies, you can find the JAR files in the `MongoDB
    Group Maven repository <https://mvnrepository.com/artifact/org.mongodb/>`__.
-   To use the {+driver-short+}, you must also download the JAR files of
+   To use the {+driver-short+}, you must also download the JAR files for
    the ``bson``, ``mongodb-driver-core``, and ``slf4j-api`` dependencies.
 
 Once you configure your dependencies, ensure they are available to your
--- a/source/quick-start.txt
+++ b/source/quick-start.txt
@@ -50,9 +50,9 @@ If you are using `Gradle <https://gradle.org/>`__, add the following to your
 
 .. include:: /includes/quick-start/gradle-versioned.rst
 
-.. tip:: Download ``.jar`` File
+.. tip:: Download the JAR File
    
-   You can directly download the ``.jar`` file for the {+driver-short+}
+   You can directly download the JAR file for the {+driver-short+}
    from the `Maven MongoDB Driver artifact entry
    <https://mvnrepository.com/artifact/org.mongodb/mongodb-driver-sync/{+full-version+}>`__.
 
--- a/source/quick-start.txt
+++ b/source/quick-start.txt
@@ -52,10 +52,12 @@ If you are using `Gradle <https://gradle.org/>`__, add the following to your
 
 .. note::
    
-   We recommend that you use a package manager to install the
+   We recommend that you use a build tool to install the
    {+driver-short+}. However, if you must download the driver
-   and dependency packages, you can find the JAR files in the `Maven
-   repository <https://mvnrepository.com/artifact/org.mongodb/mongodb-driver-sync/>`__.
+   and dependency packages, you can find the JAR files in the `MongoDB
+   Group Maven repository <https://mvnrepository.com/artifact/org.mongodb/>`__.
+   To use the {+driver-short+}, you must also install the ``bson``,
+   ``mongodb-driver-core``, and ``slf4j-api`` packages.
 
 Once you configure your dependencies, ensure they are available to your
 project which may require running your dependency manager and refreshing
--- a/source/quick-start.txt
+++ b/source/quick-start.txt
@@ -52,8 +52,8 @@ If you are using `Gradle <https://gradle.org/>`__, add the following to your
 
 .. tip:: Download the JAR File
    
-   You can directly download the JAR file for the {+driver-short+}
-   from the `Maven MongoDB Driver artifact entry
+   You can download the JAR file for the {+driver-short+}
+   from the `Maven repository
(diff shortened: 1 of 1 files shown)


source/quick-start.txt
```diff
@@ -50,6 +50,12 @@ If you are using `Gradle <https://gradle.org/>`__, add the following to your
 
 .. include:: /includes/quick-start/gradle-versioned.rst
 
+.. tip:: Download ``.jar`` File
+   
+   You can directly download the ``.jar`` file for the {+driver-short+}
```
ccho-mongodb, line 55: Suggestion:
I think JAR is usually not referenced by it's filename extension, but rather by the file format "JAR". See https://en.wikipedia.org/wiki/JAR_(file_format)

E.g.
tip:: Download the JAR File

source/quick-start.txt
```diff
@@ -50,6 +50,12 @@ If you are using `Gradle <https://gradle.org/>`__, add the following to your
 
 .. include:: /includes/quick-start/gradle-versioned.rst
 
+.. tip:: Download the JAR File
+   
+   You can directly download the JAR file for the {+driver-short+}
```
ccho-mongodb, line 55: Suggestion:
I think "directly" could provide contrast when compared with another option, but I think omitting it when there isn't a comparison being made would be better.

"You can download..."

source/quick-start.txt
```diff
@@ -50,6 +50,12 @@ If you are using `Gradle <https://gradle.org/>`__, add the following to your
...
+   from the `Maven MongoDB Driver artifact entry
+   <https://mvnrepository.com/artifact/org.mongodb/mongodb-driver-sync/{+full-version+}>`__.
+
```
ccho-mongodb, line 58: Suggestion:
I think this could use similar wording as [this download mention](https://github.com/mongodb/docs-java/blob/master/source/fundamentals/data-formats/document-data-format-bson.txt#L89-L91):

"You can download the JAR file for the MongoDB Java Driver from the Maven repository."

source/quick-start.txt
```diff
@@ -50,6 +50,12 @@ If you are using `Gradle <https://gradle.org/>`__, add the following to your
 
 .. include:: /includes/quick-start/gradle-versioned.rst
 
+.. tip:: Download the JAR File
```
ccho-mongodb, line 53: Issue/Suggestion:

Could you confirm whether you might need to make the following changes in a technical review?

My guess is that the goal of the ticket is to provide a download location so that the user can download and add them to their app classpath so that they can perform MongoDB operations. 

However, I think the mongodb-driver-sync JAR file has [these dependencies](https://central.sonatype.com/artifact/org.mongodb/mongodb-driver-sync/dependencies) which means it won't work without them. For context, there used to be another JAR file that contained all the dependencies (called the Uber JAR), but the DBX team no longer provides builds for that.

The current quick start instructions depend on Maven or Gradle to manage all the dependencies of the mongodb-driver-sync JAR.

I think one of the following solutions could improve the usefulness of this info:

- Point the user to the Sonatype repository and mention that they'll need to download the dependencies https://central.sonatype.com/artifact/org.mongodb/mongodb-driver-sync/
- Provide links to each of the dependency JAR files that they would need to download and add to their classpath to run the app.

ccho-mongodb (changes requested):
Thanks for picking up this ticket. I think this accomplishes the goal of the ticket, but I think the information might cause confusion without additional details.

I think the content might change after confirming with the DBX team, so requesting changes so I can look again after that.

source/quick-start.txt
rustagir: I'll check what DBX and the ticket creator prefer!

source/quick-start.txt
```diff
@@ -50,6 +50,13 @@ If you are using `Gradle <https://gradle.org/>`__, add the following to your
 
 .. include:: /includes/quick-start/gradle-versioned.rst
 
+.. note::
+   
+   We recommend that you use a package manager to install the
```
ccho-mongodb, line 55: Suggestion:
I think Maven and Gradle are more often described as "build tools" or "dependency managers" based on the specific part of the functionality that's being described.

For example:
- [Gradle Build Tool (newsletter page)](https://newsletter.gradle.org/)
- [Apache Maven 3.x description](https://maven.apache.org/ref/3.9.6/)
   > Maven is a project development management and comprehension tool. Based on the concept of a project object model: builds, dependency management, documentation creation, site publication, and distribution publication are all controlled from [the pom.xml declarative file](https://maven.apache.org/ref/3.9.6/maven-model/maven.html). Maven can be extended by [plugins](https://maven.apache.org/plugins/) to utilise a number of other development tools for reporting or the build process.

source/quick-start.txt
```diff
@@ -50,6 +50,13 @@ If you are using `Gradle <https://gradle.org/>`__, add the following to your
...
+   {+driver-short+}. However, if you must download the driver
+   and dependency packages, you can find the JAR files in the `Maven
+   repository <https://mvnrepository.com/artifact/org.mongodb/mongodb-driver-sync/>`__.
```
ccho-mongodb, line 58: Issue:

The current link only contains the mongodb-driver-sync JAR and not any of the dependencies. 

Suggestion:
I think it could be more helpful to specify the group (which references all org.mongodb JARs) and which ones are necessary.

```suggestion
   and its dependencies, you can find the JAR files in the `MongoDB Group Maven
   repository <https://mvnrepository.com/artifact/org.mongodb/>`__. The Java Sync
   driver requires the ``mongodb-driver-core``, ``bson``, and ``mongodb-driver-sync``
   JAR files.
```

ccho-mongodb (approved):
A couple suggestions, but do not need to review again unless requested.

PR #537, opened 2024-04-25
DOCSP-34479 - OIDC

mongoKart:
# Pull Request Info

JIRA - https://jira.mongodb.org/browse/DOCSP-34479
Staging - https://preview-mongodbmongokart.gatsbyjs.io/java/docsp-34479-oidc/fundamentals/enterprise-auth/#mongodb-oidc

## Self-Review Checklist

- [x] Are the [facets and meta keywords](https://wiki.corp.mongodb.com/display/DE/Docs+Taxonomy) accurate?


source/fundamentals/auth.txt
```diff
@@ -575,3 +576,146 @@ mechanism:
...
+
+      The following code example shows how to specify Azure IMDS OIDC authenticaton.
+      Replace the ``<username>`` placeholder with the client ID of the Azure managed
+      identity, and replace the ``<resource>`` placeholder with the percent-encoded
+      application or service that the OIDC access token is intended for.
+      
+      .. code-block:: java
+
+         MongoCredential credential = MongoCredential.createOidcCredential("<username>")
+            .withMechanismProperty("ALLOWED_HOSTS", allowedHosts)
```
mongoKart, line 634: I couldn't information on this property anywhere, beyond it being a `List<String>`; I'm guessing it's new as part of the OIDC work. It would be good if we could provide some info on it.

source/fundamentals/auth.txt
```diff
@@ -575,3 +576,146 @@ mechanism:
...
+
+The following code example shows how to use the ``"OIDC_CALLBACK"`` property to include
+a callback method in your ``MongoCredential`` object. In this example, the callback
+retrieves an OIDC token from a file named ``"access-token.dat"`` in the local file system.
+
+.. code-block:: java
+
+   MongoCredential credential = new Credential()
+      .withMechanismProperty("ALLOWED_HOSTS", allowedHosts)
+      .withMechanismProperty("OIDC_CALLBACK", (context) ->  {
```
mongoKart, line 711: What does this `context` object look like?

source/fundamentals/auth.txt
katcharov: It is new, and important, but is only used when a human callback is being used.

source/fundamentals/auth.txt
```diff
@@ -575,3 +576,146 @@ mechanism:
...
+   .. tab:: Connection String
+      :tabid: mongodb-gcp-imds-connection-string
+    
+      The following code example shows how to specify GCP IMDS OIDC authentication.
+      Replace the ``<resource>`` placeholder with the percent-encoded application or
+      service that the OIDC access token is intended for.
+     
+      .. code-block:: java
+
+         MongoClient mongoClient = MongoClients.create("mongodb://<username>@<hostname>:<port>/?authMechanism=MONGODB-OIDC&authMechanismProperties=ENVIRONMENT:gcp,TOKEN_RESOURCE:<resource>");
```
katcharov, line 671: Something will need to be said about URI-encoding the resource, once the issue is resolved in [this PR](https://github.com/mongodb/specifications/pull/1569).

mongoKart: I followed the PyMongo docs and used "percent-encoded" on line 666 above.

source/fundamentals/auth.txt
mongoKart: What do we want users to know about it?

mongoKart: found in auth spec

source/fundamentals/enterprise-auth.txt
```diff
@@ -280,3 +287,150 @@ mechanism:
       method. The code to instantiate a ``MongoClient`` resembles the following:
 
       .. include:: /includes/fundamentals/code-snippets/auth-credentials-ldap.rst
+
+.. _mongodb-oidc:
+
+``MONGODB-OIDC``
```
norareidy, line 293: S: since the other authentication mechanisms aren't monospaced, I think this header format should align with those
```suggestion
MONGODB-OIDC
```

source/fundamentals/enterprise-auth.txt
```diff
@@ -280,3 +287,150 @@ mechanism:
...
+
+Azure IMDS
+++++++++++
+
+If your application runs on an Azure VM, or otherwise uses the
+`Azure Instance Metadata Service <https://learn.microsoft.com/en-us/azure/virtual-machines/instance-metadata-service>`__
+(IMDS), you can authenticate to MongoDB by using the {+driver-short+}'s built-in Azure
+support.
+
+You can specify Azure IMDS OIDC authentication on a ``MongoClientSettings`` object either by
```
norareidy, line 318: I: it looks like the MongoCredential code snippet uses a MongoClientSettings object, but the Connection String code does not. Maybe this should say "`MongoClient` object" instead?
```suggestion
You can specify Azure IMDS OIDC authentication on a ``MongoClient`` object either by 
```

source/fundamentals/enterprise-auth.txt
```diff
@@ -280,3 +287,150 @@ mechanism:
...
+
+GCP IMDS
+++++++++++
+
+If your application runs on a GCP VM, or otherwise uses the
+`GCP Instance Metadata Service <https://cloud.google.com/compute/docs/metadata/querying-metadata>`__,
+you can authenticate to MongoDB by using {+driver-short+}'s built-in GCP
+support.
+
+You can specify GCP IMDS OIDC authentication on a ``MongoClientSettings`` object either by
```
norareidy, line 365: I: same comment as above about MongoClientSettings
```suggestion
You can specify GCP IMDS OIDC authentication on a ``MongoClient`` object either by 
```

source/fundamentals/enterprise-auth.txt
```diff
@@ -280,3 +287,150 @@ mechanism:
...
+
+Custom Callback
++++++++++++++++
+
+On some platforms, such as Azure Functions and Azure Kubernetes Service (AKS), you
+must define a custom callback method to use OIDC to authenticate. To use a custom
+callback to retrieve an OIDC token, call the ``withMechanismProperty()`` method on
+your ``MongoCredential`` object to set the following options:
+
+- ``"ALLOWED_HOSTS"``: The list of allowed hostnames or IP addresses for MongoDB
```
norareidy, line 410: I: add a period
```suggestion
- ``"ALLOWED_HOSTS"``: The list of allowed hostnames or IP addresses for MongoDB.
```

source/fundamentals/enterprise-auth.txt
```diff
@@ -280,3 +287,150 @@ mechanism:
...
+  The hostnames may include ``"*."`` as a leading wildcard, which allows for
+  matching subdomains. This property defaults to
+  ``["*.mongodb.net", "*.mongodb-qa.net", "*.mongodb-dev.net", "*.mongodbgov.net",
+  "localhost", "127.0.0.1", "::1"]``.
```
norareidy, line 414: S: looks like this value is cut off on the staging site and requires some horizontal scroll. Maybe adding a newline after  `"*.mongodb-dev.net" `could fix it:
```suggestion
  ``["*.mongodb.net", "*.mongodb-qa.net", "*.mongodb-dev.net", 
  "*.mongodbgov.net", "localhost", "127.0.0.1", "::1"]``.
```

source/fundamentals/enterprise-auth.txt
```diff
@@ -280,3 +287,150 @@ mechanism:
...
+- ``"OIDC_CALLBACK"``: The method that retrieves the OIDC token. This method must accept
```
norareidy, line 415: Q: is `"OIDC_CALLBACK"` a method? Or an option/parameter?

source/fundamentals/enterprise-auth.txt
```diff
@@ -280,3 +287,150 @@ mechanism:
...
       .. include:: /includes/fundamentals/code-snippets/auth-credentials-ldap.rst
+
+.. _mongodb-oidc:
+
+``MONGODB-OIDC``
+~~~~~~~~~~~~~~~~
+
+.. important::
+
+   The MONGODB-OIDC authentication mechanism requires MongoDB v7.0 or later running
```
norareidy, line 298: S: I'd use {+mdb-server+} here
```suggestion
   The MONGODB-OIDC authentication mechanism requires {+mdb-server+} v7.0 or later running
```

norareidy (changes requested):
Looks good, just a few suggestions + question!

source/fundamentals/enterprise-auth.txt
mongoKart: good call

source/fundamentals/enterprise-auth.txt
mongoKart: It's an authentication mechanism property that works with the OIDC auth mechanism. The whole thing is like a key-value pair, where "OIDC_CALLBACK" is the key and the callback method is the value.

source/fundamentals/enterprise-auth.txt
mongoKart: good thought. this content was taken from pymongo, which doesn't have that constant.

source/fundamentals/auth.txt
katcharov: https://github.com/mongodb/mongo-java-driver/blob/479fcddaa743f2e8cc85733e548d4b9fe67b16b4/driver-core/src/main/com/mongodb/MongoCredential.java#L670

Users should get the timeout from it, and make sure the lambda completes within that time.

source/fundamentals/enterprise-auth.txt
```diff
@@ -280,3 +287,154 @@ mechanism:
...
+
+      Replace the ``<username>`` placeholder with the client ID of the GCP managed
+      identity. Replace the ``<resource>`` placeholder with the
+      application or service that the OIDC access token is intended for.
+      You must percent-encode the resource value
+      if it contains a comma (``,``), plus sign (``+``), or ampersand (``&``). 
+      
+      .. code-block:: java
+
+         MongoCredential credential = MongoCredential.createOidcCredential("<username>")
```
katcharov, line 395: GCP does not accept a username, trying to supply one will fail.

```suggestion
         MongoCredential credential = MongoCredential.createOidcCredential(null)
```

source/fundamentals/enterprise-auth.txt
```diff
@@ -280,3 +287,154 @@ mechanism:
...
+
+Custom Callback
++++++++++++++++
+
+On some platforms, such as Azure Functions and Azure Kubernetes Service (AKS), you
+must define a custom callback method to use OIDC to authenticate. To use a custom
+callback to retrieve an OIDC token, call the ``withMechanismProperty()`` method on
+your ``MongoCredential`` object to set the following options:
+
+- ``"ALLOWED_HOSTS"``: The list of allowed hostnames or IP addresses for MongoDB.
```
katcharov, line 414: ALLOWED_HOSTS are only checked before calling human callbacks, and should not be specified otherwise.

source/fundamentals/enterprise-auth.txt
```diff
@@ -280,3 +287,154 @@ mechanism:
...
+                 .applyToClusterSettings(builder ->
+                         builder.hosts(Arrays.asList(new ServerAddress("<hostname>", <port>))))
+                 .credential(credential)
+                 .build());
+
+Custom Callback
++++++++++++++++
+
+On some platforms, such as Azure Functions and Azure Kubernetes Service (AKS), you
+must define a custom callback method to use OIDC to authenticate. To use a custom
```
katcharov, line 410: Perhaps "The driver does not offer built-in support for other platforms like ..., but you can define a custom callback..."

source/fundamentals/enterprise-auth.txt
```diff
@@ -280,3 +287,154 @@ mechanism:
...
+callback to retrieve an OIDC token, call the ``withMechanismProperty()`` method on
+your ``MongoCredential`` object to set the following options:
+
+- ``"ALLOWED_HOSTS"``: The list of allowed hostnames or IP addresses for MongoDB.
+  The hostnames may include ``"*."`` as a leading wildcard, which allows for
+  matching subdomains. By default, this property includes ``"*.mongodb.net"``,
+  ``"*.mongodb-qa.net"``, ``"*.mongodb-dev.net"``, ``"*.mongodbgov.net"``, ``"localhost"``,
+  ``"127.0.0.1"``, and ``"::1"``.
+- ``"OIDC_CALLBACK"``: The method that retrieves the OIDC token. This method must accept
+  a ``context`` object as a parameter and return an ``OidcCallbackResult`` object.
```
katcharov, line 420: Let's not call this a method, but rather "A lambda or other implementation of the `OidcCallback` functional interface, which must accept an `OidcCallbackContext` as a parameter and return an `OidcCallbackResult`".

source/fundamentals/enterprise-auth.txt
```diff
@@ -280,3 +287,154 @@ mechanism:
...
+
+The following code example shows how to include
+a callback method in your ``MongoCredential`` object. In this example, the callback
+retrieves an OIDC token from a file named ``"access-token.dat"`` in the local file system.
+
+.. code-block:: java
+
+   MongoCredential credential = new Credential()
```
katcharov, line 428: This needs to be a `MongoCredential.createOidcCredential(null)`

katcharov (changes requested):
Some of the comments might apply to other languages as well.

norareidy (approved):
LGTM!

source/fundamentals/enterprise-auth.txt
```diff
@@ -280,3 +287,154 @@ mechanism:
...
+
+If your application runs on an Azure VM, or otherwise uses the
+`Azure Instance Metadata Service <https://learn.microsoft.com/en-us/azure/virtual-machines/instance-metadata-service>`__
+(IMDS), you can authenticate to MongoDB by using the {+driver-short+}'s built-in Azure
+support.
+
+You can specify Azure IMDS OIDC authentication either by 
+using a ``MongoCredential`` object or as part of the connection string. Select the 
+:guilabel:`Connection String` or :guilabel:`MongoCredential` tab to
+see the corresponding syntax.
```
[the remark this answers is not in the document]
mongoKart: I generally don't include an explanation like this one, but did so here to keep with examples on other pages for this driver. I do agree that visual indicators of the tabs' scope would be valuable.

source/fundamentals/enterprise-auth.txt
```diff
@@ -280,3 +287,154 @@ mechanism:
...
+         MongoClient mongoClient = MongoClients.create("mongodb://<username>@<hostname>:<port>/?authMechanism=MONGODB-OIDC&authMechanismProperties=ENVIRONMENT:azure,TOKEN_RESOURCE:<resource>");
+
+   .. tab:: MongoCredential
+      :tabid: mongodb-azure-mongo-credential
+
+      Replace the ``<username>`` placeholder with the client ID or application ID of the
+      Azure managed identity or enterprise application. Replace the ``<resource>``
+      placeholder with the application or service that the OIDC access token is intended for.
+      You must percent-encode the resource value
+      if it contains a comma (``,``), plus sign (``+``), or ampersand (``&``).
```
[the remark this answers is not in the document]
mongoKart: Ah--I must have misunderstood the language in [PYTHON-3601](https://github.com/mongodb/mongo-python-driver/pull/1621/files). Thanks for the correction.

source/fundamentals/enterprise-auth.txt
```diff
@@ -280,3 +287,154 @@ mechanism:
...
+:guilabel:`Connection String` or :guilabel:`MongoCredential` tab to
+see the corresponding syntax.
+
+.. tabs::
+
+   .. tab:: Connection String
+      :tabid: mongodb-gcp-imds-connection-string
+    
+      Replace the ``<resource>`` placeholder with the percent-encoded application or
+      service that the OIDC access token is intended for.
```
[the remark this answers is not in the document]
mongoKart: I noticed this discrepancy: PyMongo uses "audience" while Java and C# use "resource". 

I would prefer to be consistent across drivers, so unless it's either flatly incorrect or confusing to end users, I'll use "audience" here too.

source/fundamentals/enterprise-auth.txt
mongoKart: I believe I understand now. I had to read the spec to see the difference between OIDC_CALLBACK and OIDC_HUMAN_CALLBACK.

source/fundamentals/enterprise-auth.txt
mongoKart: Thanks--something felt off about it.

source/fundamentals/enterprise-auth.txt
```diff
@@ -280,3 +287,160 @@ mechanism:
...
+must define a custom callback to use OIDC to authenticate from these platforms.
+To do so, use the ``"OIDC_CALLBACK"`` authentication property, as shown in the following
+code example:
+
+.. code-block:: java
+
+   MongoCredential credential = MongoCredential.createOidcCredential(null)
+      .withMechanismProperty("OIDC_CALLBACK", (context) ->  {
+         ... 
+         return new OidcCallbackResult(...);
```
katcharov, line 423:
```suggestion
         String accessToken = ...
         return new OidcCallbackResult(accessToken);
```
Since this is the machine callback, let's make it clear that they should use the single-parameter result constructor.

source/fundamentals/enterprise-auth.txt
```diff
@@ -280,3 +287,160 @@ mechanism:
...
+
+The following example uses an example callback to retrieve an OIDC token from a file
+named ``"access-token.dat"`` in the local file system:
+
+.. code-block:: java
+
+   MongoCredential credential = MongoCredential.createOidcCredential(null)
+      .withMechanismProperty("OIDC_CALLBACK", (context) ->  {
+         string accessToken = new String(Files.readAllBytes(Paths.get("access-token.dat"));
+         return new OidcCallbackResult(accessToken, 0);
```
katcharov, line 438: Let's not use this constructor, since with 0 it is equivalent to the below, and while we offer the ability to set the expiry, it is technically ignored in all drivers.
```suggestion
         return new OidcCallbackResult(accessToken);
```

source/fundamentals/enterprise-auth.txt
```diff
@@ -280,3 +287,160 @@ mechanism:
...
+
+   MongoCredential credential = MongoCredential.createOidcCredential(null)
+      .withMechanismProperty("OIDC_CALLBACK", (context) ->  {
+         ... 
+         return new OidcCallbackResult(...);
+      });
+
+The value of the ``"OIDC_CALLBACK"`` property must be a lambda or other implementation
+of the ``OidcCallback`` functional interface that accepts an ``OidcCallbackContext`` object
+as a parameter and returns an ``OidcCallbackResult`` object.
```
katcharov, line 428:
```suggestion
of the ``OidcCallback`` functional interface that accepts an ``OidcCallbackContext``
as a parameter and returns an ``OidcCallbackResult``.
```
This is in line with existing docs, like "You can supply your AWS IAM credentials to a `MongoClient` by using a `MongoCredential` instance" (does not say "`MongoClient object`").

The use of the term "object" here seems off to me, since we are talking about signatures, which concern types rather than instances. When talking about instances ("then construct and return an `OidcCallbackResult` [instance]"), "instance" is preferable to "object", if a term must be used for clarity. `OidcCallbackContext` is an interface, which are implemented rather than instantiated directly.

source/fundamentals/enterprise-auth.txt
```diff
@@ -280,3 +287,160 @@ mechanism:
...
+:guilabel:`Connection String` or :guilabel:`MongoCredential` tab to
+see the corresponding syntax.
+
+.. tabs::
+
+   .. tab:: Connection String
+      :tabid: mongodb-azure-imds-connection-string
+    
+      Replace the ``<percent-encoded audience>`` placeholder with the percent-encoded
+      value of the ``audience`` parameter configured on your MongoDB deployment.
```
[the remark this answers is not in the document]
mongoKart: changed, but left lowercase to match server docs.

source/fundamentals/enterprise-auth.txt
mongoKart: i've had docs reviewers flag this in the past, but omitting object/instance/etc. usually seems fine to me.

katcharov (approved):
LGTM!

PR #553, opened 2024-06-11
DOCSP-40348: disallow commas in oidc prop val

rustagir:
# Pull Request Info

JIRA - <https://jira.mongodb.org/browse/DOCSP-40348>
Staging - https://preview-mongodbrustagir.gatsbyjs.io/java/DOCSP-40348-oidc-comma-spec/fundamentals/enterprise-auth/#mongodb-oidc

## Self-Review Checklist

- [ ] Are the [facets and meta keywords](https://wiki.corp.mongodb.com/display/DE/Docs+Taxonomy) accurate?


source/includes/fundamentals/oidc-value-comma.rst
```diff
@@ -0,0 +1,22 @@
+.. note::
+
+   The ``authMechanismProperties`` connection string option takes a list
+   of key-value pairs. A *value* in any pair must not contain a comma
+   (``,``) character, to prevent the risk that the driver parses your
```
shuangela, line 5:
```suggestion
   (``,``) character, to prevent the driver from parsing your
```

shuangela (changes requested):
Looks great! A few comments to think about.

source/includes/fundamentals/oidc-value-comma.rst
```diff
@@ -0,0 +1,22 @@
...
+   connection string ambiguously. A value can contain a colon (``:``)
+   character, however.
+
+   For example, the following code shows a connection string that the
+   driver might parse incorrectly:
+
+   .. code-block:: java
+      :emphasize-lines: 4
+
+      String uri =
```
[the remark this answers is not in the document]
rustagir: Will wait for technical review to address this question

shuangela (approved):
LGTM! Thanks for changing those!

source/includes/fundamentals/oidc-value-comma.rst
```diff
@@ -0,0 +1,22 @@
+.. note::
+
+   The ``authMechanismProperties`` connection string option takes a list
+   of key-value pairs. A *value* in any pair must not contain a comma
+   (``,``) character, to prevent the driver from parsing your
+   connection string ambiguously. A value can contain a colon (``:``)
+   character, however.
+
+   For example, the following code shows a connection string that the
+   driver might parse incorrectly:
```
katcharov, line 10: This note seems too large. Consider the existing text,

> Replace the <percent-encoded audience> placeholder with the percent-encoded value of the audience server parameter configured on your MongoDB deployment.

We could have a large note explaining which characters would need to be percent-encoded, and which are allowed, and an example of an invalid string. However, we communicate all of this by adding the adjective "percent-encoded" in two places inline with the normal text. The note is also a bit confusing, because the parsing is not ambiguous, the issue is that `,` is a reserved character whether or not it is encoded, and whether it is in the value or the key. Also, the note location is above both tabs, but it should only be inside the connection string tab, since it only applies there.

Let's say:

> Replace the <percent-encoded audience> placeholder with the percent-encoded value of the audience server parameter configured on your MongoDB deployment. The comma (`,`) and its encoding (`%2C`) are reserved: if the unencoded value contains a comma (`,`), it must not be specified in the connection string, but in a ``MongoCredential`` instance, as demonstrated in the :guilabel:`MongoCredential` tab.

source/includes/fundamentals/oidc-value-comma.rst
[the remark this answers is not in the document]
katcharov: The value must always be percent-encoded. The existing docs correctly state "Replace the placeholder with the percent-encoded value of the audience server parameter". However, the behaviour is that commas will now be parsed as a new property, even when encoded.

katcharov (changes requested):

source/includes/fundamentals/oidc-value-comma.rst
rustagir: Thank you for clarifying!

katcharov (approved):
LGTM!

